PR #99, opened 2024-11-21
ci: For merge groups, add event type, skip publishing docs

qmonnet:
**Fix for working with merge queues.**

We do not want to publish a new version of the documentation when running CI checks in a merge queue. Let's skip the relevant steps.

Let's also specify the event type for merge groups. As per GitHub's documentation:

    Although only the checks_requested activity type is supported,
    specifying the activity type will keep your workflow specific if
    more activity types are added in the future.

Note: Looking more specifically at the docs workflow, I think we could improve the workflow regarding whether or not to publish. In particular, we could have another bool parameter for the `workflow_dispatch`, because we don't necessarily want to publish a new version when running it. I think that ideally, we should split the `build` workflow for docs into `build`, `publish`, `preview`. But given that I don't see this documentation remain in the repository long term (the docs for the Gateway will likely be in the official docs for the Fabric), I haven't spent more time on this for now.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -15,6 +15,7 @@ on:
     branches:
       - "main"
   merge_group:
+    types: ["checks_requested"]
   workflow_dispatch:
     inputs:
       debug_enabled:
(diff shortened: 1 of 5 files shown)


daniel-noland (approved):
This makes sense.  Thanks

qmonnet:
Fails to pass merge queue because `mergeability.yml` is required but doesn't run in merge queue. I'll add a commit to fix this.

PR #101, opened 2024-12-02
ci: Fix apt-get command in bump.yml, restore workflow summaries

qmonnet:
The bump.yml workflow, running on a schedule, has [always failed to complete so far][0]. This is because an "apt-get" command is missing the "install" subcommand, resulting in apt trying to interpret the name of the package we want to install as a subcommand, making the workflow fail.

Fix the command, to hopefully make the workflow succeed.

[0]: https://github.com/githedgehog/dataplane/actions/workflows/bump.yml

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -51,7 +51,7 @@ jobs:
       - name: "install envsubst"
         run: |
           sudo apt-get update
-          sudo apt-get --yes --no-install-recommends gettext
+          sudo apt-get install --yes --no-install-recommends gettext
       - run: |
           ./scripts/update-versions.sh
       - name: "Create Pull Request"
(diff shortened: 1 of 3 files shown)


qmonnet:
We can bypass the rule here, but I'll need to find a way to address this. Not sure what's the cleanest solution :confused:

qmonnet:
I couldn't find a simple way to mark the jobs with the `stable` toolchain as required. I couldn't find an easy way to mark the individual jobs as required, either: as their names contain the toolchain name obtained from the matrix, I don't think we have a solution to mark them as required and have the requirement satisfied when the jobs are skipped (and the matrix not generated).

So instead, I re-added the job summaries that we had before, so we can mark these as required. We don't get the distinction between the toolchains, but we could expand the summary to retrieve some output specific to each job if we really want to. That doesn't look super clean, but should work.

daniel-noland (approved):
LGTM

Issue #142, opened 2024-12-06
ci: bump.yml workflow failing: 


qmonnet:
The bump.yml workflow is still failing:

https://github.com/githedgehog/dataplane/actions/runs/12192389861/job/34012935400#step:9:9

```
/home/runner/work/_temp/7e997195-83d5-48ad-a437-f4358b8c7ac6.sh: line 1: ./scripts/update-versions.sh: No such file or directory
```

Indeed, there's no such script in the repo. Should it be `./scripts/bump.sh`?

PR #148, opened 2025-01-13
ci: Fix bump.yml by calling the right script

qmonnet:
/home/runner/work/_temp/7e997195-83d5-48ad-a437-f4358b8c7ac6.sh: line 1: ./scripts/update-versions.sh: No such file or directory

I believe we need to call bump.sh instead.

Fixes: https://github.com/githedgehog/dataplane/issues/142


daniel-noland (approved):
LGTM

PR #150, opened 2025-01-13
ci: Fix step condition, to fix docs publishing

qmonnet:
Operator precedence matters... `!github.event_name == 'merge_group'` is likely different from `github.event_name != 'merge_group'`, and the condition to run the publish Action would fail to validate even on `push` events. Let's fix.

Fixes: #99


qmonnet:
Ci change only, and can't be worse than the current situation where the steps don't run. I want to take a look at the `push` run, merging without review.

qmonnet:
We fixed the workflow:

image

However, I still can't access the generated page, that I would expect to find at https://docs.dataplane.hedgehog.website/.

PR #156, opened 2025-01-14
[RFC] justfile: Remove references to dev-env container

qmonnet:
I don't remember for sure, but looking at the code, it seems to me that we no longer use (or build) the `dev-env` container, we just use the `compile-env` container nowadays?

If that's the case, we may want to merge this Pr, that removes references to `dev-env` in the repo's `justfile`. If indeed we no longer need this container, we should also remove the generated images from ghcr.io.


daniel-noland (approved):
Yup.  This was useful once but is no longer needed

PR #160, opened 2025-01-17
justfile: Re-add hugepage-related targets

qmonnet:
This is a partial revert from commit 95db98f0a706 ("justfile: Remove references to dev-env container"): Re-add the justfile targets for mounting, unmounting, and allocating hugepages. These targets would have been useful a couple of times over the last days.

Also add two justfile targets, setup-test-env and teardown-test-env, to call these hugepages-related private targets.


daniel-noland (approved):
LGTM

PR #165, opened 2025-01-23
Expand DPDK bindings

daniel-noland:


dpdk/src/dev.rs
```diff
@@ -239,9 +245,14 @@ impl DevConfig {
                 ..Default::default()
             },
             rxmode: rte_eth_rxmode {
-                // TODO: let user request rx offloads instead of just enabling all supported
-                // offloads.
-                // offloads: self.info.inner.rx_offload_capa,
+                mtu: 1515,
+                mq_mode: RTE_ETH_MQ_RX_RSS,
+                max_lro_pkt_size: 8192,
```
qmonnet, line 250: Would it make sense to make these parameters more visible (and easier to adjust) by defining consts at the top of the file, rather than “burry” them in the middle of the source code?

deny.toml
```diff
@@ -65,14 +63,17 @@ allow = [
   "QPL-1.0",
   "RPSL-1.0",
   "SISSL",
-  "Sleepycat",
   "SPL-1.0",
-  "Unlicense",
+  "Sleepycat",
   "UPL-1.0",
+  "Unicode-3.0",
+  "Unlicense",
```
qmonnet, line 70: What do we need these for? Might be worth providing the motivation in the commit log.

.github/workflows/bump.yml
```diff
@@ -41,10 +41,9 @@ jobs:
         uses: "actions/checkout@v4"
       - name: "install rust"
         uses: "dtolnay/rust-toolchain@stable"
-      - uses: "cargo-bins/cargo-binstall@main"
       - name: "install cargo-deny"
         run: |
-          cargo binstall --no-confirm cargo-deny
+          cargo install cargo-deny
```
qmonnet, line 46: Note: We can ask GitHub runners to cache specific items between runs, if this were helpful. Note sure if that would solve the problem for `cargo binstall`.

qmonnet (commented):
Thanks, great work! Good thing to drop `scratch` at last.

My major comment is that your third commit (`Expand DPDK bindings`) is very big, and it's hard to make anything of it in the review, there are just too many different things happening. I glanced over it but can't really say I've been over all the changes in details.

Another note would be to consider opening follow-up Issues, once this PR is merged, to track the most important “TODO”s from your code. You have several ones, and if any of them is something we seriously need to follow up with, it's easier to track as Issues than as comments in the code.

The rest looks good. I've got some minor comments and suggestions inline below.

dpdk/src/dev.rs
daniel-noland: That is a fair point.  I think we should just expose them as command line params tho.

deny.toml
daniel-noland: done

.github/workflows/bump.yml
daniel-noland: Fortunately most of the builds are under 30 seconds so maybe not worth fiddling with

dataplane/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    // placeholder
+}
```
qmonnet, line 3: Even if it's an empty function, can we have a copyright header + SPDX tag on the file, please?

dpdk/src/dev.rs
qmonnet: Command-line param sounds good too.

If you don't intend to address it in this PR, can you please at least add a TODO comment so we can find this again and follow up?

dpdk/src/queue/rx.rs
```diff
@@ -50,25 +53,38 @@ pub struct RxQueueConfig {
...
-    Unexpected(errno::Errno),
-    /// The socket preference setting did not resolve a known socket.
-    InvalidSocket(errno::Errno),
+    #[error("The device has been removed")]
+    DeviceRemoved(Errno),
+    #[error("Invalid arguments were passed to the receive queue configuration")]
+    InvalidArgument(Errno),
+    #[error("Memory allocation failed")]
+    NoMemory(Errno),
+    #[error("An unexpected (i.e. undocumented) error occurred")]
```
qmonnet, line 71: Nit: or just “An undocumented error occurred”.

Also it would seem more natural to move that one to the end of the list.

dpdk/src/lcore.rs
qmonnet, line 1: Missing SPDX + copyright header

dpdk/src/ring/mod.rs
qmonnet, line 1: Header please

dpdk/src/ring/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    #[derive(thiserror::Error, Debug)]
+    pub enum InvalidArgument {
+        #[error("size must be a power of two ({size} given)", size=.0.size)]
+        SizeNotPowerOfTwo(Params),
+        #[error("ring name must be ASCII")]
+        NameNotAscii(Params),
+        #[error("Only alphanumeric ring names are supported (may contain -, _, and .)")]
+        IllegalCharacters(Params),
+        #[error("Name too long (max is 127 ASCII characters)")]
```
qmonnet, line 139: Nit: Lowercase/uppercase consistency (I personally prefer using an upper case here, but mostly it would be nice to remain consistent between the different error messages).

Same below in `RingCreateErr`.

qmonnet (changes requested):
Thanks a lot for reworking the commits, this is more logical and much more manageable. I'm not gonna pretend I followed everything in the lcore or ring files, but it seems mostly in order to me.

One thing: Please make sure to add the copyright + SPDX headers to all the new Rust files.

dpdk/src/ring/mod.rs
daniel-noland: Good point.  The rust convention is lower case but my hands tend to automatically capitalize the first letter when I type a sentence.  Will fix

dpdk/src/lcore.rs
daniel-noland: Will fix these

dpdk/src/queue/rx.rs
daniel-noland: Fair enough

dataplane/src/main.rs
daniel-noland: will do

daniel-noland: Resolved

dpdk/src/queue/rx.rs
daniel-noland: Resolved I think

dpdk/src/lcore.rs
daniel-noland: resolved

dpdk/src/ring/mod.rs
daniel-noland: resolved

dpdk/src/ring/mod.rs
daniel-noland: resolved

dpdk/src/dev.rs
qmonnet: > If you don't intend to address it in this PR, can you please at least add a TODO comment so we can find this again and follow up?

Please address this.

qmonnet (changes requested):

dpdk/src/queue/rx.rs
qmonnet: You kept both “unexpected” and “undocumented”, which I think is a lot for an error message. But doesn't matter much, thanks for moving it down the list.

There's another occurrence of an enum where the `unexpected` error could be moved to the end (`enum ConfigFailure` in tx.rs). I leave it to you to see if you want to move it as well.

daniel-noland: will fix

qmonnet (approved):
Looks good to me, thanks a lot for this work!

PR #167, opened 2025-01-24
Packet parser part 1

daniel-noland:
Introduce part 1 of the packet parsing and manipulation framework.


Cargo.toml
```diff
@@ -8,6 +8,13 @@ members = [
 	"errno",
 	"net",
 ]
+
+default-members = [
```
mvachhar, line 12: Is this type of build even valuable at this point?  We aren't building dataplane which is the main thing we want to build, no dpdk, etc.

justfile
```diff
@@ -473,3 +473,19 @@ mdbook *args="build":
...
+    {{ _just_debuggable_ }}
+    set -eu -o pipefail
+    if [ {{_clean}} != "clean" ]; then
+      >&2 echo "can not build-sweep with dirty branch (would risk data loss)"
+      exit 1
+    fi
+    # Get all commits since {{start}}, in chronological order
+    while read -r commit; do
+      git checkout "${commit}" || exit 1
+      { just debug=true cargo +{{rust}} build --locked --profile=dev --target=x86_64-unknown-linux-gnu; } || exit 1
```
mvachhar, line 490: I'm not convinced that having commit-by-commit checking of build is a good idea, but it is fine to leave it for now, we can drop the requirement if it becomes a problem.  Build is probably fine though, unit test is where it gets hairy.

scripts/rust.env
```diff
@@ -12,11 +13,11 @@ COVERAGE="-C instrument-coverage"
...
 TARGET_CPU_DEBUG="-C target-cpu=generic"
 TARGET_CPU_RELEASE="-C target-cpu=x86-64-v4"
 
-COMMON="${LINKER} ${RELRO} ${CRT} ${DEBUG}"
+COMMON="${LINKER} ${RELRO} ${DEBUG}"
 RUSTFLAGS_DEBUG="${COMMON} ${OPTIMIZE_OFF} ${TARGET_CPU_DEBUG} ${COVERAGE}"
-RUSTFLAGS_RELEASE="${COMMON} ${OPTIMIZE_ON} ${TARGET_CPU_RELEASE}"
-RUSTFLAGS_FUZZ="${COMMON} ${OPTIMIZE_FUZZ} ${TARGET_CPU_RELEASE}"
+RUSTFLAGS_RELEASE="${COMMON} ${OPTIMIZE_ON} ${TARGET_CPU_RELEASE} ${CRT_STATIC}"
+RUSTFLAGS_FUZZ="${COMMON} ${OPTIMIZE_FUZZ} ${TARGET_CPU_RELEASE} ${CRT_DYNAMIC}"
```
mvachhar, line 23: Does this mean we cannot fuzz test release builds?  Does this matter?

Cargo.toml
daniel-noland: yes it is useful for unit testing and aarch64 environments

scripts/rust.env
mvachhar: Looks like there are two choices here, live with this for the sanitizer or put a dynamic libc into our container.  For now, let's just live with this, and we can fix if it is an issue.

net/src/vxlan/vni.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used, clippy::expect_used)] // valid in test code
+mod test {
+    use super::*;
+
+    #[test]
+    fn zero_is_not_a_legal_vni() {
+        assert_eq!(Vni::new_checked(0).unwrap_err(), InvalidVni::ReservedZero);
```
mvachhar, line 139: We should not generate (de novo) a VNI of 0, but we should accept it if the RFC says it is a legal VNI.  Do not change this, create an issue for it.

net/src/parse.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn size(&self) -> NonZero<usize>;
+    /// Write a data structure (e.g., a packet header) to a buffer.
+    ///
+    /// Returns the number of bytes written in the event of success.
+    ///
+    /// # Errors
+    ///
+    /// Will return an error if there is not enough space in the buffer
+    /// or if serialization fails from some other (implementation-dependent) reason.
+    fn write(&self, buf: &mut [u8]) -> Result<NonZero<usize>, DeParseError<Self::Error>>;
```
mvachhar, line 31: rename this to deparse to avoid trait function conflicts, just like we do in Parse.  serde uses serialize, so we don't want that, and tons of people have read.

net/src/packet.rs
```diff
@@ -0,0 +1,251 @@
...
+        };
+        let transport = match self.transport {
+            None => 0,
+            Some(ref t) => t.size().get(),
+        };
+        NonZero::new(eth + vlan + net + transport).unwrap_or_else(|| unreachable!())
+    }
+
+    fn write(&self, buf: &mut [u8]) -> Result<NonZero<usize>, DeParseError<Self::Error>> {
+        // TODO(blocking): Deal with ip{v4,v6} extensions
```
mvachhar, line 216: Why is there no abstraction here like there is for `step` on parse?  Please create an issue to fix this and address that issue right after the PR that has a main that parses a packet.

net/src/ipv4/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+    use etherparse::err::ipv4::{HeaderError, HeaderSliceError};
+
+    #[test]
+    fn parse_back() {
+        bolero::check!().with_arbitrary().for_each(|header: &Ipv4| {
+            let mut buffer = [0u8; Ipv4::MIN_LEN.get()];
+            let bytes_written = header
+                .write(&mut buffer)
+                .unwrap_or_else(|e| unreachable!("{e:?}"));
+            assert!(bytes_written <= Ipv4::MIN_LEN);
```
mvachhar, line 410: Shouldn't this assertion just be assert_eq?

mvachhar (changes requested):

Cargo.toml
```diff
@@ -8,6 +8,13 @@ members = [
...
+	"dpdk-sysroot-helper",
+	"errno",
+	"net",
+]
```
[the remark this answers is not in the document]
daniel-noland: Truth.  Will make an edit here

net/src/ipv4/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+#[repr(transparent)]
+#[error("ttl is already zero")]
+pub struct TtlAlreadyZero;
+
+impl Parse for Ipv4 {
+    type Error = etherparse::err::ipv4::HeaderSliceError;
+
+    fn parse(buf: &[u8]) -> Result<(Self, NonZero<usize>), ParseError<Self::Error>> {
+        let (inner, rest) = Ipv4Header::from_slice(buf).map_err(ParseError::Invalid)?;
+        assert!(
```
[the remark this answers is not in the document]
daniel-noland: Yes.  I will use the [no_panic](https://docs.rs/no-panic/latest/no_panic/) crate to do this in a future PR

net/src/ipv4/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+    Udp(Udp),
+    Icmp4(Icmp4),
+    IpAuth(IpAuth),
+}
+
+impl Step for Ipv4 {
+    type Next = Ipv4Next;
+
+    fn step(&self, cursor: &mut Reader) -> Option<Self::Next> {
+        match self.inner.protocol {
```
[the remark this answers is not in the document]
daniel-noland: Likely the best thing is to turn this into a tuple struct.  Will give that a try now

net/src/ipv4/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+    fn parse(buf: &[u8]) -> Result<(Self, NonZero<usize>), ParseError<Self::Error>> {
+        let (inner, rest) = Ipv4Header::from_slice(buf).map_err(ParseError::Invalid)?;
+        assert!(
+            rest.len() < buf.len(),
+            "rest.len() >= buf.len() ({rest} >= {buf})",
+            rest = rest.len(),
+            buf = buf.len()
+        );
+        let consumed = NonZero::new(buf.len() - rest.len()).ok_or_else(|| unreachable!())?;
+        Ok((Self { inner }, consumed))
```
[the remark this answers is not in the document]
daniel-noland: Agreed.  Will fix

net/src/ipv4/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+
+impl Step for Ipv4 {
+    type Next = Ipv4Next;
+
+    fn step(&self, cursor: &mut Reader) -> Option<Self::Next> {
+        match self.inner.protocol {
+            IpNumber::TCP => cursor
+                .parse::<Tcp>()
+                .map_err(|e| {
+                    debug!("failed to parse tcp: {e:?}");
```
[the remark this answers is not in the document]
daniel-noland: Beyond the scope of this commit.  In future we should use the [metrics](https://docs.rs/metrics/latest/metrics/) crate for this since it integrates nicely with tracing et al

net/src/packet.rs
```diff
@@ -0,0 +1,251 @@
...
+    Ipv6(Ipv6),
+    Tcp(Tcp),
+    Udp(Udp),
+    Icmp4(Icmp4),
+    Icmp6(Icmp6),
+    IpAuth(IpAuth),
+    IpV6Ext(Ipv6Ext), // TODO: break out nested enum.  Nesting is counter productive here
+}
+
+impl Step for Header {
```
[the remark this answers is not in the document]
daniel-noland: Fair enough.  I will work on a better name in follow up.  Perhaps `ParseNext` or `ParsePayload`.  `ParsePayload` is attractive to me.

Logically there should be a `Parse` and a `ParseWith`, so `ParsePayload` and `ParsePayloadWith` would have symmetry which may help with understanding the structure of the code

net/src/parse.rs
daniel-noland: Yup.  Will rafactor

net/src/packet.rs
daniel-noland: This is a good point and is arguably a design miss.  I will think about good ways to fix this

net/src/ipv4/mod.rs
daniel-noland: Yup.  I think it was `assert!(bytes_written <= Ipv4::MAX_LEN)` in a prior version of the test and then the test was refactored and this happened.  Will fix

net/src/vxlan/vni.rs
daniel-noland: Having thought about this more, I'm not sure it makes sense to accept a VNI you can't respond to.  The VXLAN tunnel would be fundamentally broken if only one side of it worked.  That said, I agree with you in principal.  I'm happy to make a ticket where we can discuss tho

mvachhar (approved):

PR #187, opened 2025-01-28
README.md: Add VS Code setup instructions for rust-analyzer

mvachhar:
`rust-analyzer` will run `cargo` (and thus `rustc`, etc.) from `PATH`. But, `dataplane/compile-env` has its own `cargo` and `rustc` which causes issues. The fix is to tell `rust-analyzer` to use `cargo` from `dataplane/compile-env`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,24 @@ just cargo test --profile=release
 > After that it simply calls `cargo build`.
 > Normally we would include those kinds of setting in `Cargo.toml` but `cargo` can not currently express all the `RUSTFLAGS` we are using (thus the `just` wrapper).
 
+## VSCode Setup
+
+Because this repository uses a custom sysroot with custom libraries and binaries, you need to make sure rust-analyzer is aware of the sysroot and runs the correct cargo binary.
+To do this, add the following to your `.vscode/settings.json` file:
+
+```json
+{
+    "rust-analyzer.server.extraEnv": {
+        "CARGO": "<absolute path to dataplane directory>/bin/cargo"
+    }
+}
+```
+
+> [!NOTE]
+> Please submit a PR if you have a way to avoid the absolute path.
+> `${workspaceRoot}` and `${workspaceFolder}` won't work since rust-analyzer has a custom function that implements env var substitution in `extraEnv`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -128,6 +128,24 @@ just cargo test --profile=release
...
+```json
+{
+    "rust-analyzer.server.extraEnv": {
+        "CARGO": "<absolute path to dataplane directory>/bin/cargo"
+    }
+}
+```
+
+> [!NOTE]
+> Please submit a PR if you have a way to avoid the aboslute path.
```
Qeole, line 145:
```suggestion
> Please submit a PR if you have a way to avoid the absolute path.
```

README.md
```diff
@@ -128,6 +128,24 @@ just cargo test --profile=release
...
+> `${workspaceRoot}` and `${workspaceFolder}` won't work since rust-analyzer has a custom function that implements env var substitution in `extraEnv`.
+> `${env:xxx}` susbstitutions only work if the variable is set in `extraEnv` itself.
```
Qeole, line 147: Have you tried something like `$PWD/bin/cargo`? I doubt it works, but just in case. I don't have any other idea to address this off the top of my head.

Qeole (changes requested):
Looks good, thanks! Just one typo to fix

README.md
qmonnet, line 145:
```suggestion
> Please submit a PR if you have a way to avoid the absolute path.
```

README.md
qmonnet, line 147: Have you tried something like `$PWD/bin/cargo`? I doubt it works, but just in case. I don't have any other idea to address this off the top of my head.

qmonnet (changes requested):
Looks good, thanks! Just one typo to fix

qmonnet (approved):
Looks good, thank you!

README.md
mvachhar: Unfortunately, the `rust-analyzer` extension use custom code for env vars that doesn't have `$PWD`

README.md
mvachhar: Done.

PR #188, opened 2025-01-31
Pr/daniel noland/parser part 2

daniel-noland:
This PR significantly expands the ability to manipulate packets.

We are not quite to a stable(ish) API yet, but we are getting there.

### DIFF
--- a/dpdk/src/mem.rs
+++ b/dpdk/src/mem.rs
@@ -21,9 +21,11 @@ use core::slice::from_raw_parts_mut;
 use errno::Errno;
 use tracing::{error, info, warn};
 
+use dpdk_sys::{rte_pktmbuf_adj, rte_pktmbuf_prepend};
 // unfortunately, we need the standard library to swap allocators
 use std::alloc::System;
 use std::ffi::CString;
+use std::num::NonZero;
 
 /// DPDK memory manager
 #[repr(transparent)]
@@ -494,7 +496,45 @@ impl Mbuf {
             )
         }
     }
+
+    #[tracing::instrument(level = "trace")]
+    pub fn prepend(&mut self, len: NonZero<u16>) -> Result<&mut [u8], NotEnoughHeadRoom> {
+        let val = unsafe { rte_pktmbuf_prepend(self.raw.as_mut(), len.get()) };
+        NonNull::new(val)
+            .map(|ptr| unsafe {
+                from_raw_parts_mut(
+                    ptr.as_ptr().cast::<u8>(),
+                    self.raw.as_ref().annon2.annon1.data_len as usize,
+                )
+            })
+            .ok_or(NotEnoughHeadRoom)
+    }
+
+    #[tracing::instrument(level = "trace")]
+    pub fn adjust(&mut self, len: NonZero<u16>) -> Result<&mut [u8], MbufNotLongEnough> {
+        let val = unsafe { rte_pktmbuf_adj(self.raw.as_mut(), len.get()) };
+        NonNull::new(val)
+            .map(|ptr| unsafe {
+                from_raw_parts_mut(
+                    ptr.as_ptr().cast::<u8>(),
+                    self.raw.as_ref().annon2.annon1.data_len as usize,
+                )
+            })
+            .ok_or(MbufNotLongEnough)
+    }
 }
+
+#[non_exhaustive]
+#[repr(transparent)]
+#[derive(Debug, thiserror::Error)]
+#[error("Not enough head room in mbuf")]
+pub struct NotEnoughHeadRoom;
+
+#[non_exhaustive]
+#[repr(transparent)]
+#[derive(Debug, thiserror::Error)]
+#[error("Mbuf not long enough to remove required number of bytes from the start")]
+pub struct MbufNotLongEnough;
 
 /// A global memory allocator for DPDK
 #[non_exhaustive]
--- a/net/src/eth/mac.rs
+++ b/net/src/eth/mac.rs
@@ -3,7 +3,6 @@
 
 //! Mac address type and logic.
 
-use crate::eth::{DestinationMacAddressError, SourceMacAddressError};
 use std::fmt::Display;
 
 /// A [MAC Address] type.
@@ -13,7 +12,7 @@ use std::fmt::Display;
 ///
 /// [MAC Address]: https://en.wikipedia.org/wiki/MAC_address
 #[repr(transparent)]
-#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(bolero::TypeGenerator))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Mac(pub [u8; 6]);
@@ -100,18 +99,32 @@ impl Mac {
         (bytes[0..5] == [0x01, 0x80, 0xc2, 0x00, 0x00]) && (bytes[5] & 0x0f == bytes[5])
     }
 
-    /// Returns true iff the [`Mac`] is a legal source `Mac`.
+    /// Returns `Ok(())` iff the [`Mac`] is a legal source `Mac`.
+    ///
+    /// # Errors
     ///
     /// Multicast and zero are not legal source [`Mac`].
-    #[must_use]
-    pub fn is_valid_src(&self) -> bool {
-        !self.is_zero() && !self.is_multicast()
+    pub fn valid_src(&self) -> Result<(), SourceMacAddressError> {
+        if self.is_zero() {
+            Err(SourceMacAddressError::ZeroSource(*self))
+        } else if self.is_multicast() {
+            Err(SourceMacAddressError::MulticastSource(*self))
+        } else {
+            Ok(())
+        }
     }
 
-    /// Returns true iff the [`Mac`] is a legal destination [`Mac`].
-    #[must_use]
-    pub fn is_valid_dst(&self) -> bool {
-        self.is_valid()
+    /// Returns `Ok(())` iff the [`Mac`] is a legal destination [`Mac`].
+    ///
+    /// # Errors
+    ///
+    /// Zero is not a legal destination [`Mac`].
+    pub fn valid_dst(&self) -> Result<(), DestinationMacAddressError> {
+        if self.is_zero() {
+            Err(DestinationMacAddressError::ZeroDestination(*self))
+        } else {
+            Ok(())
+        }
     }
 
     /// Return true iff the [`Mac`] is not zero.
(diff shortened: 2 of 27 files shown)


net/src/eth/mod.rs
```diff
@@ -68,106 +49,57 @@ impl Eth {
     /// # Errors
     ///
     /// This function will return an error if the specified source or dest [Mac] are invalid.
-    pub fn new(source: Mac, destination: Mac, ether_type: EthType) -> Result<Eth, EthError> {
-        let mut header = Eth(Ethernet2Header {
-            source: source.0,
-            destination: destination.0,
+    pub fn new(
```
Fredi-raspall, line 52: I'm a bit confused here. According to the documentation line, this function may return an error. However, it seems to unconditionally return Ok(), without any of the member initialization being able to fail.

net/src/eth/mod.rs
```diff
@@ -68,106 +49,57 @@ impl Eth {
...
-        Ok(header)
+        }))
     }
 
-    /// Get the source [Mac] of the header.
+    /// Get the source [`Mac`] of the header.
     #[must_use]
-    pub fn source(&self) -> Mac {
-        Mac(self.0.source)
+    pub fn source(&self) -> SourceMac {
```
Fredi-raspall, line 66: Similarly, here, if an Eth can only exist with valid addresses (due to the constructor constraints), why is it needed the unsafe and the _unchecked(). The same applies for the destination() getter. Am I overlooking at something?

net/src/eth/mod.rs
```diff
@@ -191,11 +123,15 @@ impl Parse for Eth {
         );
         let consumed = NonZero::new(buf.len() - rest.len()).ok_or_else(|| unreachable!())?;
         let new = Self(inner);
-        // integrity check for ethernet header (slightly hacky)
-        SourceMac::new(new.source())
-            .map_err(|e| ParseError::Invalid(EthError::InvalidSource(e)))?;
-        DestinationMac::new(new.destination())
+        // integrity check for ethernet header
```
Fredi-raspall, line 126: Ok, now I seem to understand what you've done. You create an Eth from Ethernet2Header, unchecked, upfront and then validate the resulting Eth. According to the documentation, I'd have expected to work the other way: you get src/dst/ethtype from Ethernet2Header in a "fallible" way and only if all checks succeed you build an Eth. If what I said is correct, I'd amend the comments.

net/src/udp/mod.rs
```diff
@@ -3,17 +3,29 @@
...
 
 /// A UDP header.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Udp(UdpHeader);
 
+/// A UDP encapsulation.
+///
+/// At this point we only support VXLAN, but Geneve and others can be added as needed.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Encap {
```
Fredi-raspall, line 25: I think I would rename this to UdpEncap since it seems to be specific of UDP only.... just in case we have to support other non-udp encapsulations (say GRE), or 6to4's and the like.

net/src/packet.rs
```diff
@@ -15,22 +16,25 @@ use crate::parse::{
...
 use std::num::NonZero;
 use tracing::debug;
 
 const MAX_VLANS: usize = 4;
 const MAX_NET_EXTENSIONS: usize = 2;
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+// TODO: remove `pub` from all fields
+#[derive(Debug)]
 pub struct Packet {
```
Fredi-raspall, line 31: I'd reorder the fields here to reflect the order in which they are populated: eth/vlan/net ....
It's mostly cosmetic, yes.
Also, for the encap, would it make sense to nest it inside transport?

Fredi-raspall (changes requested):
Looks very good to me Daniel. Great job!

net/src/udp/mod.rs
Fredi-raspall: Now, I am seeing that Packet has a member encap: Option<Encap>. An alternative to rename it, and unless you nest it in Udp (see comments in packet.rs), would be to move this definition out of the udp module so that it can include a bunch of other encapsulations.

net/src/ipv4/dscp.rs
qmonnet, line 1: Missing SPDX/copyright

(Note: #229 should automate this check)

net/src/packet.rs
```diff
@@ -248,6 +260,199 @@ impl DeParse for Packet {
...
     }
 }
+
+#[derive(Debug, thiserror::Error)]
+#[error("Packet already has as many VLAN headers as parser can support (max is {MAX_VLANS})")]
+pub struct TooManyVlans;
+
+impl Packet {
+    /// Create a new packet with the supplied `Eth` header.
+    fn new(eth: Eth) -> Packet {
```
qmonnet, line 284: Nice, I've been missing this for my tests with NAT :+1:

net/src/packet.rs
```diff
@@ -248,6 +260,199 @@ impl DeParse for Packet {
...
+        Packet {
+            eth,
+            vlan: ArrayVec::default(),
+            net: None,
```
qmonnet, line 288: Is there a way we can set `net` for tests in other modules? I didn't find an easy way without making the constructor for `Ipv4`, or its inner `Ipv4header`, public (e.g. `pub struct Ipv4(pub Ipv4Header);`).

dpdk/src/mem.rs
```diff
@@ -22,8 +22,10 @@ use errno::Errno;
 use tracing::{error, info, warn};
 
 // unfortunately, we need the standard library to swap allocators
+use dpdk_sys::{rte_pktmbuf_adj, rte_pktmbuf_prepend};
 use std::alloc::System;
```
qmonnet, line 26: Nit: Keep comment above the relevant `use`?

qmonnet (changes requested):
Please find some very minor comments below. Other than that, I concur with Fredi, excellent work! Some commits are doing a tad too many things at once to my taste, but nothing unbearable, I appreciate the effort to get to something pretty clean :innocent:

qmonnet:
Also note the merge conflict on `net/src/eth/mac.rs`, please rebase the PR

net/src/udp/mod.rs
daniel-noland: This is fair.  Will rename.  TCP encap is rare but does exist

net/src/packet.rs
daniel-noland: I debated a lot about encap being inside transport or not.

In the end I suggest that we keep it outside of transport because putting it in the transport variant increases the size of the `Udp` (et al) object and that is a very useful struct on its own.

Fredi-raspall: How about the ordering?

net/src/eth/mod.rs
daniel-noland: Yup :)

I go back and forth on this line of thinking quite a bit.

Should I build "safe" functions out of `unsafe` functions, or make the `unsafe` functions as degenerate subsets of the safe versions?
I think each method has its own advantages depending on what you are doing.

The best way to explain it I have ever seen is to compare the code to an arch (as in the shape / architecture concept).

The arch is famously stable and reliable as an architectural component which is central to the design of many buildings.
On the other hand, it will crumble immediately if you remove the keystone.
It also requires temporary scaffolding to construct an arch because it isn't stable or "safe" before the keystone is in place.

In the same way, `unsafe` functions can be (and regularly are) sound components of "safe" functions.

Still, it is often easier to construct sound functions out of functions which are individually known to be sound rather than "be really careful" methods which skip soundness checks.

There is also the issue of populating the API with `unsafe` methods because they can be dangerously convenient.
Even if used correctly, `unsafe` code is typically harder to refactor because the act of pushing soundness requirements to the caller means that changing the function requires extra consideration at every call site of that function.

I guess this is a long-winded way of saying, "yeah, I can see it either way" :shrug:

net/src/packet.rs
daniel-noland: That is also a fair point.  There may be some very tiny advantage in terms of locality in organizing the structure to include the most likely to be used fields close together, but that is pure speculation and does not really justify the deviation from the OSI layer mental model (at least not in the absence of strong performance numbers)

net/src/eth/mac.rs
```diff
@@ -155,53 +174,83 @@ impl SourceMac {
...
     }
 }
 
+/// Errors which can occur while setting the source [`Mac`] of a [`Packet`]
+///
+/// [`Packet`]: crate::packet::Packet
+#[derive(Debug, thiserror::Error)]
+pub enum SourceMacAddressError {
+    /// Multicast [`Mac`]s are not legal as a source [`Mac`]
+    #[error("invalid source mac address: multicast macs are illegal as source macs")]
```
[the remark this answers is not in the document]
daniel-noland: fair enough.  I think this makes sense in the context of a user facing error.  Otherwise I tend to think it is best to follow the casing used in the code

net/src/ipv4/dscp.rs
daniel-noland: Yup, will fix

dpdk/src/mem.rs
daniel-noland: good catch.

net/src/packet.rs
daniel-noland: Yeah, packet generation is always tricky (especially before the whole API is reasonably stable).

In theory you should be able to use the `TypeGenerator` implementation to spit out an infinite list of `Ipv4` or `Net` objects.  That said, I need to follow up with a `TypeGenerator` for `Packet` to make this easy and sensible

daniel-noland:
will take a look on the merge conflict.  The strange thing is that I _just__ rebased the PR and didn't see any changes.  Something is goofy somewhere

net/src/eth/mod.rs
daniel-noland: yeah, this is the result of an automated refactor.  Good catch

net/src/eth/mod.rs
```diff
@@ -68,106 +49,57 @@ impl Eth {
...
-    /// Set the destination [Mac] of the header.
-    ///
-    /// # Safety
-    ///
-    /// This method does not check that the [Mac] is a valid dest [Mac].
-    pub fn set_destination_unchecked(&mut self, destination: Mac) -> &mut Eth {
-        debug_assert!(destination.is_valid_dst());
-        self.0.destination = destination.0;
+    /// Set the destination [`Mac`] of the ethernet header.
+    pub fn set_destination(&mut self, destination: DestinationMac) -> &mut Self {
```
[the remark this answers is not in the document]
daniel-noland: Correct.

I think the logic is sound here, but as you pointed out, this API will need some love to make it more consistent overall.

I think we should reserve that for part 3 tho (otherwise we will end up with an even more complex rebase of other outstanding work).

net/src/packet.rs
daniel-noland: That makes sense.

Hopefull the follow on PRs will make this passably ergonomic.

I have plans to use tools like bolero and smoltcp to make this much more test friendly in the future but for now we have packet generation as something of a pain point

qmonnet (approved):
Fix-ups look good. Thanks for addressing the feedback!

daniel-noland:
Nice.  Will rebase and autosquash now

PR #198, opened 2025-02-03
.gitignore: Add artifact directory

qmonnet:
This directory is generated when packaging the dataplane executable as a container. Tell Git to ignore it.


daniel-noland (approved):
Yup LGTM

PR #199, opened 2025-02-03
ci: Add test for building container with sterile build artifact

qmonnet:
We have the justfile/Dockerfile infra to build a container containing only the dataplane executable resulting from a sterile build, by running `just build-container`; but we have nothing to test it in CI. Incidentally, as of this writing, it is broken. Add a CI test in which we build the container.

### DIFF
--- a/.github/workflows/sterile.yml
+++ b/.github/workflows/sterile.yml
@@ -152,10 +152,16 @@ jobs:
           cargo install cargo-deny
       - run: |
           just debug=true cargo deny check
+      # Default to "dev" profile
+      - run: |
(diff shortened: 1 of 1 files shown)


qmonnet:
Hmm this shouldn't be required, the `Push containers` job should in fact handle this verification already. Not sure how the current bug fell through :thinking:

qmonnet:
Ah, we're pushing containers with profiles `debug` and `release`, but the issue occurs when we keep the default profile `dev`.

qmonnet:
workflow fails, as it's currently supposed to do

qmonnet:
We'll rename the `dev` profile to `debug`, see #200. We already have tests in place for building the container with the `debug` profile, so this PR won't be necessary.

PR #200, opened 2025-02-03
Change profile name from dev to debug

sergeymatov:


qmonnet:
@daniel-noland Was there a reason to use `dev` as the profile, or was it a typo? If we want to keep `dev` we need to change the file path instead, but looking into usual profile names I'm not sure `dev` is supposed to be a thing

justfile
```diff
@@ -37,7 +37,7 @@ hugepages_1g := "8"
 hugepages_2m := "1024"
 _just_debuggable_ := if debug == "true" { "set -x" } else { "" }
 target := "x86_64-unknown-linux-gnu"
-profile := "dev"
+profile := "debug"
```
daniel-noland, line 40: This is a fine way to change the default profile, but I think we need a little more to resolve the root issue here.

We will also need the changes currently part of #201 as well as a test in CI to confirm that our build artifacts are statically linked

mvachhar:
So what is the fate of this PR?  Are we going ot merge after #201 or close and incorporate this into #201?

daniel-noland:
I'm fine if we want to rebase this on top of #201.  I don't think it makes sense to merge as is.

It requires [this commit](https://github.com/githedgehog/dataplane/pull/201/commits/1a69f519aa1a48514035e3c9bef722cab3d509b1) to make sense to me

daniel-noland (changes requested):
See comments above

sergeymatov:
> See comments above

Done

daniel-noland (approved):
LGTM

PR #201, opened 2025-02-03
Basic dataplane

daniel-noland:
Behold, a  basic main function.

I tacked on the build fix from this monring as that is also required to make this do anything useful.

We will need to adjust the PCI address in the arguments to make this work on the dataplane test machine


dpdk/src/dev.rs
```diff
@@ -226,7 +226,7 @@ pub struct DevConfig {
     /// Rework this bad idea.
```
mvachhar, line 226: The commit message has a typo `offloas` instead of `offloads`.

mvachhar (changes requested):

dpdk/src/dev.rs
daniel-noland: sorted

mvachhar (approved):

dataplane/src/main.rs
```diff
@@ -1,6 +1,120 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright Open Network Fabric Authors
+
```
qmonnet, line 2: @daniel-noland I guess this line was changed by mistake, please restore the copyright notice.

dataplane/src/main.rs
```diff
@@ -1,6 +1,120 @@
...
+        "2",
+        "--lcores",
+        "2-4",
+        "--in-memory",
+        "--allow",
+        "0000:01:00.0,dv_flow_en=1",
+        "--huge-worker-stack=8192",
+        "--socket-mem=8192,0,0,0",
+        "--no-telemetry",
+        "--iova-mode=va",
```
qmonnet, line 41: We plan to select some of these through some configuration mechanism later, right?

qmonnet (commented):
Nice work!
Please make sure to fix the copyright header in main.rs

dataplane/src/main.rs
qmonnet: Addressed in #214

PR #202, opened 2025-02-04
Remove dev coverage reports

daniel-noland:
Coverage reports are useful, but not by default.

This setting is causing the creation of .profraw files on every test run and they just are not helpful yet.


qmonnet (approved):
Thanks

Issue #176, opened 2025-01-27
Import NAT code into dataplane repo


qmonnet:
Our code for NAT is currently in a separate, private repository.

There's no good reason to separate it, it was simply to make it easier to iterate for the early stage of the code.

There's no good reason to keep it private, although we must make sure that headers with licensing and copyright attribution are present before moving the files.

We need to clean up (?) the other repo and import all code into `dataplane` at some point.

PR #206, opened 2025-02-04
Import NAT code

qmonnet:
- **nat: Import NAT draft code**
- **nat: Replace deprecated serde_yaml with serde_yml**
- **nat: Fix warnings about dead code**
- **nat: Use struct Vni instead of u32, for VNIs**
- **nat: Make tests work**

Not fully functional yet.

Imported from the separate, private repo [githedgehog/tmp-nat](https://github.com/githedgehog/tmp-nat/). There is no attempt to complete or improve the existing code in this PR, only to merge it into the repo while keeping `cargo build` and `cargo test` functional.

Refer to individual commits for details.

Fixes: #176


qmonnet:
@daniel-noland The issue I face with CI here is that some crate (namely, `regex-automata`) is pulled twice, with different minor versions, and triggers an error in the `cargo deny` check. How would you recommend to solve this? We could skip the ban for this crate I guess, but is there a cleaner way to address the issue?

daniel-noland:
> @daniel-noland The issue I face with CI here is that some crate (namely, `regex-automata`) is pulled twice, with different minor versions, and triggers an error in the `cargo deny` check. How would you recommend to solve this? We could skip the ban for this crate I guess, but is there a cleaner way to address the issue?

Will pull this and look.

The resolver is  normally verry good about this but it does happen.  Can't say just yet

daniel-noland:
@qmonnet, I think I figured out a workable solution.  Suggested commit pushed but do let me know if you object

daniel-noland:
The root issue is here: https://github.com/tokio-rs/tracing/pull/3033

We should add an exception in the deny list for the moment and then adjust once tracing bumps to matchers 0.2

qmonnet:
Thanks Daniel :pray:

sergeymatov (approved):
lgtm. Don't forget to fix my TODOs later :)

qmonnet:
I rebased the PR to address the merge conflicts on `deny.toml` and `Cargo.lock`. No other change.

PR #207, opened 2025-02-04
.gitignore: Fix ignore pattern for artifacts

qmonnet:
We can have several artifacts, but the directory name is singular. Fix.

:facepalm:


qmonnet:
`.gitignore` trivial fix. I think this is good to merge without annoying Daniel for a review.

PR #208, opened 2025-02-04
Adjust "inner" naming convention to tuple struct convention

daniel-noland:
Feedback is that `inner` is a confusing name for these fields. `inner` was chosen on the grounds that there is an outer and an inner type (i.e. a transparent wrapper type which qualifies the wrapped inner type).

Still, if `.0` is easier for folks to think about then I won't argue.

This should have been done as part of commit 695633370144fd1e9d766974808217cf0a4b8942 (see https://github.com/githedgehog/dataplane/commit/695633370144fd1e9d766974808217cf0a4b8942) but I did not think about it at the time

### DIFF
--- a/net/src/eth/mod.rs
+++ b/net/src/eth/mod.rs
@@ -23,9 +23,7 @@ pub use contract::*;
 
 /// An ethernet header.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Eth {
-    inner: Ethernet2Header,
-}
+pub struct Eth(Ethernet2Header);
 
 /// An error which may occur in the event of an invalid ethernet header.
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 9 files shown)


mvachhar (commented):
.0 kind of sucks too, but it is more Rust-like than using inner.  If it is always the header, the name header makes a lot more sense, do we want to go with that instead?

daniel-noland:
> .0 kind of sucks too, but it is more Rust-like than using inner. If it is always the header, the name header makes a lot more sense, do we want to go with that instead?

My concern with that is that while type is a wrapper around the `ethertype` header, it is itself a header in the `net` sense.

Naming is always a trick with thin wrapper types like this unfortunately :shrug:

qmonnet (approved):
I don't mind too much either way. This is an internal convention; we change it in this PR, we can change it again in the future if we find something more suitable. I'm fine with `.0` for now, PR looks OK to me.

daniel-noland:
I see that @qmonnet approved the changes but I'm not sure we have come to a consensus here.  I'm not terribly concerned either way but we should keep the PRs moving

daniel-noland:
I suppose we should merge on consistency grounds.  We already agreed to this change for `Ipv4` so why not elsewhere

qmonnet:
You'll need to rebase, though: GitHub reports a merge conflict on `net/src/ipv6/mod.rs`

daniel-noland:
On it

PR #210, opened 2025-02-04
Use workspace feature rather than individual versions

daniel-noland:
This makes the build a bit faster and versions a little easier to track.


qmonnet (approved):
Nice, thank you

PR #211, opened 2025-02-04
Definitions for types and entities in the routing module, and first implementation of the control-plane interface

Fredi-raspall:
This is super WIP and lacks the routing part (coming soon).
Opening this PR to get early feedback as suggested by @mvachhar and because there's some overlap with the work of the interface manager w.r.t modeling interfaces and neigh cache. @daniel-noland let's go over this at some point.

I'll be massively squashing fixes....

This is ready for review now. There's a lot of stuff to be done, but it's a starting point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -571,9 +571,9 @@ checksum = "469fb0b9cefa57e3ef31275ee7cacb78f2fdca44e4765491884a2b119d4eb130"
 
 [[package]]
 name = "iptrie"
-version = "0.9.1"
+version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0b2688e40b354466a90b89a55f8ce68acd0022f7f73aabd9968128ceafa7e239"
+checksum = "db170f9cb5ebafc6762e7dd1051e17275e771294cd9f2d7efe66ae47b7ae01be"
 dependencies = [
  "ipnet",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,7 +72,7 @@ version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79947af37f4177cfead1110013d678905c37501914fba0efea834c3fe9a8d60c"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -83,7 +83,7 @@ checksum = "ca3534e77181a9cc07539ad51f2141fe32f6c3ffd4df76db8ad92346b003ae4e"
 dependencies = [
  "anstyle",
  "once_cell",
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -376,7 +376,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90eeab0aa92f3f9b4e87f258c72b139c207d251f9cbc1080a0086b86a8870dd3"
 dependencies = [
  "nix",
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -443,7 +443,7 @@ version = "0.0.1"
 [[package]]
 name = "dplane-rpc"
 version = "1.0.0"
-source = "git+https://github.com/githedgehog/dplane-rpc.git#d3547d38735804f27d88dfc29db0c1985c5871c5"
+source = "git+https://github.com/githedgehog/dplane-rpc.git?branch=master#d3547d38735804f27d88dfc29db0c1985c5871c5"
 dependencies = [
  "bytes",
  "cbindgen",
@@ -482,7 +482,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -690,6 +690,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b3b1c9bd4fe1f0f8b387f6eb9eb3b4a1aa26185e5750efb9140301703f62cd1b"
 dependencies = [
  "adler2",
+]
+
+[[package]]
+name = "mio"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2886843bf800fba2e3377cff24abf6379b4c4d5c6681eaf9ea5b0d15090450bd"
+dependencies = [
+ "libc",
+ "log",
+ "wasi 0.11.0+wasi-snapshot-preview1",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -956,11 +968,15 @@ checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 name = "routing"
 version = "0.1.0"
 dependencies = [
+ "bytes",
  "dplane-rpc",
  "ipnet",
  "iptrie",
+ "mio",
  "net",
  "thiserror",
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -985,7 +1001,7 @@ dependencies = [
  "errno 0.3.10",
  "libc",
  "linux-raw-sys",
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1116,7 +1132,7 @@ dependencies = [
  "getrandom 0.3.1",
  "once_cell",
  "rustix",
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1342,6 +1358,15 @@ name = "winapi-x86_64-pc-windows-gnu"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
+
+[[package]]
+name = "windows-sys"
+version = "0.52.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "282be5f36a8ce781fad8c8ae18fa3f9beff57ec1b52cb3de0789201425d9a33d"
+dependencies = [
+ "windows-targets",
+]
 
 [[package]]
 name = "windows-sys"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -751,6 +751,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 
 [[package]]
+name = "routing"
+version = "0.1.0"
+
+[[package]]
 name = "rustc-demangle"
 version = "0.1.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -753,6 +753,10 @@ checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 [[package]]
 name = "routing"
 version = "0.1.0"
+dependencies = [
+ "net",
+ "thiserror",
+]
 
 [[package]]
 name = "rustc-demangle"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,10 +37,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
 name = "anstyle"
 version = "1.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55cc3b69f167a1ef2e161439aa98aed94e6028e5f9a59be9a6ffb47aef1651f9"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b2d16507662817a6a20a9ea92df6652ee4f94f914589377d69f3b21bc5798a9"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79947af37f4177cfead1110013d678905c37501914fba0efea834c3fe9a8d60c"
+dependencies = [
+ "windows-sys",
+]
+
+[[package]]
+name = "anstyle-wincon"
+version = "3.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ca3534e77181a9cc07539ad51f2141fe32f6c3ffd4df76db8ad92346b003ae4e"
+dependencies = [
+ "anstyle",
+ "once_cell",
+ "windows-sys",
+]
 
 [[package]]
 name = "anyhow"
@@ -62,6 +106,12 @@ name = "arrayvec"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
+
+[[package]]
+name = "autocfg"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
 name = "backtrace"
@@ -153,7 +203,7 @@ dependencies = [
  "arbitrary",
  "bolero-generator-derive",
  "either",
- "getrandom",
+ "getrandom 0.2.15",
  "rand_core",
  "rand_xoshiro",
 ]
@@ -205,10 +255,38 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
-name = "cc"
-version = "1.2.12"
+name = "bytes"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "755717a7de9ec452bf7f3f1a3099085deabd7f2962b861dae91ecd7a365903d2"
+checksum = "f61dac84819c6588b558454b194026eb1f09c293b9036ae9b159e74e73ab6cf9"
+dependencies = [
+ "serde",
+]
+
+[[package]]
+name = "cbindgen"
+version = "0.27.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3fce8dd7fcfcbf3a0a87d8f515194b49d6135acab73e18bd380d1d93bb1a15eb"
+dependencies = [
+ "clap",
+ "heck 0.4.1",
+ "indexmap",
+ "log",
+ "proc-macro2",
+ "quote",
+ "serde",
+ "serde_json",
+ "syn 2.0.98",
+ "tempfile",
+ "toml",
+]
+
+[[package]]
+name = "cc"
+version = "1.2.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c7777341816418c02e033934a09f20dc0ccaf65a5201ef8a450ae0105a573fda"
 dependencies = [
  "shlex",
 ]
@@ -247,9 +325,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.28"
+version = "4.5.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e77c3243bd94243c03672cb5154667347c457ca271254724f9f393aee1c05ff"
+checksum = "8acebd8ad879283633b343856142139f2da2317c96b05b4dd6181c61e2480184"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -257,12 +335,14 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.27"
+version = "4.5.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b26884eb4b57140e4d2d93652abfa49498b938b3c9179f9fc487b0acc3edad7"
+checksum = "f6ba32cbda51c7e1dfd49acc1457ba1a7dec5b64fe360e828acb13ca8dc9c2f9"
 dependencies = [
+ "anstream",
  "anstyle",
  "clap_lex",
+ "strsim",
 ]
 
 [[package]]
@@ -271,7 +351,7 @@ version = "4.5.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bf4ced95c6f4a675af3da73304b9ac4ed991640c36374e4b46795c49e17cf1ed"
 dependencies = [
- "heck",
+ "heck 0.5.0",
  "proc-macro2",
  "quote",
  "syn 2.0.98",
@@ -282,6 +362,12 @@ name = "clap_lex"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f46ad14479a25103f283c0f10005961cf086d8dc42205bb44c46ac563475dca6"
+
+[[package]]
+name = "colorchoice"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5b63caa9aa9397e2d9480a9b13673856c78d8ac123288526c37d7839f2a86990"
 
 [[package]]
 name = "ctrlc"
@@ -334,7 +420,7 @@ version = "0.1.0"
 dependencies = [
  "dpdk-sys",
  "dpdk-sysroot-helper",
- "errno",
+ "errno 0.1.0",
  "net",
  "serde",
  "thiserror",
@@ -353,6 +439,22 @@ dependencies = [
 [[package]]
 name = "dpdk-sysroot-helper"
 version = "0.0.1"
+
+[[package]]
+name = "dplane-rpc"
+version = "1.0.0"
+source = "git+https://github.com/githedgehog/dplane-rpc.git#d3547d38735804f27d88dfc29db0c1985c5871c5"
+dependencies = [
+ "bytes",
+ "cbindgen",
+ "log",
+ "mac_address",
+ "num-derive",
+ "num-traits",
+ "tracing",
+ "tracing-subscriber",
+ "tracing-test",
+]
 
 [[package]]
 name = "either"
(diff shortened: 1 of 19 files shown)


routing/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[dependencies]
+ipnet = "2.11.0"
+iptrie = "0.9.1"
+mio = { workspace = true, features = ["os-ext", "net"] }
+net = { workspace = true }
+thiserror = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", branch = "master" }
```
qmonnet, line 18:
```
[...]
Caused by:
  Unable to update https://github.com/githedgehog/dplane-rpc.git?branch=master#d3547d38

Caused by:
  failed to fetch into: /home/runner/.cargo/git/db/dplane-rpc-c044331708d83dd5

Caused by:
  revision d3547d38735804f27d88dfc29db0c1985c5871c5 not found

Caused by:
  failed to authenticate when downloading repository
[...]
```

You'll need to make your repository public for CI to pass.

Fredi-raspall: Yes, I know. I forgot to do that.

routing/src/nexthop.rs
```diff
@@ -14,19 +13,19 @@ use std::net::IpAddr;
 use std::option::Option;
 #[cfg(test)]
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
```
qmonnet, line 17: Just flagging the existing discussion with Manish about the locks

qmonnet (approved):
Looks good from my side, I love how you manage to keep your code very clean :rocket:. Awesome work!

I think I'll try to switch to your `Prefix` for the NAT module. Whether or not we use the latest type in `iptrie`, it makes sense to use a consistent abstraction everywhere we use the `RTrieMap`s.

Fredi-raspall:
> Looks good from my side, I love how you manage to keep your code very clean 🚀. Awesome work!

Thanks!
 
> I think I'll try to switch to your `Prefix` for the NAT module. Whether or not we use the latest type in `iptrie`, it makes sense to use a consistent abstraction everywhere we use the `RTrieMap`s.

Yeah, that would keep things homogeneous, and better sooner than later.

routing/src/nexthop.rs
Fredi-raspall: Yes. I'm not too concerned about deadlocks ... yet, because the writers are verly localized and we mostly read elsewhere.

routing/src/cpi.rs
```diff
@@ -0,0 +1,205 @@
...
+    pub cpi_sock_path: Option<String>,
+}
+
+fn open_unix_sock(path: &String) -> Result<UnixDatagram, RouterError> {
+    let _ = std::fs::remove_file(path);
+    let sock = UnixDatagram::bind(path).map_err(|_| RouterError::InvalidSockPath)?;
+    let mut perms = fs::metadata(path)
+        .map_err(|_| RouterError::InvalidSockPath)?
+        .permissions();
+    perms.set_mode(0o777);
```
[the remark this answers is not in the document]
Fredi-raspall: Absolutely. I did that in order to test. Since it is not clear how we'll finally launch the dataplane (in terms of containers volumes/bind mounts), I made my life easy here.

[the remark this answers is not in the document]
qmonnet: True, but if you can do just as well with fewer permissions, I'd drop the unnecessary ones. If it _is_ more convenient for testing, then I guess it should be alright, but something to keep in mind somewhere :thinking:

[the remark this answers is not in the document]
qmonnet: Maybe creating an issue to track this would make sense. On the other hand my impression is we don't like longstanding issues too much and people actively try to close them so I don't know

qmonnet:
Ah but there are all from `dplane_rpc`, probably me not having the right path or the right settings. I'll look into it.

Fredi-raspall:
> Ah but there are all from `dplane_rpc`, probably me not having the right path or the right settings. I'll look into it.

Yes, those are errors of the dplane-rpc crate.
Master branch should be ok.

Fredi-raspall:
The test that is failing above is due to an assertion not being met in iptrie.
I've opened an issue https://github.com/Orange-OpenSource/iptrie/issues/4.

qmonnet (approved):

Cargo.lock
```diff
@@ -482,7 +482,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
```
qmonnet: Do you need these downgrades? (Three in the commit)

deny.toml
```diff
@@ -126,7 +126,7 @@ registries = [
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
 # Lint level for when multiple versions of the same crate are detected
-multiple-versions = "deny"
+#multiple-versions = "deny"
```
qmonnet, line 129: This was required to make the `dplane-rpc` dependency work (see also the next hunk in this file). @daniel-noland any suggestion on a cleaner way to address this?

Cargo.lock
qmonnet: No I don't really mind, I was just curious about the downgrade - it looked like something accidental. But other than that, you're probably right I don't think we need these

deny.toml
daniel-noland: I looked into it and the commit https://github.com/githedgehog/dataplane/pull/233/commits/8fbc40cf9776118de5cc29e778336bfecc2d05eb should sort this.  That commit is in the parser part 3 series

qmonnet: No, it doesn't. There's no duplicate for `syn` in this PR. See the output I get from applying your commit instead of commenting the above directive:

<details><summary>click to unfold</summary>

```console
$ just cargo deny check
error[duplicate]: found 2 duplicate entries for crate 'errno'
   ┌─ /var/home/qmo/dev/dataplane/Cargo.lock:41:1
   │  
41 │ ╭ errno 0.1.0 /var/home/qmo/dev/dataplane/errno
42 │ │ errno 0.3.10 registry+https://github.com/rust-lang/crates.io-index
   │ ╰──────────────────────────────────────────────────────────────────┘ lock entries
   │  
   ├ errno v0.1.0
     └── dpdk v0.1.0
         └── dataplane v0.1.0
   ├ errno v0.3.10
     └── rustix v0.38.44
         └── tempfile v3.16.0
             └── cbindgen v0.27.0
                 └── (build) dplane-rpc v1.0.0
                     └── routing v0.1.0

error[duplicate]: found 2 duplicate entries for crate 'getrandom'
   ┌─ /var/home/qmo/dev/dataplane/Cargo.lock:45:1
   │  
45 │ ╭ getrandom 0.2.15 registry+https://github.com/rust-lang/crates.io-index
46 │ │ getrandom 0.3.1 registry+https://github.com/rust-lang/crates.io-index
   │ ╰─────────────────────────────────────────────────────────────────────┘ lock entries
   │  
   ├ getrandom v0.2.15
     ├── bolero-generator v0.12.0
     │   ├── bolero v0.12.0
     │   │   └── (dev) net v0.0.1
     │   │       ├── dataplane v0.1.0
     │   │       ├── dpdk v0.1.0
     │   │       │   └── dataplane v0.1.0 (*)
     │   │       └── routing v0.1.0
     │   └── bolero-engine v0.12.0
     │       └── bolero v0.12.0 (*)
     └── rand_core v0.6.4
         ├── bolero-generator v0.12.0 (*)
         ├── rand v0.8.5
         │   ├── bolero v0.12.0 (*)
         │   ├── bolero-engine v0.12.0 (*)
         │   └── phf_generator v0.11.3
         │       └── phf_macros v0.11.3
         │           └── phf v0.11.3
         │               └── doxygen-rs v0.4.2
         │                   └── (build) dpdk-sys v0.0.1
         │                       └── dpdk v0.1.0 (*)
         ├── rand_chacha v0.3.1
         │   └── rand v0.8.5 (*)
         └── rand_xoshiro v0.6.0
             ├── bolero-engine v0.12.0 (*)
             └── bolero-generator v0.12.0 (*)
   ├ getrandom v0.3.1
     └── tempfile v3.16.0
         └── cbindgen v0.27.0
             └── (build) dplane-rpc v1.0.0
                 └── routing v0.1.0

error[duplicate]: found 2 duplicate entries for crate 'heck'
   ┌─ /var/home/qmo/dev/dataplane/Cargo.lock:50:1
   │  
50 │ ╭ heck 0.4.1 registry+https://github.com/rust-lang/crates.io-index
51 │ │ heck 0.5.0 registry+https://github.com/rust-lang/crates.io-index
   │ ╰────────────────────────────────────────────────────────────────┘ lock entries
   │  
   ├ heck v0.4.1
     └── cbindgen v0.27.0
         └── (build) dplane-rpc v1.0.0
             └── routing v0.1.0
   ├ heck v0.5.0
     └── clap_derive v4.5.28
         └── clap v4.5.29
             ├── cbindgen v0.27.0
             │   └── (build) dplane-rpc v1.0.0
             │       └── routing v0.1.0
             └── dataplane v0.1.0

warning[unnecessary-skip]: skip 'syn' applied to a crate with only one version
    ┌─ /var/home/qmo/dev/dataplane/deny.toml:202:9
    │
202 │ name = "syn"
    │         ━━━ unnecessary skip configuration
203 │ reason = "syn is basically always a build time dependency so we aren't shipping multiple versions"
    │           ─────────────────────────────────────────────────────────────────────────────────────── reason

advisories ok, bans FAILED, licenses ok, sources ok
1;31merror: 1mRecipe `cargo` failed with exit code 2
```
</details>

Duplicates are for `errno`, `getrandom` and `heck`, all new versions apparently related to dplane-rpc's use of `cbindgen`.

routing/src/nexthop.rs
mvachhar: I think this lock usage is fine for these reasons.  And because they are contained and don't lead out of this stage, I think it is fine.  But we need to watch this, it could quickly spiral out of control.  I want @daniel-noland to look at this and propose alternatives though.  If there is an easy way to avoid this locking, that would be preferable.

mvachhar:
@daniel-noland Can you suggest an alternative to locking here if there is a good one, or decide if the locking here is fine?  Either way, we should not hold up this PR for a fix given that the locking is contained to routing and the writers are localized, but if there is a better way we should do it as soon as it is practical.

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+    //////////////////////////////////////////////////////////////////
+    pub fn get_vrf(&self) -> Option<&Vrf> {
+        self.vrf.as_deref()
+    }
+
+    //////////////////////////////////////////////////////////////////
+    /// Add (assign) an IP address to an interface
+    //////////////////////////////////////////////////////////////////
+    pub fn add_ifaddr(&mut self, ifaddr: &IfAddress) {
+        self.addresses.insert(*ifaddr);
```
daniel-noland, line 167: very much not blocking, but I do suggest making this function return `bool` to indicate if the insert did anything.

Also, since `IfAddress: Copy` you don't need to pass by reference here.  You can just send in a `IfAddress` I think

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+        self.addresses.remove(ifaddr);
+    }
+
+    //////////////////////////////////////////////////////////////////
+    /// Tell if an interface has a certain IP address assigned
+    /// (regardless of the mask)
+    //////////////////////////////////////////////////////////////////
+    pub fn has_address(&self, address: &IpAddr) -> bool {
+        for (addr, _) in &self.addresses {
+            if addr == address {
```
daniel-noland, line 183: Why not use [`contains`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.contains)?

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+
+#[allow(dead_code)]
+impl IfTable {
+    //////////////////////////////////////////////////////////////////
+    /// Create an interface table. All interfaces should live here.
+    //////////////////////////////////////////////////////////////////
+    pub fn new() -> Self {
+        Self(HashMap::new())
+        // TODO: set a fast hasher
+    }
```
daniel-noland, line 202: Yup.  This is a fair todo.  Likely `ahash` but we should benchmark if it turns out to be a control here

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    //////////////////////////////////////////////////////////////////
+    /// Add an interface to the table
+    //////////////////////////////////////////////////////////////////
+    pub fn add_interface(&mut self, iface: Interface) {
+        self.0.insert(iface.ifindex, iface);
+    }
```
daniel-noland, line 209: we may want this method to return `Option<Interface>` or bool to indicate if the `insert` did something.

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+    use super::{IfDataDot1q, IfDataEthernet, IfState, IfTable, IfType, Interface, Vrf};
+    use net::eth::mac::Mac;
+    use net::vlan::Vid;
+    use std::net::IpAddr;
+    use std::str::FromStr;
+    use std::sync::Arc;
+
+    // create a test interface table
+    pub fn build_test_iftable() -> IfTable {
+        /* create interface table */
```
daniel-noland, line 280: This is a good start, but ideally we would implement `TypeGenerator` or `ValueGenerator` of types like `IfTable` and then test properties more abstractly.

daniel-noland (commented):
just submitting comments from review of the first few commits (up to add interface submodule).

Looking pretty good so far

routing/src/adjacency.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub(crate) fn new() -> Self {
+        Self(HashMap::new())
+        // Todo: use a fast hasher
+    }
+}
+
+#[allow(dead_code)]
+impl AdjacencyTable {
+    pub(crate) fn add_adjacency(&mut self, address: IpAddr, mac: Mac, ifindex: IfIndex) {
+        self.0.insert(
```
daniel-noland, line 30: I think  it makes more sense to have this method accept an `Adjacency` directly.

That way `Adjacency` can impose any needed validity constraints in its constructor.

It is also more consistent with the `get_adjacency` method

routing/src/adjacency.rs
```diff
@@ -0,0 +1,40 @@
+use crate::interface::IfIndex;
+use net::eth::mac::Mac;
+use std::collections::HashMap;
+use std::net::IpAddr;
+
+#[allow(dead_code)]
+pub struct Adjacency {
+    pub address: IpAddr,
+    pub mac: Mac,
+    pub ifindex: IfIndex,
```
daniel-noland, line 15: It is fine for the moment, but I doubt that we want all of the members to be public long term.

routing/src/vrf.rs
```diff
@@ -0,0 +1,447 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! VRF module to store Ipv4 and Ipv6 routing tables
+
+use std::hash::Hash;
+use std::net::IpAddr;
+#[cfg(test)]
+use std::str::FromStr;
+
```
daniel-noland, line 10: why not put this `use` under the `test` module?

routing/src/vrf.rs
```diff
@@ -0,0 +1,447 @@
...
+/// We'll use the RPC definitions for these
+use dplane_rpc::msg::{RouteDistance, RouteMetric, RouteType};
+
+/// Every VRF is univocally identified with a numerical VRF id
+pub type VrfId = u32;
+
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+#[doc = "A next-hop in the VRF"]
+pub struct RouteNhop {
```
daniel-noland, line 26: you can just use `///`, no need for `#[doc = ]` unless you want to do something more fancy in terms of preprocessing the docs.

routing/src/vrf.rs
```diff
@@ -0,0 +1,447 @@
...
+use crate::prefix::Prefix;
+use iptrie::{Ipv4Prefix, Ipv6Prefix, RTrieMap};
+use net::vxlan::Vni;
+
+/// We'll use the RPC definitions for these
+use dplane_rpc::msg::{RouteDistance, RouteMetric, RouteType};
+
+/// Every VRF is univocally identified with a numerical VRF id
+pub type VrfId = u32;
+
```
daniel-noland, line 23: This is another place where I recommend a newtype rather than an alias.

That said, not a blocker

routing/src/vrf.rs
```diff
@@ -0,0 +1,447 @@
...
+    /////////////////////////////////////////////////////////////////////////
+    /// Set the VNI for a Vrf
+    /////////////////////////////////////////////////////////////////////////
+    pub fn set_vni(&mut self, vni: Vni) {
+        self.vni = Some(vni);
+    }
+
+    #[inline(always)]
+    #[must_use]
```
daniel-noland, line 129: `inline` annotations are almost never necessary if you are using LTO (as we are).

I recommend avoiding them unless we need to bully the compiler for some reason.

routing/src/vrf.rs
```diff
@@ -0,0 +1,447 @@
...
+            };
+            route.s_nhops.push(shim);
+        }
+    }
+
+    #[inline(always)]
+    fn deregister_shared_nhop(&mut self, shim: ShimNhop) {
+        let key = shim.rc.key;
+        drop(shim);
+        self.nhstore.del_nhop(&key);
```
daniel-noland, line 153: why the manual drop?  It will drop when it goes out of scope at the end of the function

routing/src/vrf.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+    #[inline(always)]
+    fn lpm_v6(&self, target: &Ipv6Prefix) -> (&Ipv6Prefix, &Route) {
+        self.routesv6.lookup(target)
+    }
+    pub fn lpm(&self, target: &IpAddr) -> (Prefix, &Route) {
+        match *target {
+            IpAddr::V4(a) => {
+                let (p, r) = self.lpm_v4(&Ipv4Prefix::from(a));
+                (Prefix::IPV4(*p), r)
```
daniel-noland, line 279: not a big deal, but you should be able to do `self.lpm_v4(&a.into())` here

daniel-noland (commented):
second reound of comments

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+use std::net::IpAddr;
+use std::sync::Arc;
+
+type VrfId = u32; /* temporary */
+pub struct Vrf {
+    vrfid: VrfId,
+}
+
+/// A type to uniquely identify a network interface
+pub type IfIndex = u32;
```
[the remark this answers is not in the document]
Fredi-raspall: > Is there an advantage to a u32 here instead of a pointer to some interface struct somewhere?

We need to identify interfaces and since those may mirror kernel interfaces, I thought that a (unique) index to identify them would be okay. There's one method of the IfTable to add / delete interfaces. My expectation is that that method gets called as a result of creating interfaces in the kernel for instance.

routing/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "routing"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+net = { path = "../net" }
+thiserror.workspace = true
```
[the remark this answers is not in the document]
Fredi-raspall: It's set like that later on I believe.

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+    Down = 1,
+    Up = 2,
+}
+
+#[allow(dead_code)]
+/// An object representing a network interface and its state
+pub struct Interface {
+    pub name: String,
+    pub description: Option<String>,
+    pub ifindex: IfIndex,
```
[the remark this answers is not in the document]
Fredi-raspall: I did not meant to, but it could. When there are many interfaces, operators may forget their purpose and like to have a long description for the interface purpose. I added it essentially for that.

routing/src/interface.rs
Fredi-raspall: > very much not blocking, but I do suggest making this function return bool to indicate if the insert did anything.

Why? Note that it's a HashSet. If the (address,len) existed, it will be replaced (i.e noop). If it  didn't, it will surely be added. So, what would knowing if it existed or not help the caller?

routing/src/interface.rs
Fredi-raspall: I think that the method was meant to find if the interface had a certain address configured (no matter the mask). If we'd use contains(), we would need to add the masklen, since the key is (address, len), right?
Now, if you asked me the utility of that function, .... that's another story :sweat_smile:

routing/src/interface.rs
Fredi-raspall: Glad that you mention `ahash` because that's the one I was thinking of using. Do you think it's a good choice?

routing/src/prefix.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Tell if a prefix is a root prefix
+    pub fn is_root(&self) -> bool {
+        match self {
+            Prefix::IPV4(_) => *self == Prefix::root_v4(),
+            Prefix::IPV6(_) => *self == Prefix::root_v6(),
+        }
+    }
+    /// Get the inner Ipv4Prefix from a Prefix
+    pub(crate) fn get_v4(&self) -> &Ipv4Prefix {
+        match self {
```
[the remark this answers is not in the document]
Fredi-raspall: I meant it! ... and I dislike panicking in general. However, in this case, it's justified, IMO.
Let me explain:
* most of the code tries to use Prefix, over any version-specific (v4/v6).
* get_v4/get_v6 are called, that's because there's the assumption that a particular version is in use.
* If that assumption is wrong, then, there's definitely a bug somewhere.

routing/src/prefix.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Build an IpAddr from a prefix
+    pub fn as_address(&self) -> IpAddr {
+        match *self {
+            Prefix::IPV4(p) => p.network().into(),
+            Prefix::IPV6(p) => p.network().into(),
+        }
+    }
+}
+impl From<(&IpAddr, u8)> for Prefix {
+    fn from(tuple: (&IpAddr, u8)) -> Self {
```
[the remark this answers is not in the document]
Fredi-raspall: You're absolutely right

routing/src/prefix.rs
```diff
@@ -0,0 +1,162 @@
...
+        match *tuple.0 {
+            IpAddr::V4(a) => Prefix::IPV4(Ipv4Prefix::new(a, tuple.1).unwrap()),
+            IpAddr::V6(a) => Prefix::IPV6(Ipv6Prefix::new(a, tuple.1).unwrap()),
+        }
+    }
+}
+impl From<&Ipv4Net> for Prefix {
+    fn from(value: &Ipv4Net) -> Self {
```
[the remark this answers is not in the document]
Fredi-raspall: Yes, and there's probably other places where I do similar things. I'll re-check.
Too many years passing pointers around :sweat_smile:

routing/src/vrf.rs
Fredi-raspall: Agreed.

routing/src/vrf.rs
Fredi-raspall: Happy to remove inline if you think it's not necessary. Note however, that the inline functions are not pub and only called from the ip-version agnostic counterparts (I did that split to keep the code simple).  So, providing that hint to the compiler may be okay ?

routing/src/vrf.rs
Fredi-raspall: Because we want to decrement the refcount _before_ we call self.nhstore.del_nhop() in order to keep stale next-hops around. If we would let it drop by itself, that would happen after the call, right?

routing/src/vrf.rs
Fredi-raspall: Ack!

daniel-noland (approved):
I made a lot of comments but I think we can go ahead and merge this.

We will end up refactoring it when we do #155 but that is ok.

daniel-noland:
Also, @qmonnet and @Fredi-raspall, please see #235 for fixes to the cargo deny check.

Sorry, I thought I had taken care of that but I think the problem was a little more complex than it appeared at first

routing/src/vrf.rs
mvachhar: Does the compiler even use the hint?  I know in C/C++ it often ignores it in favor of its own heuristic, so I'm curious if this is really honored in Rust.

routing/src/interface.rs
daniel-noland: oh, quite right.  I clearly didn't read that method correctly

routing/src/interface.rs
daniel-noland: Every benchmark of `ahash` I have ever done shows that it is appallingly fast.  It feels like cheating.  It is a really high-quality hash in terms of collision rate as well

routing/src/prefix.rs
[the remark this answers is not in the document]
daniel-noland: ah, that is quite fair.  Not a big deal.  Maybe a good opportunity to explore the IDE's automated refactor if you haven't tried that for rust yet.  I have good luck with it regularly (tho it does go off the rails sometimes)

routing/src/vrf.rs
daniel-noland: In my experience the compiler will do its best to honor `#[inline(always)]`

That said, the old joke is that LLVM's heuristic for inlining functions is "yes"

daniel-noland: In my experience the compiler will do its best to honor `#[inline(always)]`.

The annotation is generally most useful when you have a performance sensitive function being called across crate boundaries and you don't want LTO for some reason.  rustc uses "thin local" lto by default which basically means LTO within a single crate.  If you turn that to "thin" or "fat" then you get LTO across all crates in the compile job.

As for "will LLVM inline this," the old joke is that LLVM's heuristic for deciding if it should inline a function is "yes"

Still, @Fredi-raspall is correct that the annotation is harmless here.

routing/src/vrf.rs
daniel-noland: ah, that makes sense.

routing/src/prefix.rs
```diff
@@ -0,0 +1,162 @@
...
+pub enum Prefix {
+    IPV4(Ipv4Prefix),
+    IPV6(Ipv6Prefix),
+}
+
+#[allow(dead_code)]
+impl Prefix {
+    /// Build 0.0.0.0/0. "Default" is a very overloaded term. Calling this root_v4 instead of default_v4.
+    pub fn root_v4() -> Prefix {
+        Prefix::IPV4(Ipv4Prefix::default())
```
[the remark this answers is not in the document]
Fredi-raspall: I tried what you propose (which I like!), but notice that the method calls the default() for IPv4Prefix. In order for that to work, it should call a const method, right? I could write that, but the problem is that IPv4Prefix fields are private and it does not provide any const constructor.

routing/src/rmac.rs
```diff
@@ -0,0 +1,150 @@
...
+        // get second and check that its MAC was updated
+        let r = store.get_rmac(new_vni(3002), remote);
+        assert!(r.is_some());
+        assert_eq!(
+            r.unwrap().mac,
+            Mac::from([0x11, 0x22, 0x33, 0x44, 0x55, 0x66])
+        );
+
+        println!("{store:#?}")
+    }
```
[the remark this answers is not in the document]
Fredi-raspall: Not really, can be removed. I sometimes run the tests with --nocapture. I thought we'd (re)run them like that in the test runners if tests failed?

routing/src/vrf.rs
Fredi-raspall: Agreed, it should go there, thanks.

routing/src/prefix.rs
```diff
@@ -0,0 +1,162 @@
...
+        Prefix::IPV4(Ipv4Prefix::from(*value))
+    }
+}
+impl From<&Ipv6Net> for Prefix {
+    fn from(value: &Ipv6Net) -> Self {
+        Prefix::IPV6(Ipv6Prefix::from(*value))
+    }
+}
+impl<'a> From<&'a Prefix> for &'a Ipv4Prefix {
+    fn from(value: &Prefix) -> &Ipv4Prefix {
```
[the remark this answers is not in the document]
Fredi-raspall: I thought of it. Let me open another PR for that.

routing/src/interface.rs
```diff
@@ -0,0 +1,363 @@
...
+pub struct Vrf {
+    vrfid: VrfId,
+}
+
+/// A type to uniquely identify a network interface
+pub type IfIndex = u32;
+
+/// An Ipv4 or Ipv6 address and mask configured on an interface
+type IfAddress = (IpAddr, u8);
+
```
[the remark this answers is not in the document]
Fredi-raspall: Ok. Will open another PR with the suggestion. Thanks!

qmonnet:
OK, yes I think it's in a good shape and merging it (while there are no conflicts) will help moving the other PRs forward. Looks like the remaining items can all be addressed with follow-up PRs anyway. Thanks!

PR #213, opened 2025-02-05
net: rename UnicastIpv6Address to UnicastIpv6Addr.

Fredi-raspall:
This way, it follows the same naming convention as UnicastIpv4Addr.


qmonnet (approved):
Consistent naming is great! Thanks :slightly_smiling_face:

PR #215, opened 2025-02-06
Make header fields public, make ip headers unicast by type

mvachhar:
* Adds instructions for setting up `clippy` with `rust-anlayzer` in vscode.
* Makes header fields in parsed packets public for use outside the `net` crate
* Makes ip source addresses unicast by type, checked when parsed/set

### DIFF
--- a/net/src/ipv4/mod.rs
+++ b/net/src/ipv4/mod.rs
@@ -31,10 +31,17 @@ impl Ipv4 {
     #[allow(unsafe_code)] // const-eval and trivially safe
     pub const MAX_LEN: NonZero<usize> = unsafe { NonZero::new_unchecked(60) };
 
+    /// Create a new IPv4 header
+    pub(crate) fn new(header: Ipv4Header) -> Result<Self, Ipv4Error> {
+        UnicastIpv4Addr::new(Ipv4Addr::from(header.source))
+            .map_err(Ipv4Error::InvalidSourceAddr)?;
+        Ok(Self(header))
+    }
+
     /// Get the source ip address of the header
     #[must_use]
-    pub fn source(&self) -> Ipv4Addr {
-        Ipv4Addr::from(self.0.source)
+    pub fn source(&self) -> UnicastIpv4Addr {
+        UnicastIpv4Addr::new(Ipv4Addr::from(self.0.source)).unwrap_or_else(|_| unreachable!())
     }
 
     /// Get the destination ip address of the header
@@ -132,6 +139,8 @@ impl Ipv4 {
     ///
     /// This method does not check to ensure that the source is valid.
     /// For example, a multicast source can be assigned to a packet with this method.
+    ///
+    /// Note(manish) Why do we even have this function?
     #[allow(unsafe_code)]
     pub unsafe fn set_source_unchecked(&mut self, source: Ipv4Addr) -> &mut Self {
         self.0.source = source.octets();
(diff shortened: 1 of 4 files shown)


net/src/packet.rs
```diff
@@ -26,11 +26,11 @@ const MAX_NET_EXTENSIONS: usize = 2;
...
-    eth: Eth,
-    net: Option<Net>,
-    transport: Option<Transport>,
-    vlan: ArrayVec<Vlan, MAX_VLANS>,
-    net_ext: ArrayVec<NetExt, MAX_NET_EXTENSIONS>,
+    pub eth: Eth,
+    pub net: Option<Net>,
+    pub transport: Option<Transport>,
+    pub vlan: ArrayVec<Vlan, MAX_VLANS>,
+    pub net_ext: ArrayVec<NetExt, MAX_NET_EXTENSIONS>,
```
qmonnet, line 33: I've got exactly the same chunk in my code :smile:

qmonnet:
image

Is this PR a draft, or ready to review?
(Note that there's a conflict on `net/src/ipv6/mod.rs`, I guess the same as in Daniel's PR)

net/src/ipv4/mod.rs
```diff
@@ -31,10 +31,16 @@ impl Ipv4 {
...
+    pub(crate) fn new(header: Ipv4Header) -> Result<Self, Ipv4Error> {
+        UnicastIpv4Addr::new(Ipv4Addr::from(header.source)).map_err(Ipv4Error::InvalidSourceAddr)?;
+        Ok(Self(header))
+    }
+
     /// Get the source ip address of the header
     #[must_use]
-    pub fn source(&self) -> Ipv4Addr {
-        Ipv4Addr::from(self.0.source)
+    pub fn source(&self) -> UnicastIpv4Addr {
```
daniel-noland, line 42: I love this change.  I should have written it this way from the start.

In a perfect world we would be able to do things more like TypeScript + [Zod's brand method](https://github.com/colinhacks/zod?tab=readme-ov-file#brand) (oh the irony) to eliminate commutativity and associativity issues with qualifier types like this (while also avoiding an infinity mirror of increasingly complex types).

In the mean time, I think this is a great change.

net/src/ipv6/mod.rs
```diff
@@ -22,10 +22,7 @@ pub mod addr;
 
 /// An IPv6 header
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Ipv6 {
```
daniel-noland, line 25: This change will conflict with #208 which was just merged I think.  May want to drop it from PR or rebase onto main

net/src/ipv6/mod.rs
```diff
@@ -29,10 +29,20 @@ impl Ipv6 {
     #[allow(unsafe_code)]
     pub const MIN_LEN: NonZero<usize> = unsafe { NonZero::new_unchecked(40) };
 
+    /// Create a new [`Ipv6`] header
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Ipv6Error::InvalidSourceAddr`] error if the source address is invalid.
+    pub fn new(header: Ipv6Header) -> Result<Self, Ipv6Error> {
```
daniel-noland, line 37: Not strictly required for this PR, but on a longer timeline it would be ideal to remove `etherparse` types from the public api of this crate (we can't 1.0 this crate until that happens).  That said, this is hardly the only place in the crate where we expose those types so :shrug:

net/src/ipv6/mod.rs
```diff
@@ -180,8 +172,19 @@ pub struct MulticastSourceForbidden;
 #[error("hop limit already zero")]
 pub struct HopLimitAlreadyZeroError;
 
+/// Error which is triggered during construction of an [`Ipv6`] object.
+#[derive(thiserror::Error, Debug)]
+pub enum Ipv6Error {
+    /// Source address is invalid because it is multicast.
+    #[error("multicast source forbidden")]
```
daniel-noland, line 179: Consider changing this to 

```rust
#[error("multicast source forbidden (received {0})")]
```

mvachhar:
This is a draft PR because I wanted the tests to pass before troubling folks.  I think it'll all work fine, but I wasn't certain.

net/src/ipv6/mod.rs
mvachhar: Done.

net/src/ipv6/mod.rs
mvachhar: Yep, we can address this issue in bulk later.

net/src/ipv6/mod.rs
mvachhar: Rebased, should have done this in the first place.

net/src/packet.rs
mvachhar: Correct.  I suggest we move these to pub for now and then #188 can set them back to pub(crate) or whatever.

daniel-noland: Works for me :)

daniel-noland (approved):
LGTM

spritzes holy water on commits :sweat_drops:

PR #216, opened 2025-02-06
Run clippy in CI (plus some CI cleanup)

daniel-noland:
1. run clippy (it is honestly embarrassing that I forgot this before)
2. build the docs (equally embarrassing)
3. tame nightly clippy
4. tame nightly doc lints (we had a number of broken links)
5. make nightly build optional again

I know that testing your own CI is always convoluted and meta, but I do think [a quick fault injection test](https://github.com/githedgehog/dataplane/pull/218/commits/a4eef004ba724f4bf62b9619235bd6ae4a306d7a) is in order here.  Thus I bid thee, [behold my works](https://github.com/githedgehog/dataplane/actions/runs/13193895912)

I cooked a mock PR on top of this PR to illustrate that injecting a fault into the nightly (and only into the nightly) will cause us to pass CI but still generate [an admonition that nightly is broken](https://github.com/githedgehog/dataplane/pull/218)).

### DIFF
--- a/.github/workflows/sterile.yml
+++ b/.github/workflows/sterile.yml
@@ -73,7 +73,7 @@ jobs:
           password: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: "install rust"
-        uses: "dtolnay/rust-toolchain@stable"
+        uses: "dtolnay/rust-toolchain@master"
         with:
           toolchain: "${{ matrix.rust }}"
           targets: "x86_64-unknown-linux-gnu"
@@ -156,7 +156,7 @@ jobs:
           password: "${{ secrets.GITHUB_TOKEN }}"
 
       - name: "install rust"
-        uses: "dtolnay/rust-toolchain@stable"
+        uses: "dtolnay/rust-toolchain@master"
         with:
           toolchain: "${{ matrix.rust }}"
           targets: "x86_64-unknown-linux-gnu"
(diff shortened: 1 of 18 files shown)


.github/workflows/sterile.yml
```diff
@@ -100,6 +100,12 @@ jobs:
         run: |
           just debug=true rust="${{matrix.rust}}" profile=release target=x86_64-unknown-linux-musl sterile cargo test
 
+      - name: "run clippy"
+        if: ${{ always() }}
+        run: |
+          just debug=true rust="${{matrix.rust}}" profile=dev target=x86_64-unknown-linux-gnu sterile cargo clippy
+
+
```
qmonnet: Do you expect Clippy in the sterile build to report errors that it wouldn't raise in dev.yaml?

daniel-noland: No but this is more of a bisection tool.

If sterile passes and dev rejects or vise versa then we have a basis to debug.  Clippy is normally a fast test so it seems fine here

qmonnet (approved):
Looks good to me, thanks!

qmonnet:
Except that we'll need to fix all Clippy reports before CI passes :sweat_smile:

daniel-noland:
> Except that we'll need to fix all Clippy reports before CI passes 😅

Indeed.  We seem to be passing now and also I reworked and extended the version you reviewed.

Take another look when you get a spare minute please

qmonnet (approved):
Looks good, thank you!

PR #218, opened 2025-02-07
Demo/daniel noland/fault injection

daniel-noland:


qmonnet:
That's a big huge warning. I'd maybe reduce the size using a `h3` header rather than `h1`.

daniel-noland:
> That's a big huge warning. I'd maybe reduce the size using a `h3` header rather than `h1`.

thusly?

qmonnet:
Yes it's slightly less prominent in the middle of the conversation. I've got mixed feelings about the CI inserting comments instead of just keeping all the results at the bottom, so at least let's keep them with a decent size :slightly_smiling_face: (Not a fan of comments from CI but I do understand we've got no way to get “warning-only“ results and I'm OK with your solution. I just wish there were a better alternative.)

PR #219, opened 2025-02-07
Add cmd line parser

Fredi-raspall:
goal: allow dataplane process to get other params, non-EAL related.

The crate used, clap, allows deriving parser functions automatically and defining default values.
However, that does not seem to work for certain types.


dataplane/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+    lcores: Option<String>,
+    #[arg(long, value_name = "in-memory flag", default_value_t = true)]
+    in_memory: bool,
+    #[arg(long, value_name = "PCI devices to probe")]
+    allow: Vec<String>,
+    #[arg(long, value_name = "huge pages", default_value_t = 8192)]
+    huge_worker_stack: u32,
+    #[arg(long, value_name = "socket memory")]
+    socket_mem: Option<String>,
+    #[arg(long, value_name = "enable/disable telemetry", default_value_t = true)]
```
daniel-noland, line 22: we should remove this argument as I [forcibly disabled DPDK](https://github.com/githedgehog/dpdk-sys/blob/e8a896e986507124b8a33049b6d6137233914a38/nix/dpdk/default.nix#L71) telemetry [in a prior commit to dpdk-sys](https://github.com/githedgehog/dpdk-sys/commit/c35a59b9f59782e6507e09a45e6163a705bac1f1)

(I understand that you are just mirroring the prior, also outdated, parameters here)

dataplane/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+    no_telemetry: bool,
+    #[arg(long, value_name = "iova mode(va|pa)")]
+    iova_mode: Option<String>,
+    #[arg(long, value_name = "loglevel for a specific component")]
```
daniel-noland, line 26: I'm not sure I understand what this argument is about.  Are we plumbing this into the tracing crate in this commit?

dataplane/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Parser)]
+#[command(name = "Hedgehog Fabric Gateway dataplane")]
+#[command(version = "1.0")] // FIXME
+#[command(about = "A next-gen dataplane for next-gen fabric gateway", long_about = None)]
+pub(crate) struct CmdArgs {
+    #[arg(long, value_name = "core-id used as main", default_value_t = 2)]
+    main_lcore: u8,
+    #[arg(long, value_name = "map lcore set to cpu set")]
+    lcores: Option<String>,
+    #[arg(long, value_name = "in-memory flag", default_value_t = true)]
```
daniel-noland, line 14: I suspect that we should "hard code" the `--in-memory` option.

Otherwise we risk leaking truly massive amounts of hugepage memory if DPDK gets a `kill -9` or some other event which prevents it from cleaning up.

dataplane/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+
+pub(crate) use clap::Parser;
+#[derive(Parser)]
+#[command(name = "Hedgehog Fabric Gateway dataplane")]
+#[command(version = "1.0")] // FIXME
+#[command(about = "A next-gen dataplane for next-gen fabric gateway", long_about = None)]
+pub(crate) struct CmdArgs {
+    #[arg(long, value_name = "core-id used as main", default_value_t = 2)]
+    main_lcore: u8,
+    #[arg(long, value_name = "map lcore set to cpu set")]
```
daniel-noland, line 12: I agree with allowing this as a command line flag for the moment.

That said, in the long run, the set of lcores used for workers (and symmetrically \[and inversely] the set of cores used for everything else) needs to be a function of linux kernel command line parameters like [isolcpus](https://wiki.linuxfoundation.org/realtime/documentation/howto/tools/cpu-partitioning/isolcpus), or [tickless configurations](https://wiki.linuxfoundation.org/realtime/documentation/howto/tools/ticklesskernel?s[]=no&s[]=hz) like `no_hz=full`.

There is also [NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access) to consider.


Given that these parameters are either hardware specific or boot time configuration, I'm not sure we can assume the command line is a great place to get this data.

That said,  we could self-inspect for these parameters at launch (ideally in a parent process which invokes execs the dataplane).  That process could hand us this information so maybe command line is the best place to get it.

dataplane/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+        /* --allow */
+        for a in self.allow.iter() {
+            out.push("--allow".to_string());
+            out.push(a.to_owned());
+        }
+
+        // To be removed
+        if self.allow.len() == 0 {
+            out.push("--allow".to_string());
+            out.push("0000:01:00.0,dv_flow_en=1".to_string());
```
daniel-noland, line 74: This is a good default for the moment, but we need to be careful here.  `0000:01:00.0` is the PCI address of the CX7 card in the dev machine.  We can't count on it in general and we should let our deployment process depend on it.

That said, this is still a step in the right direction since the parameter is, at minimum, configurable now.

dataplane/src/args.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+        if self.no_telemetry == false {
+            out.push("--no-telemetry".to_string());
+        } else {
+            out.push("--telemetry".to_string()); /* DPDK default */
+        }
+
+        /* IOVA mode */
+        out.push(format!(
+            "--iova-mode={}",
```
daniel-noland, line 55: this is another place where we need to be careful.  The CX7 wants virtual memory, but other cards will require physical memory (virtual memory bypass).  Ideally we will set this as a function of the driver of the network card we are going to poll.

But again, this is as good as we are going to do for the moment

dataplane/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "Apache-2.0"
 
 [dependencies]
 
+clap = { version = "4.5.27", features = ["derive"] }
```
daniel-noland, line 10: Please move `clap` to the workspace level `Cargo.toml` (with `default-features` disabled and `features = []`).

Then change this line to read

```toml
clap = { workspace = true, features = ["derive"] }
```

This strategy has the advantage of centralizing version numbers and enforcing that all crates in this worspace are on a common version and using common compile options (features).

`cargo deny` will bounce us if we fail this check, but only on a per-crate basis.

I find that debugging can become _enormously_ confusing if you end up in a situation where you ship or depend on multiple versions of the same package.
It is also far easier on our (future) operations and sec-ops teams when we can give them a single, unambiguous, and exact answer to what version of any given thing is in production at any given time.

daniel-noland (changes requested):
Overall I really like this PR.

I had a number of comments, but only two actual concerns.

Address comments or ping me on slack and we can discuss.

NIce work :)

daniel-noland:
> We should also decide if we want to call the main proces "dataplane" or gateway?

This is a good question.

On thinking about it, I suggest that we convert the `dataplane` crate into a library and make another (much smaller) crate called `gateway` which is limited to things like parsing command line parameters and calling `dataplane` functions.

I am a big believer in thin applications and fat libraries.  It makes testing easier, and who knows when we might want to recycle dataplane logic for some other purpose.

Fredi-raspall:
> Overall I really like this PR.
> 
> I had a number of comments, but only two actual concerns.
> 
> Address comments or ping me on slack and we can discuss.
> 
> NIce work :)

Thanks ... Fine to discuss if you're around!

dataplane/src/args.rs
Fredi-raspall: I looked into DPDK documentation and invoked the dataplane passing **--log-level=help** and this is what I got.
```
Log type is a pattern matching items of this list (plugins may be missing):
	bus.auxiliary
	bus.pci
	lib.eal
	lib.ethdev
	lib.hash
	lib.hash.crc
	lib.hash.fbk
	lib.hash.gfni
	lib.hash.thash
	lib.hash.thash_poly
	lib.mbuf
	lib.mempool
	lib.net
	lib.rcu
	lib.ring
	lib.telemetry
	pmd.common.mlx5
	pmd.net.mlx5
	user1
	user2
	user3
	user4
	user5
	user6
	user7
	user8

Syntax using globbing pattern:     --log-level pattern:level
Syntax using regular expression:   --log-level regexp,level
Syntax for the global level:       --log-level level
Logs are emitted if allowed by both global and specific levels.

Log level can be a number or the first letters of its name:
	1   emergency
	2   alert
	3   critical
	4   error
	5   warning
	6   notice
	7   info
	8   debug
```
I understood that with this command you can set the log levels specifically for certain modules.
So, I thought that this could be interesting. However, you're right, will the binding actually honor this ?
I'd say they should ...so, I just added the command as a direct pass-through to the EAL.

dataplane/src/args.rs
Fredi-raspall: I'm fine with that. Mind it defaults to true. But if you think that it should always be there, we can remove the option.

dataplane/Cargo.toml
Fredi-raspall: Agreed. Thanks for the explanation @daniel-noland .

Fredi-raspall:
@daniel-noland I've pushed some changes and also renamed the crate to "gateway". I have NOT changed the name of the dataplane folder so as not to confuse folks that have outstanding PRs.

dataplane/Cargo.toml
```diff
@@ -5,6 +5,10 @@ edition = "2021"
 publish = false
 license = "Apache-2.0"
 
+[[bin]]
+name = "gateway"
+path = "src/main.rs"
```
qmonnet, line 10: This will require some additional changes in the `Dockerfile`, as well as in the `justfile` for targets `sterile-build` and `build-container` where we reference binaries or Rust packages called `dataplane`.

qmonnet (commented):
What's the motivation for renaming `dataplane` to `gateway`?

[EDIT: Sorry I had missed part of the conversation. I like Daniel's suggestion of a dataplane library + gateway binary, for what it's worth.]

dataplane/src/args.rs
Fredi-raspall: @daniel-noland I'm going to leave these for a second iteration.

dataplane/src/args.rs
Fredi-raspall: @daniel-noland Any idea about how to move forward with this?
I can remove the option if you're not convinced of its usefulness.

mvachhar:
> What's the motivation for renaming `dataplane` to `gateway`?
> 
> [EDIT: Sorry I had missed part of the conversation. I like Daniel's suggestion of a dataplane library + gateway binary, for what it's worth.]

I like the suggestion as well, though this repo is really just the dataplane as the full gateway as FRR, the agent etc.  I'm open to a better name for either the dataplane library or gateway binary.  We can also have a binary called dataplane and library called dataplane built out of the same crate.

Fredi-raspall:
> > What's the motivation for renaming `dataplane` to `gateway`?
> > [EDIT: Sorry I had missed part of the conversation. I like Daniel's suggestion of a dataplane library + gateway binary, for what it's worth.]
> 
> I like the suggestion as well, though this repo is really just the dataplane as the full gateway as FRR, the agent etc. I'm open to a better name for either the dataplane library or gateway binary. We can also have a binary called dataplane and library called dataplane built out of the same crate.

I think dataplane is a bit unspecific. True, this repo does not contain the whole thing. Happy to rename it in any other way. Else, we can keep on having dataplane library and dataplane binary, altough a different name would make the distinction more explicit and keep generic stuff in the library and specifics in the binary. I can remove those commits from this PR since the original purpose of this PR was another.

Fredi-raspall:
> > > What's the motivation for renaming `dataplane` to `gateway`?
> > > [EDIT: Sorry I had missed part of the conversation. I like Daniel's suggestion of a dataplane library + gateway binary, for what it's worth.]
> > 
> > 
> > I like the suggestion as well, though this repo is really just the dataplane as the full gateway as FRR, the agent etc. I'm open to a better name for either the dataplane library or gateway binary. We can also have a binary called dataplane and library called dataplane built out of the same crate.
> 
> I think dataplane is a bit unspecific. True, this repo does not contain the whole thing. Happy to rename it in any other way. Else, we can keep on having dataplane library and dataplane binary, altough a different name would make the distinction more explicit and keep generic stuff in the library and specifics in the binary. I can remove those commits from this PR since the original purpose of this PR was another.

@mvachhar  I've removed the patches that do any rename. There is one outstanding comment from @daniel-noland regarding the logging.

qmonnet:
Seems good to merge as far as I can tell. The logging thing can be refined as a follow-up if necessary.

dataplane/src/args.rs
daniel-noland: Sorry, I misunderstood the meaning of the command line argument.

This is for DPDK logging (distinct from tracing).  All  good

daniel-noland:
I agree. Sorry I was out of action for a bit there.  All ahead full!

daniel-noland (approved):

PR #222, opened 2025-02-12
Add a pipeline interface to chain together processing stages

mvachhar:
## Draft

This PR is here as a draft for discussion and refinement, when ready we can merge.

This PR:
* Adds a `Pipeline` object and `PipelineStage` trait to build a packet processing pipeline
* Adds a dummy Config object to use in the `PipelineStage` trait.

This this PR does *not* do:
* Implement a proper thread-safe config interface
* Fully implement the MetaPacket interface - we need to develop this based on stage needs
  * As part of this, we need to figure out how to represent the selected output interface/encap/etc.
* Does not deal with any async interface to pause processing a packet and wakeup stages later

### DIFF
--- a/dataplane/src/config.rs
+++ b/dataplane/src/config.rs
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use std::collections::HashMap;
+
+pub struct Vpc {
+    #[allow(dead_code)]
+    pub vni: u32,
+}
+
+pub struct Config {
+    // Add appropriate configuration fields here
+    #[allow(dead_code)]
+    pub vni_to_vpc_id: HashMap<u32, Vpc>,
+}
+
+impl Config {
+    pub fn new() -> Self {
+        Self {
+            vni_to_vpc_id: HashMap::new(),
+        }
+    }
+}
--- a/dataplane/src/main.rs
+++ b/dataplane/src/main.rs
@@ -1,6 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
+mod args;
+mod config;
+mod nat;
+mod pipeline;
+
 use dpdk::dev::{Dev, TxOffloadConfig};
 use dpdk::eal::Eal;
 use dpdk::lcore::{LCoreId, WorkerThread};
@@ -9,12 +14,12 @@ use dpdk::queue::rx::{RxQueueConfig, RxQueueIndex};
 use dpdk::queue::tx::{TxQueueConfig, TxQueueIndex};
 use dpdk::{dev, eal, socket};
 use net::packet::Packet;
-use net::parse::Parse;
-use tracing::{info, warn};
-mod args;
-mod nat;
+use net::parse::{DeParse, Parse};
+use tracing::{info, trace, warn};
 
-use args::{CmdArgs, Parser};
+use crate::args::{CmdArgs, Parser};
+use crate::config::Config;
+use crate::pipeline::{MetaPacket, Metadata, Passthrough, Pipeline};
 
 #[global_allocator]
 static GLOBAL_ALLOCATOR: RteAllocator = RteAllocator::new_uninitialized();
@@ -29,6 +34,18 @@ fn init(args: impl IntoIterator<Item = impl AsRef<str>>) -> Eal {
         .with_thread_names(true)
         .init();
     rte
+}
+
+// FIXME(mvachhar) construct pipline elsewhere, ideally from config file
+fn setup_pipeline() -> Pipeline {
+    let mut pipeline = Pipeline::new();
+    let config = Config::new();
+
+    let passthrough = Passthrough::new();
+    pipeline.add_stage(Box::new(passthrough)).unwrap();
+    pipeline.update_config(&config).unwrap();
+
+    pipeline
 }
 
 fn main() {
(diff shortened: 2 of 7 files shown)


dataplane/src/config.rs
```diff
@@ -0,0 +1,20 @@
+use std::collections::HashMap;
+
+pub struct Vpc {
+    #[allow(dead_code)]
+    pub vni: u32,
+}
+
+pub struct Config {
```
sergeymatov, line 8: Isn't it similar to https://github.com/githedgehog/dataplane/commit/b92a7b1f2e1bfe17d0943112cd524cae311d0b34#diff-c801271fc4b78850c43936f61ea0be4871f52b297f1be345b207f2a1a4aa5a92R201 << GlobalContext object that we store in NAT code?

dataplane/src/config.rs
```diff
@@ -0,0 +1,20 @@
+use std::collections::HashMap;
+
+pub struct Vpc {
+    #[allow(dead_code)]
+    pub vni: u32,
+}
```
qmonnet, line 6: I've created https://github.com/githedgehog/dataplane/pull/225 to move the `struct Vpc` from the `nat` module into a separate, dedicated module so that it would be easier to pull it here.

dataplane/src/main.rs
```diff
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
+mod config;
+mod pipeline;
```
qmonnet, line 5: Nit: Some other modules are defined below, _after_ the list of `use ...`. I don't know what's the established practice, but whether it's before or after the `use ...` block, can we have them all at a single location please?

dataplane/src/main.rs
```diff
@@ -105,16 +111,42 @@ fn main() {
...
-            for pkt in pkts.iter_mut() {
-                let Ok((packet, _rest)) = Packet::parse(pkt.raw_data_mut()) else {
-                    info!("failed to parse packet");
-                    continue;
-                };
-                info!("received packet: {packet:?}");
+        WorkerThread::launch(lcore_id, move || {
+            // FIXME(mvachhar) construct pipline elsewhere, ideally from config file
+            let mut pipeline = Pipeline::new();
+            let config = Config::new();
```
qmonnet, line 117: Even if not using a config file, we should probably move this to a dedicated function.

See also #225

dataplane/src/pipeline.rs
```diff
@@ -0,0 +1,127 @@
+#![deny(
+    unsafe_code,
+    missing_docs,
+    clippy::all,
+    clippy::pedantic,
+    clippy::unwrap_used,
+    clippy::expect_used,
+    clippy::panic
+)]
```
qmonnet, line 9: Is there a particular reason to have these here? If we want these in general, can we make them repo-wide (or crate-wide)? I think we have some similar stuff in `net` but probably not in `dataplane` yet.

dataplane/src/config.rs
qmonnet, line 1: Missing SPDX/copyright headers

dataplane/src/pipeline.rs
qmonnet, line 1: Missing SPDX/copyright headers

qmonnet (changes requested):
Nice work, looks good to me overall.

The structure looks OK to me, the pipeline stage is a superset that what I had in my branch in progress for plugging the NAT. It should be easy enough for me to adjust.

In addition to docs and tests that you mentioned, we should probably add some tracing/logging to the pipeline struct and stages.

Please find some other comments below.

dataplane/src/config.rs
mvachhar: Yeah, the config here is just a dummy.  We need to agree on a common data model and put that in place of the placeholder I have here.

dataplane/src/main.rs
mvachhar: Fixed to put them all above use definitions, as per @daniel-noland's suggestion.

dataplane/src/config.rs
mvachhar: Fixed.

dataplane/src/pipeline.rs
mvachhar: Fixed.

qmonnet (approved):
Looks OK to me (but it would be good to have ACKs from other people too), with the understanding that there will of course be follow-ups and changes to the Config object at the very least. Having a basic pipeline in the code will be helpful to plug other components such as NAT.

You can drop the first commit, that was merged in its own PR.

Fredi-raspall (changes requested):
LGTM @mvachhar !! The one thing I'd like to clarify is the point about the associated type for the state that each stage may require.

mvachhar:
> LGTM @mvachhar !! The one thing I'd like to clarify is the point about the associated type for the state that each stage may require.

Is `&mut Self` not good enough for that purpose?

dataplane/src/pipeline.rs
```diff
@@ -0,0 +1,127 @@
...
+
+pub trait PipelineStage {
+    fn start(&mut self) -> Result<(), Box<dyn Error>>;
+
+    // FIXME(mvachhar) This interface is likely to change once we figure out
+    // how to do this in a thread-safe manner, please use sparingly while
+    // we figure out the best way to do this.
+    fn update_config(&mut self, config: &Config) -> Result<(), Box<dyn Error>>;
+
+    fn process(
```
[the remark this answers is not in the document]
Fredi-raspall: In fact, do all stages need to refer to the same config? You could get rid of the super-temporary Config this way.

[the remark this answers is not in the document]
mvachhar: Well, the stages need to be told about the config changes, right?  So `update_config` needs it.

dataplane/src/config.rs
mvachhar: Yep, we should use that when ready, but config is just a dummy.

dataplane/src/main.rs
mvachhar: Done.

dataplane/src/pipeline.rs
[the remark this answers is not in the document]
mvachhar: Anyway, we have to figure out how all this will work, not for this PR.

dataplane/src/pipeline.rs
```diff
@@ -0,0 +1,172 @@
...
+        &mut self,
+        packets: Box<dyn Iterator<Item = MetaPacket>>,
+    ) -> Box<dyn Iterator<Item = MetaPacket>> {
+        self.stages
+            .iter_mut()
+            .fold(packets, |packets, stage| stage.process(packets))
+    }
+}
+
+pub struct Passthrough {}
```
daniel-noland, line 106: Not blocking for merge here, but just FYI, 

```rust
pub struct Passthrough;
```

is equally legal syntax (and somewhat more idiomatic for zero sized types with no fields)

daniel-noland: Also, if you want to make sure that the user of this `struct` needs to go through the `Passthrough::new` method in order to create a `Passthrough` I would add `#[non_exhaustive]` above the `struct` definition

daniel-noland (approved):
Looks good to me.

I made some infromational comments but nothing blocking

dataplane/src/pipeline.rs
mvachhar: Done.

PR #225, opened 2025-02-12
Reorganise NAT elements, main() function

qmonnet:
- **dp: Split main() function into EAL init, devices init, workers start**
- **nat: Split out Vpc, Pif, PrefixTrie from nat module**

First commit splits the `main()` function into different blocks to make it easier to read, and work with.

Second commit splits some components used in the NAT module out of that module, and into their respective new modules, `vpc` and `vpc::prefix_trie`, in order to make them accessible to separate modules from the dataplane.

### DIFF
--- a/dataplane/src/main.rs
+++ b/dataplane/src/main.rs
@@ -24,7 +24,7 @@ use crate::pipeline::{MetaPacket, Metadata, Passthrough, Pipeline};
 #[global_allocator]
 static GLOBAL_ALLOCATOR: RteAllocator = RteAllocator::new_uninitialized();
 
-fn init(args: impl IntoIterator<Item = impl AsRef<str>>) -> Eal {
+fn init_eal(args: impl IntoIterator<Item = impl AsRef<str>>) -> Eal {
     let rte = eal::init(args);
     tracing_subscriber::fmt()
         .with_max_level(tracing::Level::DEBUG)
@@ -48,16 +48,8 @@ fn setup_pipeline() -> Pipeline {
     pipeline
 }
 
-fn main() {
-    let args = CmdArgs::parse();
-    let eal: Eal = init(args.eal_params());
-
-    let (stop_tx, stop_rx) = std::sync::mpsc::channel();
-    ctrlc::set_handler(move || stop_tx.send(()).expect("Error sending SIGINT signal"))
-        .expect("failed to set SIGINT handler");
-
-    let devices: Vec<Dev> = eal
-        .dev
+fn init_devices(eal: &Eal) -> Vec<Dev> {
+    eal.dev
         .iter()
         .map(|dev| {
             let config = dev::DevConfig {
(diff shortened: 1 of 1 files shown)


dataplane/src/main.rs
```diff
@@ -19,7 +19,7 @@ use args::{CmdArgs, Parser};
 #[global_allocator]
 static GLOBAL_ALLOCATOR: RteAllocator = RteAllocator::new_uninitialized();
 
-fn init(args: impl IntoIterator<Item = impl AsRef<str>>) -> Eal {
```
mvachhar, line 22: I think these changes to split up main are good, but I'll defer to @daniel-noland since he is more familiar with what has to happen for dpdk bring up.

dataplane/src/vpc/prefix_trie.rs
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: Apache-2.0
```
mvachhar, line 1: We should probably just drop this code in favor of the [`iptrie` crate](https://crates.io/crates/iptrie), which I think @Fredi-raspall is using for the FIB.  Eventually, we'd replace `iptrie` with a tree bitmap that we implement and optimize when `iptrie` becomes a meaningful bottleneck.

dataplane/src/vpc/mod.rs
```diff
@@ -0,0 +1,114 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
```
mvachhar, line 2: These look reasonable for now, though they'll probably change again when we work with @sergeymatov's protoc file for the config coming from the agent.

mvachhar (changes requested):

dataplane/src/vpc/prefix_trie.rs
qmonnet: The PR currently has no functional change. Do you want me to remove the prefix_trie code we have and rework the PIF table to use `iptrie` in this PR? I can take a look into it tomorrow, if this is what you suggest.

mvachhar: I think it is worth the effort and shouldn't be too difficult to move, I think.  If it takes more than 1/2 a day, then we can discuss further.

daniel-noland (approved):
This change set looks fine to me.

That said, this part of the code base is in major flux at the moment so I expect this will all change again soon.

dataplane/src/vpc/prefix_trie.rs
qmonnet: Switching to iptrie as a preparatory PR in https://github.com/githedgehog/dataplane/pull/230

dataplane/src/vpc/mod.rs
qmonnet: Note that this PR doesn't change the existing objects. I'm just moving stuff around. So yes, they will likely need some adjustments in any case

dataplane/src/main.rs
qmonnet: Daniel approved :tada:

PR #227, opened 2025-02-13
net:eth: implement Display for Mac

Fredi-raspall:


qmonnet (approved):
Looks good, thank you!

PR #229, opened 2025-02-13
Add CI check for SPDX/copyright headers

qmonnet:
Let's automate the check on SPDX header and copyright notice for new files. This commit introduces a workflow to ensure that all source code files have these headers. It's simpler to rely on the check than to expect reviewers to think about it each time a new file is added to the repository.

"Source code files" are currently defined as Rust files (.rs), shell scripts (.sh), and justfile.

We could optimize this workflow to run only when source files are modified. But the check is to trivial that it's currently not worth the pain.


qmonnet:
Works as I expected:
image

I'm removing the test commit.

mvachhar (approved):

PR #230, opened 2025-02-14
nat: Use iptrie instead of custom LPM implementation

qmonnet:
Replace our PrefixTrie implementation with the one from the iptrie crate.

We get some added complexity due to the fact that (as far as I can tell, and in spite of what the crate's README says) there seems to be no way to store a type that represents both IPv4 and IPv6 prefixes; so we end up with two distinct maps. We wrap these into a struct PrefixTrie (distinct from the one we had so far: it no longer implements the map), to avoid dealing with IP versions in the implmementation of the PifTable and the GlobalContext. We also have to implement the Debug trait for the PrefixTrie, given that the LPM map we use does not provide it.

PrefixTrie's insert() function will eventually be reworked, when we change the definition of the endpoints to use CIDR objects rather than a combination of address + prefix length.

Note: iptrie is also introduced in #211.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ clap = { version = "4.5.27", default-features = false, features = ["std"] }
 ctrlc = { version = "3.4.5", default-features = false, features = [] }
 doxygen-rs = { version = "0.4.0", default-features = false, features = [] }
 etherparse = { version = "0.17.0", default-features = false, features = [] }
+iptrie = { version = "0.9.1", default-features = false, features = [] }
 serde = { version = "1.0.213", default-features = false }
 serde_yml = { version = "0.0.12", default-features = false, features = [] }
 thiserror = { version = "2.0.11", default-features = false, features = [] }
--- a/dataplane/Cargo.toml
+++ b/dataplane/Cargo.toml
@@ -6,11 +6,11 @@ publish = false
 license = "Apache-2.0"
 
 [dependencies]
-
 clap = { workspace = true, features = ["derive"] }
 ctrlc = { workspace = true, features = ["termination"] }
-net = { workspace = true, features = ["serde"] }
 dpdk = { workspace = true }
+iptrie = { workspace = true }
+net = { workspace = true, features = ["serde"] }
 serde = { workspace = true, features = ["derive"] }
 serde_yml = { workspace = true }
 thiserror = { workspace = true }
--- a/dataplane/src/nat.rs
+++ b/dataplane/src/nat.rs
@@ -1,15 +1,23 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
+use iptrie::{Ipv4Prefix, Ipv6Prefix, RTrieMap};
 use net::vxlan::Vni;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fs;
-use std::hash::Hash;
 use std::net::IpAddr;
 use std::path::Path;
 use tracing::{error, warn};
+
+#[derive(thiserror::Error, Debug)]
+pub enum NatError {
+    #[error("Failed to create IP prefix")]
+    BadPrefix,
+    #[error("PIF already exists")]
+    PifExists,
+}
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 struct Pif {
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -36,6 +36,7 @@ clap = { version = "4.5.27", default-features = false, features = ["std"] }
 ctrlc = { version = "3.4.5", default-features = false, features = [] }
 doxygen-rs = { version = "0.4.0", default-features = false, features = [] }
 etherparse = { version = "0.17.0", default-features = false, features = [] }
+iptrie = { version = "0.9.1", default-features = false, features = [] }
```
Fredi-raspall, line 39: This is already introduced in https://github.com/githedgehog/dataplane/pull/211

dataplane/Cargo.toml
```diff
@@ -6,11 +6,11 @@ publish = false
 license = "Apache-2.0"
 
 [dependencies]
-
 clap = { workspace = true, features = ["derive"] }
 ctrlc = { workspace = true, features = ["termination"] }
-net = { workspace = true, features = ["serde"] }
 dpdk = { workspace = true }
+iptrie = { workspace = true }
```
Fredi-raspall, line 12: Same here. Nothing wrong, you'll have to deal with it when merge/rebase.
Raising this because I'd actually merge https://github.com/githedgehog/dataplane/pull/211 first for the reason mentioned below.

dataplane/src/nat.rs
```diff
@@ -19,88 +27,60 @@ struct Pif {
...
-
-    #[tracing::instrument(level = "trace")]
-    fn find(&self, keys: Vec<K>) -> Option<V> {
-        let mut node = &self.root;
-        let mut best_match = None;
-
-        for key in keys {
-            if let Some(val) = &node.value {
-                best_match = Some(val.clone());
+    fn insert(&mut self, addr: &IpAddr, len: &u8, value: String) -> Result<(), NatError> {
```
Fredi-raspall, line 46: This is absolutely fine. However, in https://github.com/githedgehog/dataplane/pull/211, I added an enum Prefix that works for both IPv4 and IPv6. I would use that type here so that:
   * the same implementation works also for IPv6 (with minor tweaks like using trie_ipv4 or trie_ipv6)
   * we don't leak the Ipv4Prefix, Ipv6Prefix type (specific to the iptrie) outside of the "scope" where they are absolutely needed. The Prefix uses "standard" types.
   * we have a common type to represent prefixes throughout.

Fredi-raspall: Oh, sorry, now I realize that you added also IPv6.
Up to you then to use Prefix.

dataplane/src/nat.rs
```diff
@@ -19,88 +27,60 @@ struct Pif {
...
+        match addr {
+            IpAddr::V4(ip) => {
+                let prefix = Ipv4Prefix::new(*ip, *len).or(Err(NatError::BadPrefix))?;
+                if self.trie_ipv4.insert(prefix, value).is_some() {
```
Fredi-raspall, line 50: Hum. I am not sure this is correct (or what you intend to do).
IIRC, trie.insert() will insert the new value (string) and return the previous one, if there was any.
With this code, an insert of value=BAR would overwrite value=FOO (if the value existed for the same prefix) and then the code would complain that it existed already. But the end value for the prefix would be BAR, not FOO.... right?

dataplane/src/nat.rs
```diff
@@ -19,88 +27,60 @@ struct Pif {
...
+                    return Err(NatError::PifExists);
+                }
             }
-            if let Some(child) = node.children.get(&key) {
-                node = child;
-            } else {
-                break;
+            IpAddr::V6(ip) => {
+                let prefix = Ipv6Prefix::new(*ip, *len).or(Err(NatError::BadPrefix))?;
+                if self.trie_ipv6.insert(prefix, value).is_some() {
```
Fredi-raspall, line 56: Same here.

dataplane/src/nat.rs
Fredi-raspall: I think you could do something like (untested):
```
if self.trie_ipv4.get(prefix, value).is_some() {
     Err(NatError::PifExists);
} else {
    self.trie_ipv4.insert(prefix, value);
    Ok(())
}
```

dataplane/src/nat.rs
```diff
@@ -19,88 +27,60 @@ struct Pif {
...
-                for i in 0..prefix_len {
-                    bits.push((ipv6.octets()[i as usize / 8] >> (7 - (i % 8))) & 1);
-                }
-            }
+    fn find(&self, ip: &IpAddr) -> Option<String> {
+        let res = match ip {
+            IpAddr::V4(ip) => self.trie_ipv4.lookup(ip).1,
+            IpAddr::V6(ip) => self.trie_ipv6.lookup(ip).1,
+        };
+        if res.is_empty() {
```
Fredi-raspall, line 70: Here you're using the fact that the RTrieMap always has a root object, correct?
When you will add a delete() method, be careful to forbid deletions for 0/0 and 0::/0.
Else the RtrieMap will panic.

Fredi-raspall (changes requested):
Nice work Quentin!

dataplane/Cargo.toml
qmonnet: If it's the exact same change for this line, I expect git/GitHub is able to silently discard this chunk when rebasing, so it shouldn't create a merge conflict, whichever PR gets in first.

dataplane/src/nat.rs
qmonnet: I'll take a look, thanks.

Note also that iptrie [might support both IPv4 and IPv6 in a single map in the near future](https://github.com/Orange-OpenSource/iptrie/issues/3#issuecomment-2658574297).

dataplane/src/nat.rs
qmonnet: Yes, the lookup always returns something, so I've got to ensure the value we get is different from the one attached to the root - which in our case is the default value for `String`, an empty string.

I could probably add a comment to make it clear.

I'll try to keep that in mind for the `delete()` method, thanks! :)

dataplane/src/nat.rs
qmonnet: Yes, correct. I realised that I needed to fix it at some point last night, but eventually forgot about it :facepalm:. I'll address it, thank you.

qmonnet:
My current work for this has currently moved to branch https://github.com/githedgehog/dataplane/tree/pr/qmonnet/split-nat-with-prefix, based on top of Fredi's PR from #211.

qmonnet:
Superseded with #257

PR #233, opened 2025-02-18
Parser part 3

daniel-noland:


net/src/headers.rs
qmonnet, line 1: The commit is about renaming packet.rs into headers.rs, but I think you forgot to actually remove net/src/packet.rs. Please double-check

net/src/icmp4/mod.rs
```diff
@@ -12,11 +17,15 @@ use std::num::NonZero;
 pub struct Icmp4(Icmpv4Header);
 
 impl Parse for Icmp4 {
-    type Error = LengthError;
+    type Error = UnreachableError;
```
qmonnet, line 24: Nothing major, but these changes should be in the commit where you introduce the unreachable issues, not in the usize/u16 swap commit.

net/src/ipv6/mod.rs
```diff
@@ -23,13 +23,19 @@ use tracing::{debug, trace};
 pub mod addr;
 pub mod flow_label;
 
+#[allow(unused_imports)] // re-export
+#[cfg(any(test, feature = "arbitrary"))]
+pub use contract::*;
```
qmonnet, line 28: I have no idea where this is coming from or how this is related to the rest of the commit.

dataplane/src/main.rs
```diff
@@ -1,28 +1,28 @@
-// SPDX-License-Identifier: Apache-2.0
-// Copyright Open Network Fabric Authors
+// // SPDX-License-Identifier: Apache-2.0
+// // Copyright Open Network Fabric Authors
+//
```
qmonnet, line 3: Double commenting looks like a mistake.

dataplane/src/main.rs
```diff
@@ -1,28 +1,28 @@
...
+use crate::pipeline::NetworkFunction;
+pub mod args;
+pub mod config;
+pub mod nat;
+pub mod packet;
 
 use crate::args::{CmdArgs, Parser};
-use crate::config::Config;
-use crate::pipeline::{MetaPacket, Metadata, Passthrough, Pipeline};
+pub mod pipeline;
```
qmonnet: Why are these public?

dataplane/src/pipeline.rs
```diff
@@ -1,173 +1,168 @@
...
+        self.nf2.process(self.nf1.process(input))
     }
+}
 
-    fn update_config(&mut self, _: &Config) -> Result<(), Box<dyn Error>> {
-        Ok(())
-    }
+pub trait Pipeline<Buf: PacketBufferMut>: NetworkFunction<Buf> {
+    fn add_stage<NF: NetworkFunction<Buf>>(self, nf: NF) -> impl NetworkFunction<Buf>;
+}
```
qmonnet, line 103: This is not “Introduce PacketBuffer [...] to dataplane”, this is a complete rewrite of the pipeline architecture.

deny.toml
```diff
@@ -198,6 +198,10 @@ reason = "needed for test tracing.  Remove this exception when https://github.co
 name = "regex-syntax"
 reason = "needed for test tracing.  Remove this exception when https://github.com/tokio-rs/tracing/pull/3033 lands"
 
+[[bans.skip]]
+name = "syn"
+reason = "syn is basically always a build time dependency so we aren't shipping multiple versions"
```
qmonnet, line 203: Is this necessary? It doesn't seem to be the case on `main`, but maybe you add a new version with your PR?

qmonnet (changes requested):

qmonnet:
Thanks Daniel, this is some great progress.

Unsurprisingly, I've got some concerns with the size and shape of the PR. We have _twenty-six_ commits at the moment, that cover at least:

1. Various minor local fixes, tweaks and improvements
1. Package-wide renames, mixed with fixes
1. New types, generation, and fuzzing tests, all related to fuzzing as far as I can tell
1. New features, such as the implementation of `PacketBuffer`, its piping through the DPDK mbuf model
1. Based on `PacketBuffer`, a complete rewrite of the pipeline model

Are all of these really constituting “Parser part 3”? It really feels to me like these different items have little in common.

I looked at the different commits, but I don't feel confident I can catch issues, let alone significant issues, in a changeset this big. I'm not confident - but may be wrong - other folks in the team can, either. What I fear is that this PR will get merged after getting superficial reviews, including on the new pipeline model, because it's so hard to comment efficiently on something of this size.

In the future, I would much rather have several, smaller pull requests. For example, each of the bullet points listed above could be its own, separate PR (the last two, related to `PacketBuffer`, might arguably go together). My impression is that we're decently fast at reviewing and merging PRs, so I wouldn't expect separate PRs to be an issue and to block someone's workflow for long? Or do you have concerns with doing smaller PRs? Maybe something we could address?

net/src/ipv6/mod.rs
```diff
@@ -430,10 +453,13 @@ mod contract {
     impl TypeGenerator for Ipv6 {
         fn generate<D: Driver>(u: &mut D) -> Option<Self> {
             let mut header = Ipv6(Ipv6Header::default());
+            #[allow(unsafe_code)] // safe in test context
+            unsafe {
+                header.set_next_header(u.gen()?);
```
mvachhar, line 485: Why was this not needed before, but only after the u16 change?

net/src/parse.rs
```diff
@@ -59,6 +59,10 @@ pub(crate) trait ParsePayloadWith {
     fn parse_payload_with(&self, param: &Self::Param, cursor: &mut Reader) -> Option<Self::Next>;
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("Maximum legal packet buffer size is 2^16 (requested {0})")]
+pub struct IllegalBufferLength(pub usize);
```
mvachhar, line 64: THere is no std::io or some Rust way of saying buffer to larger or something like that?

net/src/parse.rs
```diff
@@ -165,3 +165,14 @@ pub enum DeParseError<E> {
...
 }
+
+pub trait NonZeroUnSizedNumericUpcast {
+    fn cast(self) -> NonZero<usize>;
+}
+
+impl NonZeroUnSizedNumericUpcast for NonZero<u16> {
+    fn cast(self) -> NonZero<usize> {
+        #[allow(unsafe_code)] // trivially sound since input is already non-zero
+        unsafe { NonZero::new_unchecked(self.get() as usize) }
```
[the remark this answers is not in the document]
mvachhar: Also, can we call this something better than cast.  Seenig it in context in later commits, I hat this name.  Maybe upsize, or can we use the `From` trait instead so it is clear what is happening?

dpdk/src/mem.rs
```diff
@@ -399,7 +402,74 @@ pub struct Mbuf {
...
+
+impl AsMut<[u8]> for Mbuf {
+    fn as_mut(&mut self) -> &mut [u8] {
+        self.raw_data_mut()
+    }
+}
+
+impl Headroom for Mbuf {
+    fn headroom(&self) -> u16 {
+        unsafe { rte_pktmbuf_headroom(self.raw.as_ptr()) }
```
mvachhar, line 424: How does this work, how does it know how much headroom, can it fail?

dataplane/src/main.rs
```diff
@@ -107,34 +95,30 @@ fn start_rte_workers(devices: &Vec<Dev>) {
         let rx_queue = devices[0].rx_queue(RxQueueIndex(i as u16)).unwrap();
         let tx_queue = devices[0].tx_queue(TxQueueIndex(i as u16)).unwrap();
         WorkerThread::launch(lcore_id, move || {
-            let mut pipeline = setup_pipeline();
-            pipeline.start().unwrap();
+            let mut phase0 = DecrementTtl
```
mvachhar, line 98: We can put this type of sample code into tests for the pipeline, we can stick with just passthrough here.

dataplane/src/pipeline.rs
```diff
@@ -1,173 +1,168 @@
...
-
-        let input_packets = vec![test_packet];
-        let output: Vec<_> = pipeline
-            .process_packets(Box::new(input_packets.into_iter()))
-            .collect();
-
-        assert_eq!(1, output.len());
-        assert_eq!(src_mac, output[0].packet.eth.source());
-        assert_eq!(dst_mac, output[0].packet.eth.destination());
+    fn biscuit() {
```
mvachhar, line 135: Can we get meaningful function names, or is this scratch code that should be dropped?

mvachhar: For sample code we should write some docs and have the docs get tested using the doc-test stuff, which should run with cargo test.

mvachhar:
I forgot to mention in my review overview, we should examine if the `Chain` stuff and pipeline changes are strictly needed.  I think it might be better to drop those particular changes in favor of making Mbuf 'static compatible so `Packet` can be properly boxed.

dataplane/src/main.rs
daniel-noland: they (briefly) needed to be public in order to make a benchmark for the rework with the `Chain` struct.  Since we are dropping that whole commit I'm going to fix

daniel-noland: That said, in a binary crate it make no real difference

dataplane/src/main.rs
daniel-noland: Yes, and also I'm going to drop this whole commit so I won't do that in the PR

dataplane/src/main.rs
daniel-noland: this should be resolved by dropping the introducing commit

dataplane/src/pipeline.rs
daniel-noland: I agree in principal but the point is moot here since this whole commit is dropped now

dpdk/src/mem.rs
daniel-noland: [DPDK is tracking the headroom](https://doc.dpdk.org/guides/prog_guide/mbuf_lib.html#figure-mbuf1)

net/src/ipv6/mod.rs
daniel-noland: this is a rebase mistake.  Will fix

net/src/ipv6/mod.rs
daniel-noland: also a rebase mistake

net/src/headers.rs
daniel-noland: yeah, I was clearly very tired when I did this rebase

net/src/parse.rs
daniel-noland: I could likely recycle something from std::io.  That would give less control over the generated error message.

I kinda expect we will never see this error (but I won't say that too loud tho because the programming gods will strike me down)

deny.toml
daniel-noland: I will confirm.  This commit is what confused me regarding the fate of #211 and which resulted in #235

I swear I had cargo deny passing in both PRs but in the morning it was a completely different issue so I don't know what happened

dataplane/src/pipeline.rs
daniel-noland: True.  That said, this should be deleted now

net/src/parse.rs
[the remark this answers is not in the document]
daniel-noland: If we use `From` then the `.cast` will just be replaced with `.into`.

I think `to_usize` is fair?

daniel-noland:
I agree with dropping the pipeline changes (under the assumption that I can tame the lifetime issues).

I'm going to put the PR back in draft.

I'll rework all the commits and address comments.  I'll also redo the pipeline thing.

Thanks for review.

daniel-noland (commented):
Going back to draft for now to address comments

daniel-noland:
My intention is to break the PR into several different PRs and then close this one as "won't merge"

qmonnet:
Closing the PR as this is outdated and superseded with a bunch of other PRs. Please reopen if necessary.

PR #235, opened 2025-02-18
Tame cargo deny

daniel-noland:
Hopefully this calms cargo deny down for @Fredi-raspall

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ resolver = "3"
 dpdk = { path = "./dpdk" }
 dpdk-sys = { path = "./dpdk-sys" }
 dpdk-sysroot-helper = { path = "./dpdk-sysroot-helper" }
+dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", branch = "master" }
 errno = { path = "./errno", package = "dataplane-errno" }
 net = { path = "./net" }
 routing = { path = "./routing" }
@@ -34,6 +35,7 @@ arbitrary = { version = "1.4.1", default-features = false, features = [] }
 arrayvec = { version = "0.7.6", default-features = false, features = [] }
 bindgen = { version = "0.71.1", default-features = false, features = [] }
 bolero = { version = "0.12.0", default-features = false, features = [] }
+bytes = { version = "1.9.0", default-features = false, features = [] }
 # std feature currently globally required for clap
 clap = { version = "4.5.30", default-features = false, features = ["std"] }
 ctrlc = { version = "3.4.5", default-features = false, features = [] }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -2,10 +2,10 @@
 
 members = [
 	"dataplane",
+	"errno",
```
mvachhar, line 5: Why did the location of this move?

daniel-noland: the linter put them in alphabetical order :woman_shrugging:

daniel-noland: I'm happy to take that out of the pr, it just seems harmless

mvachhar: It is specifically not alphabetical, which is why I asked, I don't care either way, was just curious.

daniel-noland: hmmm, you are right.  I always just let the linter do its thing but it looks like it went off the rails here.  Wil fix

Cargo.toml
```diff
@@ -26,7 +26,8 @@ resolver = "3"
 dpdk = { path = "./dpdk" }
 dpdk-sys = { path = "./dpdk-sys" }
 dpdk-sysroot-helper = { path = "./dpdk-sysroot-helper" }
-errno = { path = "./errno" }
+dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", branch = "master" }
```
Fredi-raspall, line 29: Why does this need to be in the top-level if only the routing crate will use it ?

Cargo.toml
```diff
@@ -35,17 +36,20 @@ arbitrary = { version = "1.4.1", default-features = false, features = [] }
...
 bindgen = { version = "0.71.1", default-features = false, features = [] }
 bolero = { version = "0.12.0", default-features = false, features = [] }
-clap = { version = "4.5.27", default-features = false, features = ["std"] }
+bytes = { version = "1.9.0", default-features = false, features = [] }
+clap = { version = "4.5.30", default-features = false, features = ["std"] }
 ctrlc = { version = "3.4.5", default-features = false, features = [] }
 doxygen-rs = { version = "0.4.0", default-features = false, features = [] }
 etherparse = { version = "0.17.0", default-features = false, features = [] }
+ipnet = { version = "2.11.0", default-features = false, features = [] }
+iptrie = { version = "0.10.2", default-features = false, features = [] }
```
Fredi-raspall, line 45: I'm fine to upgrade this one, but 0.10.2 has been very recently released. I wonder how should we be dealing in those cases when it comes to upgrades, given that our coverage isn't high at the moment?

Fredi-raspall (approved):
Looks good to me @daniel-noland , thanks!

qmonnet:
#211 has just been merged, this PR now needs a rebase on the `main` branch

qmonnet:
Now rebased

Cargo.toml
qmonnet: I don't have too much input here, but iptrie v0.9 has been released in December 2024 only (so, not long ago) and doesn't seem to have a huge amount of users, judging by the number of contributors and GitHub stars - for what it's worth. So I don't think we're taking a lot more risks here.

PR #249, opened 2025-02-19
Revert "Syn is no longer duplicated"

daniel-noland:
This reverts commit 90a9056c

Turns out syn is still duplicated


qmonnet:
I don't see that crate being duplicated at the moment, what branch did you see this on?

daniel-noland:
I was getting bounced by cargo deny at the end of parser part 3 without this

qmonnet:
OK Let's try to add add it when we merge the related change, in that case

daniel-noland:
I suspect that syn was de-duplicated since this was created.  I'm fine with leaving it as a singular requirement for now.  But given that it is a build dependency I doubt that it matters if it is duplicated.  So we can re-open and merge this if it actually comes up again.

Closing for now.

PR #251, opened 2025-02-19
Incorrect ethertype in ICMPv6 parse

daniel-noland:
We were just using the wrong constant.

### DIFF
--- a/net/src/ipv6/mod.rs
+++ b/net/src/ipv6/mod.rs
@@ -258,10 +258,10 @@ impl ParsePayload for Ipv6 {
                 })
                 .map(|(val, _)| Ipv6Next::Udp(val))
                 .ok(),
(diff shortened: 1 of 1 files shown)


net/src/ipv6/mod.rs
```diff
@@ -258,7 +258,7 @@ impl ParsePayload for Ipv6 {
                 })
                 .map(|(val, _)| Ipv6Next::Udp(val))
                 .ok(),
-            IpNumber::ICMP => cursor
+            IpNumber::IPV6_ICMP => cursor
                 .parse::<Icmp6>()
                 .map_err(|e| {
                     debug!("failed to parse icmp4: {e:?}");
```
qmonnet, line 264: Please fix the debug message, too

qmonnet (changes requested):

PR #252, opened 2025-02-19
Focused fuzzers for Ipv4/6

daniel-noland:
This allows us to generate "interesting" IPv4/IPv6 packets more easily

### DIFF
--- a/net/src/ipv6/mod.rs
+++ b/net/src/ipv6/mod.rs
@@ -424,7 +424,7 @@ impl From<Ipv6ExtNext> for Header {
 mod contract {
     use crate::ip::NextHeader;
     use crate::ipv6::Ipv6;
-    use bolero::{Driver, TypeGenerator};
+    use bolero::{Driver, TypeGenerator, ValueGenerator};
     use etherparse::Ipv6Header;
     use std::net::Ipv6Addr;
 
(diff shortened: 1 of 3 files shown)


net/src/ipv6/mod.rs
```diff
@@ -422,24 +422,58 @@ impl From<Ipv6ExtNext> for Header {
...
+    impl ValueGenerator for GenWithNextHeader {
+        type Output = Ipv6;
+
+        fn generate<D: Driver>(&self, u: &mut D) -> Option<Ipv6> {
             let mut header = Ipv6(Ipv6Header::default());
             header
                 .set_source(u.gen()?)
                 .set_destination(Ipv6Addr::from(u.gen::<u128>()?))
-                .set_next_header(u.gen()?)
+                .set_next_header(self.0)
```
qmonnet, line 463: Why is this safe for IPv6, but not for IPv4?

qmonnet (approved):
Looks good, thank you! Just one question below.

net/src/ipv6/mod.rs
daniel-noland: That is a minor bug I will fix in a future PR.  Good point tho

PR #242, opened 2025-02-19
Implement CommonEthType

daniel-noland:
This is a useful primitive for more complex fuzzing later.

PR #253, opened 2025-02-19
Focused fuzzers for Vlan

daniel-noland:
Waiting on #242

### DIFF
--- a/net/src/vlan/mod.rs
+++ b/net/src/vlan/mod.rs
@@ -337,8 +337,9 @@ impl ParsePayload for Vlan {
 /// Contracts for Vlan types
 #[cfg(any(test, feature = "arbitrary"))]
 mod contract {
+    use crate::eth::ethtype::{CommonEthType, EthType};
     use crate::vlan::{InvalidPcp, InvalidVid, Pcp, Vid, Vlan};
-    use bolero::{Driver, TypeGenerator};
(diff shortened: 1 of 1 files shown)


qmonnet:
When I commented on the size of your PR yesterday, I didn't mean you need one PR per commit, either :sweat_smile:. Absolutely fine to have this commit with #242, on which it depends, for example. Even having all the fuzzing stuff together sounds totally acceptable. Anyway, thanks a lot for making the reviews much more manageable.

net/src/vlan/mod.rs
```diff
@@ -337,8 +337,9 @@ impl ParsePayload for Vlan {
 /// Contracts for Vlan types
 #[cfg(any(test, feature = "arbitrary"))]
 mod contract {
+    use crate::eth::ethertype::{CommonEthType, EthType};
```
qmonnet, line 340: We'll probably need to rename this (`ethtype`)?

qmonnet (approved):
Ack for the second commit

qmonnet:
Now rebased on `main`, with the conflict `ethertype` -> `ethtype` addressed.

qmonnet (approved):

qmonnet:
Ah right GitHub won't remove the request for review from `Dataplane reviewers` as long as the PR is in draft mode.

PR #255, opened 2025-02-19
ICMPv4 generation

daniel-noland:


qmonnet (approved):
Contains some unaddressed TODOs, but looks good otherwise.

net/src/icmp4/mod.rs
```diff
@@ -7,6 +7,10 @@ use crate::parse::{DeParse, DeParseError, LengthError, Parse, ParseError, ParseP
 use etherparse::Icmpv4Header;
 use std::num::NonZero;
 
+#[allow(unused_imports)] // re-export
```
qmonnet, line 10: Why do you need this, given that this is seemingly unnecessary for ICMPv6 generation?

qmonnet: Why not having it for ICMPv6 then?

PR #256, opened 2025-02-19
Fix dataplane tests and re-add package to default workspace members

qmonnet:
- **nat: Remove tracing_test**
- **workspace: Add dataplane to default members**

Refer to individual commit logs for details.


daniel-noland (approved):
I like it.  LGTM

PR #257, opened 2025-02-19
NAT: Replace home-made LPM with iptrie, use Prefix to handle IPv4/IPv6

qmonnet:
Make the nat module use the map imlementation from the iptrie crate, just like we do in the routing package. Also make it reuse routing's Prefix objects for handling either IPv4 or IPv6 addresses.

See individual commits for details.


routing/src/lib.rs
```diff
@@ -11,7 +11,7 @@ mod encapsulation;
 mod errors;
 mod interface;
 mod nexthop;
-mod prefix;
+pub mod prefix;
```
Fredi-raspall, line 14: Yes all of those will need to be public. I have a local commit. I can open PR  with some fixes like that.

routing/src/prefix.rs
```diff
@@ -121,10 +131,68 @@ impl Display for Prefix {
     }
 }
 
+impl Serialize for Prefix {
```
Fredi-raspall, line 134: Why do you need this for? It's absolutely fine, since I was planning to add something similar.... I'm just curious.

Fredi-raspall (commented):
Looks good to me @qmonnet ! Thanks!

routing/src/prefix.rs
qmonnet: Right now, only for tests. We have a test in the nat module that parses the VPCs and PIFs from a YAML file - from the early experiments in the tmp-nat repo. In order to parse the file, I need the objects to implement `Deserialize`. Technically, we don't use `Serialize` at the moment, but having just `Deserialize` and not `Serialize` for the object felt a bit silly.

I'm not particularly proud of the serializer I've come up with, if you've got something more robust feel free to improve or replace, by all means!

routing/src/lib.rs
qmonnet: Right now I only needed to access the `Prefix` stuff, so no rush from my side. But sure, happy to review a PR once you have something ready

routing/src/prefix.rs
Fredi-raspall: It's fine, I was just curious. Thanks!

Fredi-raspall (approved):
I believe this can be merged now.

PR #260, opened 2025-02-21
Rust 2024 edition

daniel-noland:
~~Don't merge this till the official dpdk-sys bump which coresponds to lands (https://github.com/githedgehog/dpdk-sys/pull/88)~~

Upstream landed.  This is ready to roll.


qmonnet (approved):
Looks good, although I note a bunch of cosmetic changes. How did they end up in the same commit, is it that we're using a newer version of `rustfmt` now that you've updated dpdk-sys?

PR #261, opened 2025-02-21
Add abstraction for header manipulation and Packet structure for dataplane

mvachhar:
Add back some of the changes that didn't make it from #233

* Adds abstraction for ethparse headers in net so that we can later enforce correct manipulations
* Add tests for full header parsing
* Add `Packet` to dataplane and drop `MetaPacket`
  * `Packet` manages its own Mbuf and abstracts the buffer type for testing

These changes are a prelude to a better Pipeline interface that can actually be tested and works with different Iterator types.  That PR is still WIP, but close.

### DIFF
--- a/dataplane/src/packet.rs
+++ b/dataplane/src/packet.rs
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use net::buffer::PacketBufferMut;
+use net::eth::EthError;
+use net::headers::{AbstractHeaders, AbstractHeadersMut, Headers, TryHeaders, TryHeadersMut};
+use net::parse::{DeParse, DeParseError, Parse, ParseError};
+use std::cmp::Ordering;
+use std::num::NonZero;
+
+#[derive(Debug)]
+pub struct Packet<Buf: PacketBufferMut> {
+    headers: Headers,
+    /// The total number of bytes _originally_ consumed when parsing this packet
+    /// Mutations to `packet` can cause the re-serialized size of the packet to grow or shrink.
+    consumed: NonZero<u16>,
+    pub mbuf: Buf, // TODO: find a way to make this private
+}
+#[derive(Debug, thiserror::Error)]
+pub struct InvalidPacket<Buf: PacketBufferMut> {
+    #[allow(unused)]
+    mbuf: Buf,
+    #[source]
+    error: ParseError<EthError>,
+}
+
+impl<Buf: PacketBufferMut> Packet<Buf> {
+    pub fn new(mbuf: Buf) -> Result<Packet<Buf>, InvalidPacket<Buf>> {
+        let (headers, consumed) = match Headers::parse(mbuf.as_ref()) {
+            Ok((packet, consumed)) => (packet, consumed),
+            Err(error) => {
+                return Err(InvalidPacket { mbuf, error });
+            }
+        };
+        Ok(Packet {
+            headers,
+            consumed,
+            mbuf,
+        })
+    }
+
+    pub(crate) fn reserialize(self) -> Buf {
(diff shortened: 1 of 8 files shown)


dataplane/src/packet.rs
```diff
@@ -0,0 +1,94 @@
...
+                mbuf
+            }
+            Ordering::Greater => {
+                let trim = self.consumed.get() - needed.get();
+                if trim > self.headers.size().get() {
+                    panic!("attempting to trim a nonsensical amount of data: {trim}");
+                }
+                match mbuf.trim_from_start(trim) {
+                    Ok(_) => {}
+                    Err(e) => unreachable!("configuration error: {:?}", e),
```
daniel-noland, line 65: Not blocking, but we will want to make an issue requiring us to confirm, at the management plane level, that we don't accept configurations which require more headroom than we actually get.

That might be a tricky chore tho.

daniel-noland (commented):
This all looks fine to me.

That said, I am not a good candidate for review since I helped compose this.

We should wait for others to chime in before this gets merged.

dataplane/src/packet.rs
daniel-noland: This is extra tricky if we can't count on having a specific nic.

We will need the dataplane to report available headroom in a given mbuf at startup and forward that the the management plane when it validates configs.

That complicates things because a config which is valid on one machine might be illegal on another network card.

It is likely worth a change to dpdk-sys to ensure we have 256 bytes of headroom just to decrease the likelihood of causing problems for customers.

That said, we will still _need_ the check regardless of how much headroom we configure.

mvachhar:
> > feat!(dataplane): Use Packet instead of MetaPacket
> Can you please provide more details? What's the motivation for this change? (It's nice to keep track of these in commit logs, by the way!)

Since `net::packet::Packet` has been renamed to `Headers`, we can rename `MetaPacket` to `Packet`.  Moreover, the new `Packet` interface manages its own buffer which will be useful as the interface is extended to deal with stripping outer packets, adding outer packets etc., though that support is not fully there yet.  

I'll the above text to the commit message.
 
> > a better Pipeline interface that can actually be tested and works with different Iterator types.
> 
> Tell me more! :) Could we have a few simple tests in the PR to make sure that this works with iterators, please? It would also help understand how pipeline stages will interact with the interface.

Take a look at [pr/mvachhar/dyn-pipeline](https://github.com/githedgehog/dataplane/tree/pr/mvachhar/dyn-pipeline).  The code still needs more reorganization, documentation, some more tests, and examples which I'm working on today.  There is a test in there though that shows how to use the DynChain component as well as static chaining.  There are a bunch of caveats to static chaining (hence the DynChain), but hopefully the docs will clarify when ready.

The iterators thing was non-trivial, but I think the Dyn stuff with DynIter will address most of the issues with iterator compatibility and dynamic pipeline stages (now dynamic network functions), the new network functions stuff will handle the static chaining case.  Sadly, too much static chaining causes the linker to run out of memory so there will be some caveats in how and when it can be used, which I'll document.

dataplane/src/packet.rs
mvachhar: I think the better answer here is probably to allow prechaining mbufs in the Packet interface instead of raising this error.  It'll cost performance but not cause issues for the customer.

qmonnet (approved):
Changes look OK to me, although I haven't looked in details at the fuzzing stuff.

Side notes:
- As discussed offline - We need to discuss about what conventions we use for `feat()` and for “breaking” changes. I don't consider the commits in this PR as user-facing breaking changes.
- If you reused code by Daniel, it's good practice to indicate it in the commit log. One good way to do that is using a tag/footer such as `Co-authored-by:`. At the very least, please keep Daniel's sign-off, given that he provided the code in the first place we should keep the relevant Certificate of Origin.

PR #262, opened 2025-02-21
ci: Add linter for making sure we follow the  Conventional Commits spec

qmonnet:
We have decided to enforce a structured formatting for the commit subject and description in the repository, namely the [Conventional Commit specification][0]. Add a CI workflow to make sure that new commits in Pull Requests abide by the specification.

I first tried to use https://github.com/wagoid/commitlint-github-action, which pulls by default the rules from [`@commitlint/config-conventional`][1], but I found them too restrictive. In particular, it enforces additional requirements on case that are not in the spec; worse, it unconditionally enforces maximum line lengths on subject, body and footer length. While I'm all in for restricting the subject's length, and generally wrapping the commit description, this makes it impossible to paste long links in the body or footer, or a large code snippet or command output in the body. So I went with https://github.com/marketplace/actions/conventional-commitlint instead, which seems to enforce the spec (and limit subject's length, which I believe is good) only.

We can switch to something more flexible in the future and refine our own rule set if necessary, but I find this seems to be a good start.

[0]: https://www.conventionalcommits.org/en/v1.0.0/
[1]: https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional

### DIFF
--- a/.github/workflows/lint-commitlint.yml
+++ b/.github/workflows/lint-commitlint.yml
@@ -1,0 +1,25 @@
+# Make sure commi subject and description follow the Conventional Commits spec.
+# See https://www.conventionalcommits.org/
+
+name: "lint-commitlint.yml"
+
+on:
+  pull_request: {}
+
+concurrency:
+  group: "${{ github.workflow }}:${{ github.event.pull_request.number }}"
+  cancel-in-progress: true
+
+jobs:
+  commitlint:
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 0
+          persist-credentials: "false"
+      - name: "Run commitlint on Pull Request's commits"
+        uses: "opensource-nepal/commitlint@v1"
+        with:
+          verbose: true


qmonnet:
Seems to work

image

Fredi-raspall (changes requested):
I'm going to block this .... :rofl: 
... until mid March. I have two ongoing branches with ~20 commits each, and don't feel like re-writing the history.

Plus, the linter should detect huge commits (esp. from Dan :see_no_evil: )

qmonnet:
> I'm going to block this .... 🤣 ... until mid March. I have two ongoing branches with ~20 commits each, and don't feel like re-writing the history.

@Fredi-raspall How about we enable the workflow now, but don't make it required for merging yet? This give us feedback but would allow us to bypass the workflow even if it fails, and “I've created all these commits before we changed the commit format” sounds like a reasonable reason to skip.

> Plus, the linter should detect huge commits (esp. from Dan 🙈 )

I get the idea but I think this is super hard to implement. How do you distinguish a medium-sized commit that packs multiple, unrelated things from a huge commit that is in fact just a tree-wide renaming or a clippy reports fixup?

Fredi-raspall:
> > I'm going to block this .... 🤣 ... until mid March. I have two ongoing branches with ~20 commits each, and don't feel like re-writing the history.
> 
> @Fredi-raspall How about we enable the workflow now, but don't make it required for merging yet? This give us feedback but would allow us to bypass the workflow even if it fails, and “I've created all these commits before we changed the commit format” sounds like a reasonable reason to skip.

@qmonnet , that works. Let's not make our lives harder.

> 
> > Plus, the linter should detect huge commits
> 
> I get the idea but I think this is super hard to implement. How do you distinguish a medium-sized commit that packs multiple, unrelated things from a huge commit that is in fact just a tree-wide renaming or a clippy reports fixup?

I was mostly teasing Dan on this one.

Fredi-raspall (approved):

.github/workflows/lint-commitlint.yml
```diff
@@ -0,0 +1,25 @@
+# Make sure commi subject and description follow the Conventional Commits spec.
```
Frostman, line 1: Commi!

qmonnet: https://github.com/githedgehog/dataplane/pull/264

PR #265, opened 2025-02-21
New pipeline implementation

mvachhar:
This is a new pipeline implementation that has a much nicer interface by leveraging Rust generics.

It should work correctly now with different iterator types, and should support wrapping iterators in `dyn_iter::DynIter` objects to support when the iterator type is unknown.

If someone can port their network function to this interface during review, it would be much appreciated.


mvachhar:
Given [this update](https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html) in Rust 2024 edition, do you think we need to change the signature of the `NetworkFunction`,  In particular, do we want a different requirement around `&’a mut Self`.  This bound allows capture of data from `Self` in the returned iterator, but the constraint is too strict if data from Self is never captured in the iterator.  I’m not sure it is a big deal, but thought it is worth considering removing the ‘a lifetime to see what the new default `impl Trait` behavior gets us.

dataplane/src/main.rs
```diff
@@ -102,7 +102,7 @@ fn init_devices(eal: &Eal) -> Vec<Dev> {
         .collect()
 }
 
-fn start_rte_workers(devices: &Vec<Dev>) {
+fn start_rte_workers(devices: &[Dev]) {
```
daniel-noland, line 99: non blocking note:

This change is perfectly fine here (and is likely best in this case).

In the more general case of a function which needs to accept a borrowed, fixed length, random access iterator, I generally write this:

```rust
fn start_rte_workers<T: AsRef<Dev>>(devices: T) {
    let devices = devices.as_ref();
    /* whatever */
}
```

That API can be more friendly (especially in the case of functions which accept any variant of `str`).

In this case this function is one-and-done so no need for the generic.

dataplane/src/pipeline/static_nf.rs
```diff
@@ -0,0 +1,107 @@
...
+
+/// Statically chains two [`NetworkFunction`] objects together.
+///
+/// The `chain` method takes two [`NetworkFunction`] objects and returns a new [`NetworkFunction`]
+/// that applies the first function, then the second.
+///
+/// This trait is automatically implemented for all objects that implement [`NetworkFunction`].
+///
+/// > [!WARNING]
+/// > Do not use long chains of statically chained network functions.
```
daniel-noland, line 54: The behavior I saw was more of the "find bugs in LLVM's symbol name length bounds."

I think the warning is good in spirit, but perhaps the details are not ideal.
Even if we explain the full issue, the situation may change and rot the warning.

Perhaps just leave it as "compiler bugs"?  Maybe that is worse.  Not sure :/

daniel-noland: Also, we might define "long"

dataplane/src/pipeline/sample_nfs.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+            packet
+        })
+    }
+}
+
+/// Network function that decrements the TTL value of an IP packet.
+///
+/// The function has no effect if the packet is not an IP packet.
+/// If the TTL is 0, an error is logged using [`trace!`].
```
daniel-noland, line 52: nit: I would recommend adding these docs on the process method rather than at the type level.

dataplane/src/pipeline/test_utils.rs
```diff
@@ -0,0 +1,101 @@
...
+use std::net::Ipv4Addr;
+
+use crate::packet::{InvalidPacket, Packet};
+use crate::pipeline::sample_nfs::{BroadcastMacs, DecrementTtl, InspectHeaders, Passthrough};
+use crate::pipeline::{DynNetworkFunction, nf_dyn};
+
+/// Generates an infinite sequence of network functions.
+///
+/// The sequence is a repeating pattern of:
+/// - [`InspectHeaders`]
```
daniel-noland, line 24: nit: this is another place where I would suggest putting the docs on the function rather than the type

dataplane/src/pipeline/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+//! This module provides the building blocks for constructing pipelines of network functions.
+//! There are two main methods provided for linking network functions together in sequence:
+//!
+//! - `StaticChain`: A trait for statically chaining network functions together.
+//! - `DynPipeline`: A pipeline that can be dynamically constructed at runtime.
+//!
+//! ## Network Functions
+//!
+//! A network function is any object that implements the [`NetworkFunction`] trait.
+//! You can look at the [`sample_nfs`] module for some examples of simple network functions.
```
daniel-noland, line 15: The word "object" is somewhat overloaded in the context of rust in that many folks use it to refer to anything which is `dyn` safe.  I would recommend swapping "object" for `struct` or instead say "A network function is anything that ..."

daniel-noland (commented):
I only had nitpicks so I would normally approve here.

That said, since I am co-author of some of this and it is a very important changeset I think it is best to get another set of eyes on it before we merege.

daniel-noland (commented):
see comments for minor nits (mostly about docs).

Overall, very nice work.

daniel-noland:
> Given [this update](https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html) in Rust 2024 edition, do you think we need to change the signature of the `NetworkFunction`, In particular, do we want a different requirement around `&’a mut Self`. This bound allows capture of data from `Self` in the returned iterator, but the constraint is too strict if data from Self is never captured in the iterator. I’m not sure it is a big deal, but thought it is worth considering removing the ‘a lifetime to see what the new default `impl Trait` behavior gets us.

I think the question is more about if we want to allow the implementation to capture from `Self` isn't it?  It seems like the answer is _no_, but I need to think about it

mvachhar:
> > Given [this update](https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html) in Rust 2024 edition, do you think we need to change the signature of the `NetworkFunction`, In particular, do we want a different requirement around `&’a mut Self`. This bound allows capture of data from `Self` in the returned iterator, but the constraint is too strict if data from Self is never captured in the iterator. I’m not sure it is a big deal, but thought it is worth considering removing the ‘a lifetime to see what the new default `impl Trait` behavior gets us.
> 
> I think the question is more about if we want to allow the implementation to capture from `Self` isn't it? It seems like the answer is _no_, but I need to think about it

That is exactly the question.  I think the answer is also no, you have to copy out of Self, but think about things like the final route destination, do we want to copy that out or just have a link to the appropriate data structure via Arc.  I guess if it uses Arc, then you wouldn’t borrow from self.

dataplane/src/pipeline/sample_nfs.rs
mvachhar: Really, what docs would apply to the entire trait then, since the whole purpose of the trait is to encapsulate that one function?

dataplane/src/main.rs
mvachhar: I suspect the interface manager is going to change some of this in main anyway so I vote to leave as is here.

mvachhar:
> I only had nitpicks so I would normally approve here.
> 
> That said, since I am co-author of some of this and it is a very important changeset I think it is best to get another set of eyes on it before we merege.

Agreed.  I want Quentin to try and stack his NAT changes on this to catch any interface design issues.

dataplane/src/pipeline/static_nf.rs
mvachhar: Done.  I chose 8 as long since that gave ok compile times and didn't OOM the linker.

dataplane/src/pipeline/mod.rs
mvachhar: Done.  I reworded to "anything"

dataplane/src/pipeline/sample_nfs.rs
daniel-noland: I would leave the first sentence in the trait doc and the rest on the function.

The reasoning is mostly around IDE integration.

My IDE (and I assume most others) have solid integration with rustdoc and I think it is more common to mouse over / press a hotkey on the function call than on the trait to see the docs.
That is mostly because the trait is often not visible as text at the call site.

daniel-noland:
> which does capture context since `process` is declared as `process(&self, datum)` (see this [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=b742fbf9b083a6e837db0b170489f34a)), I suggest we stick with what we've got now.

I think that makes sense.

We will likely continue to tune and refine these APIs so no reason to block on this now regradless.

Fredi-raspall:
> If someone can port their network function to this interface during review, it would be much appreciated.

On it!

qmonnet:
Dependabot's complaint is probably related to #274

mvachhar:
> Dependabot's complaint is probably related to #274

It is, I checked and asked @daniel-noland for a fix since I'm not quite sure how to achieve what he wants (and so didn't submit my own fix).

dataplane/src/pipeline/sample_nfs.rs
mvachhar: Fixed.

dataplane/src/pipeline/test_utils.rs
mvachhar: Fixed.

dataplane/src/main.rs
mvachhar: This got pushed in a cleanup PR, so is now moot here.

mvachhar:
@Fredi-raspall Quentin has approved these changes.  Do you have more feedback, or want another day to look at this or can we merge now?

Fredi-raspall:
> @Fredi-raspall Quentin has approved these changes. Do you have more feedback, or want another day to look at this or can we merge now?

It is good to go for me. I've opened https://github.com/githedgehog/dataplane/pull/290 with minor things / suggestions.

Fredi-raspall (approved):
Looks good to me @mvachhar !

PR #271, opened 2025-02-22
Clean up justfile

daniel-noland:
Boring cleanup of our build system.

This took a while, but it is likely worth it to maintain a healthy build system.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -168,21 +168,7 @@ jobs:
       - id: "build_each_commit"
         name: "build each commit"
         run: |
-          # Run a simple build for each separate commit (for "pull_request")
-          # or for the HEAD of the branch (other events).
-          set -eu -o pipefail
-          COMMITS=${{ github.sha }}
-          if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
-            # Get all commits from Pull Request, in chronological order
-            COMMITS=$(git rev-list --reverse ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }})
-          fi
-          for commit in $COMMITS ; do
-            git checkout $commit || exit 1
-            printf "::group::Build commit %s\n" "$(git log --oneline --no-decorate -n 1)"
-            (just debug_justfile="${{matrix.debug_justfile}}" cargo +${{matrix.rust.version}} build --locked --profile=dev --target=x86_64-unknown-linux-gnu) || exit 1
-            printf "::endgroup::\n"
-          done
-          printf "::notice::HEAD remains at %s\n" "$(git log --oneline --no-decorate -n 1)"
+          just debug_justfile="${{matrix.debug_justfile}}" rust="${{matrix.rust.version}}" build-sweep
         continue-on-error: ${{ matrix.rust.optional }}
 
       - name: "Note failure of optional steps"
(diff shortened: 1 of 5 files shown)


mvachhar (approved):
Just a question about documenting shell restrictions, but not a blocker.

justfile
```diff
@@ -4,24 +4,6 @@
 set unstable := true
 
 run_id := uuid()
-SHELL := shell("""
```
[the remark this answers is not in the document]
daniel-noland: hmmm, that is a reasonable idea.  Will make another docs pr.

.github/workflows/dev.yml
```diff
@@ -168,21 +168,7 @@ jobs:
...
-            COMMITS=$(git rev-list --reverse ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }})
-          fi
-          for commit in $COMMITS ; do
-            git checkout $commit || exit 1
-            printf "::group::Build commit %s\n" "$(git log --oneline --no-decorate -n 1)"
-            (just debug_justfile="${{matrix.debug_justfile}}" cargo +${{matrix.rust.version}} build --locked --profile=dev --target=x86_64-unknown-linux-gnu) || exit 1
-            printf "::endgroup::\n"
-          done
-          printf "::notice::HEAD remains at %s\n" "$(git log --oneline --no-decorate -n 1)"
+          just debug_justfile="${{matrix.debug_justfile}}" rust="${{matrix.rust.version}}" build-sweep
```
qmonnet, line 171: You didn't check that one in CI:

image

I'm also unsure `main` is defined when pulling the files with the `checkout` GitHub Action, you may have to provide the explicit commits as arguments to the just target.

And I'm concerned that this didn't cause the workflow to fail :confused:.

justfile
```diff
@@ -10,18 +10,25 @@ set dotenv-path := "."
...
+
 profile := "debug"
 [private]
 _container_repo := "ghcr.io/githedgehog/dataplane"
+
 # the rust channel to use (choose stable, beta, or nightly)
+
 rust := "stable"
 [private]
 _dpdk_sys_container_repo := "ghcr.io/githedgehog/dpdk-sys"
```
qmonnet: Having a blank line all around comments, but not between other declarations, looks weird and not easy to read. Do we really want that? Or should we at least use blank lines between all declarations for the paragraph, in that case?

.github/workflows/lint-just-fmt.yml
```diff
@@ -0,0 +1,46 @@
+# Make sure justfile source code is consistently formatted.
+
+name: "lint-just-fmt.yml"
+
+on:
+  pull_request: {}
+  merge_group:
+    types: ["checks_requested"]
```
qmonnet, line 8: Can we run this just on PRs that change the `justfile`? Either using the workflow syntax, or using a dedicated Action if we plan on marking this workflow as required.

qmonnet (commented):
@daniel-noland 

The CI at least needs fixing

.github/workflows/dev.yml
daniel-noland: oh, I only just saw this comment.  Will look into that because it sure seems like a problem.

justfile
daniel-noland: This is just how the stupid just format function works.  I also hate it.

.github/workflows/dev.yml
qmonnet: Trying a fix in https://github.com/githedgehog/dataplane/pull/294

qmonnet: Now merged

PR #273, opened 2025-02-23
Use documented version of free

daniel-noland:
Trivial swap, but it is better to use the non `_w` vesions of dpdk-sys functions as they include doxygen -> rustdoc docs.


mvachhar (approved):
Why did we ever use free_w instead?

daniel-noland:
> Why did we ever use free_w instead?

It was briefly needed because I had not tuned bindgen properly.

It just didn't get cleaned up.

PR #274, opened 2025-02-23
Minor improvements to dependabot

daniel-noland:
Mostly this should help keep our GitHub actions up to date.

I think this config is correct but it is hard to say because testing dependabot before merge is non trivial it seems.


qmonnet (approved):
Let's see if that works :) Thanks for updating this!

qmonnet:
image

PR #278, opened 2025-02-24
fix(dependabot): incorrect syntax

daniel-noland:
I wish there was an easier way to test this file before it hits main.

See [d59b3d3b963adca424766244ce643ec3db693f2b](https://github.com/githedgehog/dataplane/pull/274/commits/a41d16df37b2c9da570988df4959a4e8c489ded3) for the commit which introduced this bug.


qmonnet (approved):
Let's try this!

PR #279, opened 2025-02-24
Dependabot fix (again)

daniel-noland:
Once more with feeling


qmonnet (approved):

qmonnet:
I wonder if ChatGPT would be able to validate the Dependabot syntax if you pass it the file as input? :thinking:

PR #281, opened 2025-02-24
bump: Fix Dependabot's commits title message

qmonnet:
Dependabot uses "dep" as commit type, but the linter we have currently enforces the use of one of:

    build, ci, docs, feat, fix, perf, refactor, style, test, chore,
    revert, bump

We could change the linter, but for now let's simply change the commit type to "bump", which sounds accurate.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ updates:
   - package-ecosystem: "github-actions"
     target-branch: "main"
     commit-message:
-      prefix: "dep(github-actions):"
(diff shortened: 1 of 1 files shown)


daniel-noland:
oh, I duplicated your work in #282 I think.  Will fix that in a moment.

daniel-noland:
I'm fine with this change.  Will rebase this one and get it merged.

daniel-noland (approved):
rebased and LGTM.  Merging

PR #282, opened 2025-02-24
fix (and lint) dependabot.yml

daniel-noland:
Based on the recent nonsense PR sequence trying to update dependabot.yml, it seems reasonable to include this.

### DIFF
--- a/.github/workflows/lint-just-fmt.yml
+++ b/.github/workflows/lint-just-fmt.yml
@@ -3,7 +3,10 @@
 name: "lint-just-fmt.yml"
 
 on:
-  pull_request: {}
+  pull_request:
+    paths:
+      - "justfile"
+      - ".github/workflows/lint-just-fmt.yml"
   merge_group:
     types: ["checks_requested"]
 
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-just-fmt.yml
```diff
@@ -3,7 +3,10 @@
 name: "lint-just-fmt.yml"
 
 on:
-  pull_request: {}
+  pull_request:
+    paths:
+      - "justfile"
+      - ".github/lint-just-fmt.yml"
```
qmonnet, line 9:
```suggestion
      - ".github/workflows/lint-just-fmt.yml"
```

qmonnet (changes requested):

qmonnet:
> ✅dependabot config looks good 👍
> 

Will we get this every time? I'm not a fan of bot messages in the discussion when we can simply look at the status checks at the bottom of the page (although I know this workflow won't run often).

daniel-noland:
> Will we get this every time?

It should only run if the dependabot actually changed.  That said, it is trivial to remove

qmonnet:
I don't mind if we do remove :innocent:

daniel-noland:
> I don't mind if we do remove 😇

behold!

PR #283, opened 2025-02-24
bump(cargo): manual bump

daniel-noland:
Dependabot currently is running rust 1.84 so it can't bump our lockfile for us (due to edition change).

Running manually till they update it.


qmonnet (approved):
Looks good, thanks!

Issue #92, opened 2024-11-14
Clean up Git History


qmonnet:
In the context of #86 

This is not an Issue, but actually some sort of Pull Request, except that it's not a regular Pull Request, either. It's a request for review for branch https://github.com/githedgehog/dataplane/commits/new/qmonnet/clean-history/. The content of the tree in this branch is identical to `main`.

What matters here is the Git history. When compared to `main`, it's been reworked, to:

- Remove merge commits
- Reorder and squash commits that are very large and brings lots of fixes or successive additions/deletions of files, in order to simplify the history when individual commits don't help anyway
- Split docs changes out of the build framework large refactor commit
- Squash some commits that should go together
- Drop (now useless) commits that would update `flake.lock`, given that this file has now been removed
- Rewrite commit objects and descriptions for clarity and consistency

<details><summary>Interactive rebase operations</summary>

- reword 074ec6ae021f Basic working bindings
- fixup c71b25f7c068 Basic working flow install
- fixup 3867fb2b265e Connection tracking offloads are jacked!
- fixup 5c4fbdfa2195 Refactor
- fixup 3a6bdc41bb7f No std
- fixup d308013e3e86 Build config advancements
- fixup b14c0980fa9b Update .gitignore
- fixup af77a6fb3d3e Remove redundant dpdk-sys-old directory
- reword c36758ac33e6 Merge in design docs
- fixup 353782391283 Merge in design docs
- fixup fe57bbeba713 Merge in design docs
- fixup b67dfc0db20a Fix sysroot
- fixup c0af6c53c045 Fix broken sysroot
- drop e2ce55f1600e Update flake.lock
- drop 0b1251765dca Update flake.lock automatically
- drop 99405756f5da Revert "Update flake.lock"
- drop 4e151d2c1a92 flake.lock: Update
- drop bc3b90bc140f flake.lock: Update
- pick 36e59b511074 Bump cc in the dev-dependencies group across 1 directory
- reword 8512bbd0f14c Fix slop in dependency spec
- pick 4c150251f418 CI: Rename docs build workflow, restrict to docs changes
- pick 7229e105501c CI: Restrict docs workflow run on push to 'main' branch
- pick a6f0eff5a170 CI: Add concurrency group to docs workflow
- pick c7dac03103ac Bump libc in the dev-dependencies group across 1 directory
- pick 8fc09506f681 Bump cc in the dev-dependencies group across 1 directory
- drop 8952bca4ecfd flake.lock: Update
[...]

PR #285, opened 2025-02-24
bump(dpdk-sys): switch to non-musl version

daniel-noland:
We recently did a largeish rework of dpdk-sys.
It seems to work well for me based on current local testing but we should confirm in CI and on other dev machines before we move on.

If the commits of [dpdk-sys #92](https://github.com/githedgehog/dpdk-sys/pull/92) need to be reverted then it isn't going to get easier by waiting.


qmonnet (approved):
Looks OK

PR #290, opened 2025-02-25
Minor changes added when playing with the NF implementation

Fredi-raspall:
~~N.B. this targets https://github.com/githedgehog/dataplane/pull/265~~

### DIFF
--- a/dataplane/src/pipeline/dyn_nf.rs
+++ b/dataplane/src/pipeline/dyn_nf.rs
@@ -33,20 +33,21 @@ pub trait DynNetworkFunction<Buf: PacketBufferMut>: Any {
     /// Allow objects implementing [`DynNetworkFunction`] to identify themselves by overriding this.
     #[allow(dead_code)]
     #[allow(clippy::unnecessary_literal_bound)]
-    fn nf_name(&self) -> &str {
-        "anonymous-dynNF"
-    }
+    fn nf_name(&self) -> &str;
 }
 
 struct DynNetworkFunctionImpl<Buf: PacketBufferMut, NF: NetworkFunction<Buf> + 'static> {
     nf: NF,
+    name: String,
     _marker: PhantomData<Buf>,
 }
 
 impl<Buf: PacketBufferMut, NF: NetworkFunction<Buf>> DynNetworkFunctionImpl<Buf, NF> {
     pub fn new(nf: NF) -> Self {
+        let name = nf.nf_name().to_owned();
         Self {
             nf,
+            name,
             _marker: PhantomData,
         }
     }
@@ -69,6 +70,9 @@ pub fn nf_dyn<Buf: PacketBufferMut + 'static, NF: NetworkFunction<Buf> + 'static
 impl<Buf: PacketBufferMut, NF: NetworkFunction<Buf>> DynNetworkFunction<Buf>
     for DynNetworkFunctionImpl<Buf, NF>
 {
+    fn nf_name(&self) -> &str {
+        self.name.as_str()
+    }
     fn process_dyn<'a>(&'a mut self, input: DynIter<'a, Packet<Buf>>) -> DynIter<'a, Packet<Buf>> {
         self.nf.process(input).into_dyn_iter()
     }
--- a/dataplane/src/pipeline/dyn_nf.rs
+++ b/dataplane/src/pipeline/dyn_nf.rs
@@ -29,6 +29,13 @@ pub trait DynNetworkFunction<Buf: PacketBufferMut>: Any {
     /// [`DynPipeline::process`] with a concrete iterator type.  However, if you only
     /// have a dynamic iterator, you can use this method to process the packets.
     fn process_dyn<'a>(&'a mut self, input: DynIter<'a, Packet<Buf>>) -> DynIter<'a, Packet<Buf>>;
+
+    /// Allow objects implementing [`DynNetworkFunction`] to identify themselves by overriding this.
+    #[allow(dead_code)]
+    #[allow(clippy::unnecessary_literal_bound)]
+    fn nf_name(&self) -> &str {
+        "anonymous-dynNF"
+    }
 }
 
 struct DynNetworkFunctionImpl<Buf: PacketBufferMut, NF: NetworkFunction<Buf> + 'static> {
--- a/dataplane/src/pipeline/pipeline.rs
+++ b/dataplane/src/pipeline/pipeline.rs
@@ -64,6 +64,9 @@ impl<Buf: PacketBufferMut + 'static> DynPipeline<Buf> {
 }
 
 impl<Buf: PacketBufferMut> DynNetworkFunction<Buf> for DynPipeline<Buf> {
+    fn nf_name(&self) -> &str {
+        self.name.as_str()
+    }
     fn process_dyn<'a>(&'a mut self, input: DynIter<'a, Packet<Buf>>) -> DynIter<'a, Packet<Buf>> {
         self.nfs
             .iter_mut()
--- a/dataplane/src/pipeline/pipeline.rs
+++ b/dataplane/src/pipeline/pipeline.rs
@@ -17,12 +17,27 @@ use crate::pipeline::{DynNetworkFunction, NetworkFunction, nf_dyn};
 #[derive(Default)]
 pub struct DynPipeline<Buf: PacketBufferMut> {
     nfs: Vec<Box<dyn DynNetworkFunction<Buf>>>,
+
+    #[allow(unused)]
+    name: String,
 }
 
+#[allow(unused)]
 impl<Buf: PacketBufferMut + 'static> DynPipeline<Buf> {
-    #[allow(unused)]
     pub fn new() -> Self {
-        Self { nfs: vec![] }
+        Self {
+            nfs: vec![],
+            name: "anonymous-pipeline".to_owned(),
+        }
+    }
+    pub fn with_name(name: &str) -> Self {
+        Self {
+            nfs: vec![],
+            name: name.to_owned(),
+        }
+    }
+    pub fn name(&self) -> &str {
+        &self.name
     }
 
     /// Add a static network function to the pipeline.
@@ -42,7 +57,6 @@ impl<Buf: PacketBufferMut + 'static> DynPipeline<Buf> {
     ///
     /// [`DynNetworkFunction`]
     /// [`nf_dyn`]
-    #[allow(unused)]
     pub fn add_stage_dyn(mut self, nf: Box<dyn DynNetworkFunction<Buf>>) -> Self {
         self.nfs.push(nf);
         self
--- a/dataplane/src/pipeline/pipeline.rs
+++ b/dataplane/src/pipeline/pipeline.rs
@@ -3,6 +3,7 @@
 
 use dyn_iter::{DynIter, IntoDynIterator};
 use net::buffer::PacketBufferMut;
+use std::any::Any;
 
 use crate::packet::Packet;
 use crate::pipeline::{DynNetworkFunction, NetworkFunction, nf_dyn};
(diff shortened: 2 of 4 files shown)


dataplane/src/pipeline/pipeline.rs
```diff
@@ -17,12 +17,27 @@ use crate::pipeline::{DynNetworkFunction, NetworkFunction, nf_dyn};
...
+            name: "anonymous-pipeline".to_owned(),
+        }
+    }
+    pub fn with_name(name: &str) -> Self {
+        Self {
+            nfs: vec![],
+            name: name.to_owned(),
+        }
+    }
+    pub fn pipeline_name(&self) -> &str {
```
qmonnet, line 39: Probably no need to keep `pipeline` in `pipeline_name()`, maybe just `name()` or `get_name()`?

qmonnet (approved):
One (non-blocking) nit below, but otherwise looks good to me, thank you!

dataplane/src/pipeline/pipeline.rs
Fredi-raspall: Yes, makes sense! Thanks @qmonnet !

dataplane/src/pipeline/dyn_nf.rs
```diff
@@ -28,6 +28,13 @@ pub trait DynNetworkFunction<Buf: PacketBufferMut> {
     /// [`DynPipeline::process`] with a concrete iterator type.  However, if you only
     /// have a dynamic iterator, you can use this method to process the packets.
     fn process_dyn<'a>(&'a mut self, input: DynIter<'a, Packet<Buf>>) -> DynIter<'a, Packet<Buf>>;
+
+    /// Allow objects implementing [`DynNetworkFunction`] to identify themselves by overriding this.
+    #[allow(dead_code)]
+    #[allow(clippy::unnecessary_literal_bound)]
+    fn nf_name(&self) -> &str {
+        "anonymous-dynNF"
```
mvachhar, line 36: Is there a rust introspection feature that would allow us to calculate the type name of Self to print out?

I believe this is a good idea in concept, but I want to think about the api ergonomics a bit before approving.

dataplane/src/pipeline/pipeline.rs
```diff
@@ -17,12 +17,27 @@ use crate::pipeline::{DynNetworkFunction, NetworkFunction, nf_dyn};
...
+
+    #[allow(unused)]
+    name: String,
 }
 
+#[allow(unused)]
 impl<Buf: PacketBufferMut + 'static> DynPipeline<Buf> {
-    #[allow(unused)]
     pub fn new() -> Self {
-        Self { nfs: vec![] }
```
mvachhar, line 25: Should we just require a name here?

dataplane/src/pipeline/test_utils.rs
```diff
@@ -86,6 +87,7 @@ pub fn build_test_ipv4_packet(ttl: u8) -> Result<Packet<TestBuffer>, InvalidPack
     ipv4.set_source(UnicastIpv4Addr::new(Ipv4Addr::new(1, 2, 3, 4)).unwrap());
     ipv4.set_destination(Ipv4Addr::new(1, 2, 3, 4));
     ipv4.set_ttl(ttl);
+    ipv4.set_next_header_unchecked(NextHeader::UDP);
```
mvachhar, line 90: Why is this function called `_unchecked`?

mvachhar (changes requested):
It might be best to split out the stuff for the stage name, the tracing, and `set_next_header` into 3 PRs.  The `set_next_header` is easy to approve once I understand why you call the function unchecked.  The tracing one for the tests is fairly easy as well, though I'd like @daniel-noland to comment on if there is a better way to do this type of thing in rust.  The name thing I need to think about to see if we can just derive a property or something that makes setting stage names really easy and how that would affect this implementation.

dataplane/src/pipeline/dyn_nf.rs
mvachhar: Doing some homework, it may be best to just add this to the trait and then punt on the macros/introspection as it seems pretty complex and not worth it right now.  Especially because we only want a partial implementation of the trait.  @daniel-noland can you chime in on the effort required for something like 
```
#[derive(NetworkFunction)]
struct MyStage;
```

Where this would add the required trait and will use name as `MyStage`.  But later the user should be able to do
```
impl NetworkFunction for MyStage {
  fn process(...)
}
```

dataplane/src/pipeline/test_utils.rs
Fredi-raspall: There is a Ipv4::set_next_header() marked as unsafe. I believe the annotation is for the parsing case. I could use that one, but then nest it into an unsafe {} block.... but I see no point tbh. Ideally, the net crate would set this in the ip header when icmp/UDP/TCP etc. objects would be set... But that's not yet implemented, I believe. @daniel-noland ?

dataplane/src/pipeline/pipeline.rs
Fredi-raspall: I added a with_name() constructor. Fine to remove it, but that forces us to specify the name in the new() ctr, or None if we want to make it optional.

dataplane/src/pipeline/dyn_nf.rs
mvachhar: Let's leave it as is, fix it later.

dataplane/src/pipeline/test_utils.rs
mvachhar: I'm GTG once Daniel chimes in on this.

Fredi-raspall: If that helps, the ipv6 implementation is not unsafe:
```
    /// Set the next header [`IpNumber`]
    ///
    /// # Safety
    ///
    /// This method makes no attempt to ensure that the supplied [`next_header`] value is valid for
    /// the packet to which this header belongs (if any).
    ///
    /// [`next_header`]: NextHeader
    pub fn set_next_header(&mut self, next_header: NextHeader) -> &mut Self {
        self.0.next_header = next_header.0;
        self
    }
```
I can remove the unsafe annotation for the Ipv4 one and remove the _unchecked() method.
cc @mvachhar @daniel-noland

Fredi-raspall: @mvachhar I've done some corrections on the naming of NFs. I can break this PR into 3, but given the size of the PR I believe it's not worth it?

mvachhar:
@daniel-noland Can you take a look at this PR and give your feedback?

daniel-noland:
> @daniel-noland Can you take a look at this PR and give your feedback?

Sure.  I was waiting because it is marked as draft

net/src/ipv4/mod.rs
```diff
@@ -251,6 +251,13 @@ impl Ipv4 {
         self.0.protocol = next_header.0;
         self
     }
+
+    /// Set the next layer protocol without unsafe.
+    /// This method could be defined to accept [`IpNumber`].
+    pub fn set_next_header_unchecked(&mut self, next_header: NextHeader) -> &mut Self {
```
daniel-noland, line 257: I think there is some kind of misunderstanding because I don't understand the logic of this at all.

All the `unsafe` keyword in the extant `set_next_header` method indicates is that it is the caller's responsibility to maintain the invariants of the packet.  If you call `set_next_header` with `NextHeader` as TCP then you are responsible for ensuring that the transport layer of the packet object really does contain TCP.

It is `unsafe` precisely _because_ this struct _can't_ check that property since the header is the only thing in view.

If we want a safe version of this method it would need to be implemented on the `Headers` trait (at minimum).

If you wish to use the extant `set_next_header` method then you absolutely can.  You would need to put it in an `unsafe` block (which is completely fine IMO, so long as your call site maintains the invariant).

I fully agree that this is awkward in the context of constructing headers for testing.  This is why I implemented `TypeGenerator` (and friends) for `Headers` (although that also needs work to make things easy).

net/src/ipv4/mod.rs
```diff
@@ -251,6 +251,13 @@ impl Ipv4 {
         self.0.protocol = next_header.0;
         self
     }
+
+    /// Set the next layer protocol without unsafe.
+    /// This method could be defined to accept [`IpNumber`].
```
daniel-noland, line 256: This method _should not_ accept `IpNumber`.  There are two strong reasons for wrapping that type with `NextHeader`.

1. The [orphan rules](https://doc.rust-lang.org/reference/items/implementations.html?highlight=orphan#orphan-rules) prevent us from implementing `TypeGenerator` (or any other trait we don't own) for `IpNumber` (or any other type we don't own).  Using that type directly would complicate fuzz testing.
2. `etherparse` is not at 1.0 yet.  The rules for _this crate_ reaching 1.0 require that we not expose any types of any crate that is not at 1.0 or higher.

That said, removing etherparse types from the API of our `pub` interfaces is very much a work in progress.  Still, no reason (which I know of) to go backwards.

dataplane/src/pipeline/test_utils.rs
```diff
@@ -99,3 +101,13 @@ pub fn build_test_ipv4_packet(ttl: u8) -> Result<Packet<TestBuffer>, InvalidPack
 
     Packet::new(buffer)
 }
+
+/// Initializes tracing for tests that emit logs
+pub fn init_test_tracing() {
```
daniel-noland, line 106: I see what you are doing here (I did the same thing a little while back).

Then I found  [tracing_test](https://docs.rs/tracing-test/latest/tracing_test/) (which is already a workspace dependency).  You can skip this using `#[traced_test]` as an annotation on the test functions.

The [trap with this approach is described in the docs](https://docs.rs/tracing-test/latest/tracing_test/#rationale--why-you-need-this)

dataplane/src/pipeline/static_nf.rs
```diff
@@ -23,6 +23,13 @@ pub trait NetworkFunction<Buf: PacketBufferMut> {
         &'a mut self,
         input: Input,
     ) -> impl Iterator<Item = Packet<Buf>> + 'a;
+
+    /// Allow objects implementing [`NetworkFunction`] to identify themselves by overriding this.
+    #[allow(dead_code)]
+    #[allow(clippy::unnecessary_literal_bound)]
+    fn nf_name(&self) -> &str {
```
daniel-noland, line 30: I have no particular objection to the idea of an `nf_name` method.

That said, is the goal to make the `nf_name` a per-instance construct or a per implementation construct.

If the former, the I think this is basically correct.  If the later then I think you would want the return type to be `&'static str`

daniel-noland (changes requested):
I'm a little confused (as with the other PR in this sequence).

Ideally we get the chance to chat about the goal because I think one of us is mixed up.  Not sure which one tho

dataplane/src/pipeline/static_nf.rs
Fredi-raspall: It's meant to be per instance. I've been working on a setup that has two NFs of the same type chained and the name (which btw  should be unique (*)) should be able to distinguish both.

(*) = to ensure uniqueness, I'm considering that the add_nf() method or the like could actually return some form of Id that would ensure uniqueness; the code as it is now assumes the caller picks a distinct name per instance.

dataplane/src/pipeline/test_utils.rs
Fredi-raspall: Ack! Thanks @daniel-noland , will remove the method.

Fredi-raspall:
> > @daniel-noland Can you take a look at this PR and give your feedback?
> 
> Sure. I was waiting because it is marked as draft

Yes, I marked it as Draft since I added a method that does not work. The issue is related to https://github.com/githedgehog/dataplane/pull/311.

I've commented this on the commit in case we want to merge this regardless, until the issue gets fixed.

qmonnet:

PR #291, opened 2025-02-25
Add packet metadata

Fredi-raspall:
~~.. on top of https://github.com/githedgehog/dataplane/pull/290.~~

### DIFF
--- a/dataplane/src/packet_meta.rs
+++ b/dataplane/src/packet_meta.rs
@@ -3,6 +3,7 @@
 
 use routing::vrf::VrfId;
 use std::collections::HashMap;
+use std::net::IpAddr;
 
 #[derive(Debug, Default)]
 pub struct InterfaceId(u32);
@@ -30,7 +31,7 @@ impl BridgeDomain {
 
 #[allow(unused)]
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
-pub enum DropReason {
+pub enum DoneReason {
     InternalFailure,      /* catch-all for internal issues */
     NotEthernet,          /* could not get eth header */
     NotIp,                /* could not get IP header - maybe it's not ip */
@@ -47,20 +48,22 @@ pub enum DropReason {
     HopLimitExceeded,     /* TTL / Hop count was exceeded */
     Filtered,             /* The packet was administratively filtered */
     Unhandled,            /* there exists no support to handle this type of packet */
+    MissL2resolution,     /* adjacency failure: we don't know mac of some ip next-hop */
     Delivered,            /* the packet buffer was delivered by the NF - e.g. for xmit */
 }
 
 #[allow(unused)]
 #[derive(Debug, Default)]
 pub struct PacketMeta {
-    pub iif: InterfaceId,             /* incoming interface - set early */
-    pub oif: Option<InterfaceId>,     /* outgoing interface - set late */
-    pub is_l2bcast: bool,             /* frame is broadcast */
-    pub is_iplocal: bool,             /* frame contains an ip packet for local delivery */
-    pub vrf: Option<VrfId>,           /* for IP packet, the VRF to use to route it */
-    pub bridge: Option<BridgeDomain>, /* the bridge domain to forward the packet to */
-    pub drop: Option<DropReason>,     /* if Some, the reason why a packet was purposedly dropped.
-                                      This includes the delivery of the packet by the NF */
+    pub(crate) iif: InterfaceId,         /* incoming interface - set early */
+    pub(crate) oif: Option<InterfaceId>, /* outgoing interface - set late */
+    pub(crate) nh_addr: Option<IpAddr>,  /* IP address of next-hop */
+    pub(crate) is_l2bcast: bool,         /* frame is broadcast */
+    pub(crate) is_iplocal: bool,         /* frame contains an ip packet for local delivery */
+    pub(crate) vrf: Option<VrfId>,       /* for IP packet, the VRF to use to route it */
+    pub(crate) bridge: Option<BridgeDomain>, /* the bridge domain to forward the packet to */
+    pub(crate) done: Option<DoneReason>, /* if Some, the reason why a packet was marked as done.
+                                         This includes the delivery of the packet by the NF */
 
     #[cfg(test)]
     pub descr: &'static str, /* packet annotation (we may enable for testing only) */
@@ -73,7 +76,7 @@ pub struct PacketMeta {
 #[allow(unused)]
 pub struct PacketDropStats {
     pub name: String,
-    reasons: HashMap<DropReason, u64>,
+    reasons: HashMap<DoneReason, u64>,
     //Fredi: Todo: replace by ahash or use a small vec indexed by the DropReason value
 }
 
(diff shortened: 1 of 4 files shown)


dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,60 @@
...
+    InterfaceUnsupported, /* the operation is not supported on the interface */
+    VrfUnknown,           /* the vrf does not exist */
+    NatOutOfResources,    /* can't do NAT due to lack of resources */
+    RouteFailure,         /* missing routing information */
+}
+
+#[allow(unused)]
+#[derive(Debug, Default)]
+pub struct PacketMeta {
+    pub iif: InterfaceId,             /* incoming interface - set early */
```
mvachhar, line 51: As long as we understand this might change in the future.

mvachhar (commented):
I want another set of eyes before final approval.  If someone else approves I'm GTG.

dataplane/src/packet_meta.rs
Fredi-raspall: You mean the type InterfaceId ? Absolutely.

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    Unhandled,            /* there exists no support to handle this type of packet */
+    Delivered,            /* the packet buffer was delivered by the NF - e.g. for xmit */
+}
+
+#[allow(unused)]
+#[derive(Debug, Default)]
+pub struct PacketMeta {
+    pub iif: InterfaceId,             /* incoming interface - set early */
+    pub oif: InterfaceId,             /* outgoing interface - set late */
+    pub is_l2bcast: bool,             /* frame is broadcast */
```
daniel-noland, line 57: I'm not sure why we would want a field to describe this.

Shouldn't we just call `.is_boradcast()` or `.is_multicast()` on the ethernet header?

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    Filtered,             /* The packet was administratively filtered */
+    Unhandled,            /* there exists no support to handle this type of packet */
+    Delivered,            /* the packet buffer was delivered by the NF - e.g. for xmit */
+}
+
+#[allow(unused)]
+#[derive(Debug, Default)]
+pub struct PacketMeta {
+    pub iif: InterfaceId,             /* incoming interface - set early */
+    pub oif: InterfaceId,             /* outgoing interface - set late */
```
daniel-noland, line 56: This will want to be an `Option` since it is not assigned at first.

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub fn with_id(id: u32) -> Self {
+        Self(id)
+    }
+}
+
+#[allow(unused)]
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum DropReason {
+    InternalFailure,      /* catch-all for internal issues */
+    NotEthernet,          /* could not get eth header */
```
daniel-noland, line 35: I'm not sure `NotEthernet` is a valid branch.  As things stand, `Packet` doesn't even exist if it fails to parse even the eth header.

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+        self.0
+    }
+    pub fn with_id(id: u32) -> Self {
+        Self(id)
+    }
+}
+
+#[allow(unused)]
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum DropReason {
```
daniel-noland, line 33: Do we want one global `DropReason` enum?  Isn't this a per network function construct?

For example, would we extend this type with more errors for IPsec?  Then more for Wireguard?  Then more for ...

I need to think about this.

On big error type seems to somewhat undermine the utility of exhaustive pattern matching

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    Delivered,            /* the packet buffer was delivered by the NF - e.g. for xmit */
+}
+
+#[allow(unused)]
+#[derive(Debug, Default)]
+pub struct PacketMeta {
+    pub iif: InterfaceId,             /* incoming interface - set early */
+    pub oif: InterfaceId,             /* outgoing interface - set late */
+    pub is_l2bcast: bool,             /* frame is broadcast */
+    pub is_iplocal: bool,             /* frame contains an ip packet for local delivery */
```
daniel-noland, line 58: this seems like something that needs to be a method rather than a bool

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub route: Option<VrfId>,         /* frame is IP packet to be routed */
```
daniel-noland, line 59: route seems like a strange name for something with is `Option<VrfId>`.  I don't understand this field

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub bridge: Option<BridgeDomain>, /* frame is to be bridged over a bridge domain */
+    pub drop: Option<DropReason>,     /* if Some, the reason why a packet was purposedly dropped.
+                                      This includes the delivery of the packet by the NF */
+
+    #[cfg(test)]
+    pub descr: &'static str, /* packet annotation (we may enable for testing only) */
```
daniel-noland, line 65: I don't understand this field.  Attaching a `'static str` to a packet is confusing to me

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    #[cfg(test)]
+    /* Keep the Packet in spite of calling packet.fate(). This is for testing */
+    pub keep: bool,
+}
+
+#[derive(Default, Debug)]
+#[allow(unused)]
+pub struct PacketDropStats {
```
daniel-noland, line 73: I see and appreciate what you are doing here.  We will want to make sure this integrates with the tracing /  metrics stack so that we don't need to manually write integrations with graphana and stuff

daniel-noland (changes requested):
Given the number of questions I had on review, I'm not 100% sure I follow the whole goal.

I suspect a call to discuss is in order.  It is very possible I just don't see the plan here.

Fredi-raspall:
> Given the number of questions I had on review, I'm not 100% sure I follow the whole goal.
> 
> I suspect a call to discuss is in order. It is very possible I just don't see the plan here.

Happy to have the discussion!

dataplane/src/packet_meta.rs
Fredi-raspall: It's actually set by doing that (actually comparing against dmac == Mac::BROADCAST, but that can be easily fixed). This is just an annotation so that the same check need not be done later.
Here I am assuming that for debugging, dumping the meta-data may be easier to read than dumping the whole packet.

dataplane/src/packet_meta.rs
Fredi-raspall: Fair enough! Will change.

dataplane/src/packet_meta.rs
Fredi-raspall: Happy to have some hierarchy. I'd use a single type, though. You could have some hierarchy by nesting enums and have a category Ipsec, another for routing, etc.. But we have to start by something.

dataplane/src/packet_meta.rs
Fredi-raspall: A method of what type?

dataplane/src/packet_meta.rs
Fredi-raspall: Happy to rename. Let me explain the purpose and then we decide the name.
When receiving a packet, we need a way to know what vrf to look up, right?
That vrf may be identified by some vrfId. Now, since the packet processing is a sequence of stages, an early stage (ingress) may do this and then annotate it to a forwarding stage that would use the annotation to perform the lookup in such a vrf.

dataplane/src/packet_meta.rs
Fredi-raspall: The purpose of that was for testing. A numerical id could be used instead.
The idea is (was) that you could name crafted, test packets (e.g. "this is a martian and should be dropped") and then in the logs display it. Again, can be removed or replaced. But note the `#[cfg(test)]`

dataplane/src/packet_meta.rs
Fredi-raspall: Oh, great idea!

dataplane/src/packet_meta.rs
Fredi-raspall: ... or are you saying that we would not get to see the packet at all ?

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+    RouteFailure,         /* missing routing information */
+    HopLimitExceeded,     /* TTL / Hop count was exceeded */
+    Filtered,             /* The packet was administratively filtered */
+    Unhandled,            /* there exists no support to handle this type of packet */
+    Delivered,            /* the packet buffer was delivered by the NF - e.g. for xmit */
+}
+
+#[allow(unused)]
+#[derive(Debug, Default)]
+pub struct PacketMeta {
```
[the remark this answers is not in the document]
Fredi-raspall: Ok. Will revisit! Thx!

dataplane/src/packet_meta.rs
Fredi-raspall: I've renamed it to vrf.

dataplane/src/packet_meta.rs
```diff
@@ -0,0 +1,125 @@
...
+use std::collections::HashMap;
+
+#[derive(Debug, Default)]
+pub struct InterfaceId(u32);
+#[allow(unused)]
+impl InterfaceId {
+    pub fn get_id(&self) -> u32 {
+        self.0
+    }
+    pub fn set_id(&mut self, id: u32) {
```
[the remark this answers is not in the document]
Fredi-raspall: Removed in favor of constructor.

dataplane/src/packet_meta.rs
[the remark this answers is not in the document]
Fredi-raspall: I've changed all of them to pub(crate).
Atm, the meta field in Packet is pub. We may change that too to pub(crate) or private and then have accessors, like
```
pub fn get_meta_mut(&mut self) -> &mut PacketMeta {
    &mut self.meta
}
pub fn get_meta(&self) -> &PacketMeta {
    &self.meta
}
```

Let me know what you think is best.

PR #292, opened 2025-02-25
Headers: Add TryIp, TryTransport (+ ...Mut)

qmonnet:
- **feat(headers): Add TryIp/TryIpMut to get IP-agnostic ref to L3 header**
- **feat(headers): Add TryTransport, TryTransportMut**


dataplane/src/pipeline/pipeline.rs
```diff
@@ -148,5 +148,14 @@ mod test {
             (p2_ttl as usize) - DynStageGenerator::num_ttl_decs(num_stages),
             p2_out.try_ipv4().unwrap().ttl() as usize
         );
+
+        // Check try_ip() and try_ipv4() are consistent
+        let p1_ipv4 = p1_out.try_ipv4().expect("Expected IPv4 packet");
```
mvachhar, line 153: I'm not a fan of testing this here instead of net/src/headers.rs but not worth holding things up over it.

mvachhar (approved):

dataplane/src/pipeline/pipeline.rs
qmonnet: Yep, I thought about adding tests to header.rs but having something just for `TryIp` felt weird. Adding for all other traits felt beyond the scope. We can probably refactor these at some point. Thanks!

PR #298, opened 2025-02-25
Bump 2: electric boogaloo 

daniel-noland:
Once again, I try to calm dependabot.

Mostly by depending on it for less.

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -39,8 +39,6 @@ jobs:
           password: "${{ secrets.GITHUB_TOKEN }}"
       - name: "Checkout"
         uses: "actions/checkout@v4"
-      - name: "install rust"
-        uses: "dtolnay/rust-toolchain@stable"
       - name: "install envsubst"
         run: |
           sudo apt-get update
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -57,7 +57,7 @@ jobs:
         uses: "peter-evans/create-pull-request@v7"
         with:
           branch: "bump/dpdk-sys"
-          title: "build(dpdk-sys): bump"
+          title: "bump(dpdk-sys): new-version"
           labels: |
             automated
             dependencies
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -41,12 +41,6 @@ jobs:
         uses: "actions/checkout@v4"
       - name: "install rust"
         uses: "dtolnay/rust-toolchain@stable"
-      - name: "install cargo-deny"
-        run: |
-          cargo install cargo-deny
-      - name: "deny check"
-        run: |
-          cargo deny check
       - name: "install envsubst"
         run: |
           sudo apt-get update
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -4,6 +4,9 @@
 name: "bump.yml"
 
 on:
+  pull_request:
+    paths:
+      - ".github/workflows/bump.yml"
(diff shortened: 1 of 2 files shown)


.github/workflows/bump.yml
```diff
@@ -72,3 +72,75 @@ jobs:
         timeout-minutes: 60
         with:
           limit-access-to-actor: true
+
+  cargo-upgrades:
+    runs-on: "lab"
```
qmonnet, line 72: Does this need to run in the lab?

qmonnet (approved):
Plug it on :electric_plug: :+1:

.github/workflows/bump.yml
daniel-noland: the compile is really slow on the hosted runners.  It took like 6 minuts just to build cargo-deny

daniel-noland:
Look on my works, ye Mighty, and despair!

Also, look at this #302 and share your thoughts.

Note that the bump bump failures are expected due to the way GH actions works.

The base branch isn't set unless you merge to main so those failures during dev are expected.  They should vanish once this hits main.

.github/workflows/bump.yml
qmonnet: Can't we find a binary for cargo-deny? Or a Docker image? We should probably avoid recompiling it everytime if we could.

But OK

.github/workflows/bump.yml
```diff
@@ -72,3 +67,81 @@ jobs:
...
+        # _resolves_ the license / security issue we have had / would have had without the upgrade.
+        run: |
+          cargo deny check
+        continue-on-error: true
+      - name: "cargo upgrade"
+        id: upgrade
+        run: |
+          cargo update
+          echo "### :rocket: Upgrades available" > upgrade.log
+          echo "" >> upgrade.log
```
qmonnet, line 106:
```suggestion
          echo "" >> upgrade.log
          echo "```" >> upgrade.log
```

.github/workflows/bump.yml
```diff
@@ -72,3 +67,81 @@ jobs:
...
+          cargo update
+          echo "### :rocket: Upgrades available" > upgrade.log
+          echo "" >> upgrade.log 
+          cargo upgrade --incompatible=allow  >> upgrade.log
+          cat upgrade.log > "${GITHUB_STEP_SUMMARY}"
+          {
+            echo 'upgrade<<EOF';
+            cat upgrade.log;
+            echo 'EOF';
+          } >> "${GITHUB_OUTPUT}"
```
qmonnet, line 113:
```suggestion
          } >> "${GITHUB_OUTPUT}"
          echo "```" >> upgrade.log
```

qmonnet (commented):
How about adding backticks around the table? This won't change a thing for the commit log, but the PR description should be better formatted (monospace font to keep the formatting of the table)

.github/workflows/bump.yml
daniel-noland: That is a good idea.  As I recall binstall was broken on the lab machines for some reason but it would be fine on the hosted runners I think

qmonnet: Otherwise they publish binaries on the GH repo (https://github.com/EmbarkStudios/cargo-deny/releases/latest), although now I think about it this is maybe where `binstall` might be fetching them.

daniel-noland:
Ok, it turns out that

1. we can use binstall
2. the hosted runners are also super slow to build binstall (which is crazy).  It took them almost 4 minutes just to build that.
3. the fast run is to use binstall and the hosted runner.

PR #308, opened 2025-02-26
Fix Clippy's pedantic reports in dataplane package

qmonnet:
- **chore(dataplane): Fix pedantic Clippy warnings**
- **chore(dataplane): Deny pedantic Clippy warnings for dataplane package**


daniel-noland (approved):
All clear improvements as far as I can see.  Nice work.

Issue #309, opened 2025-02-26
Rust Nighly: `cargo clippy` fails


qmonnet:
Running `cargo clippy` on the workspace with the Nightly toolchain fails (and has always failed, it seems). From the logs, this looks like a compiler error:

```
Run just debug_justfile="false" rust=nightly cargo clippy
   Compiling dpdk-sysroot-helper v0.0.1 (/home/runner/_work/dataplane/dataplane/dpdk-sysroot-helper)
    Checking net v0.0.1 (/home/runner/_work/dataplane/dataplane/net)
    Checking dataplane-errno v0.1.0 (/home/runner/_work/dataplane/dataplane/errno)
   Compiling dpdk-sys v0.0.1 (/home/runner/_work/dataplane/dataplane/dpdk-sys)
   Compiling dpdk v0.1.0 (/home/runner/_work/dataplane/dataplane/dpdk)

thread 'rustc' panicked at /rustc/85abb276361c424d64743c0965242dd0e7b866d1/compiler/rustc_hir/src/hir.rs:1165:18:
can't get the span of an arbitrary parsed attribute: Parsed(Repr([(ReprInt(SignedInt(I32)), errno/src/lib.rs:542:8: 542:11 (#0))]))
stack backtrace:
   0:     0x73caf2bb4864 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hfbcadcb664e1e745
   1:     0x73caf341622c - core::fmt::write::h96d3ad630dab013b
   2:     0x73caf44683d1 - std::io::Write::write_fmt::h3607b096b8efb3ee
   3:     0x73caf2bb46c2 - std::sys::backtrace::BacktraceLock::print::h5c5089ef3846b40c
   4:     0x73caf2bb6ece - std::panicking::default_hook::{{closure}}::ha2f1ea5200a78a69
   5:     0x73caf2bb6aa4 - std::panicking::default_hook::he72a4c09abcc8836
   6:     0x73caf1d1c4a7 - std[fe011135ce22c601]::panicking::update_hook::<alloc[d9798e16cf59ec93]::boxed::Box<rustc_driver_impl[685633a5d1c26086]::install_ice_hook::{closure#1}>>::{closure#0}
   7:     0x73caf2bb7743 - std::panicking::rust_panic_with_hook::h9c6e02ac63684bb2
[...]

PR #310, opened 2025-02-26
ci: Disable "cargo clippy" with Nightly toolchain in dev workflow

qmonnet:
The step fails, presumably because of a compiler bug. As a result, all new Pull Requests get messages telling that the workflow has failed. This brings no useful info. If the step is broken for Nightly, let's as well disable it for now.

Link: https://github.com/githedgehog/dataplane/issues/309

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -142,6 +142,9 @@ jobs:
         continue-on-error: ${{ matrix.rust.optional }}
 
       - id: "clippy"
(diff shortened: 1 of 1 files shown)


daniel-noland (approved):
Yup.  It sure is broken at the moment.

We will likely want to revert after the next dpdk-sys bump (which will bring a new nightly)

Kinda impressive that we have been at this for months and this is the first broken nightly.

PR #311, opened 2025-02-27
Make `DynNetworkFunction` derive `Any`

mvachhar:
Make it so that `DynNetworkFunction` implements `Any`.  The only constraint is that `DynNetworkFunction` has to be bound by `'static` and for that to happen, `PacketBufferMut` should be `'static`.  

The one change here is that any `NetworkFunction` that is to be put into a `DynPipeline` also has to be bound by `'static`.  That was the case anyway in the public API, but it is now true internally as well.

This PR is a precursor for some methods on `DynPipeline` that will allow pipeline stage lookup and downcast back to a concrete pipeline stage, though we need to do some other things for that to work, in particular, move to Rust 1.86 beta until April when 1.86 should be released.

Non-`'static` `NetworkFunction`s can still be statically chained.

### DIFF
--- a/dataplane/src/pipeline/dyn_nf.rs
+++ b/dataplane/src/pipeline/dyn_nf.rs
@@ -3,6 +3,7 @@
 
 use dyn_iter::{DynIter, IntoDynIterator};
 use net::buffer::PacketBufferMut;
+use std::any::Any;
 use std::marker::PhantomData;
 
 use crate::packet::Packet;
@@ -17,7 +18,7 @@ use crate::pipeline::NetworkFunction;
 ///
 /// [`nf_dyn`]
 /// [`crate::pipeline::DynPipeline`]
-pub trait DynNetworkFunction<Buf: PacketBufferMut> {
+pub trait DynNetworkFunction<Buf: PacketBufferMut>: Any {
     /// The `process_dyn` method takes an iterator of [`crate::packet::Packet`] objects,
     /// However, unlike [`NetworkFunction::process`], this method does not require concrete
     /// iterator types.
@@ -30,7 +31,7 @@ pub trait DynNetworkFunction<Buf: PacketBufferMut> {
     fn process_dyn<'a>(&'a mut self, input: DynIter<'a, Packet<Buf>>) -> DynIter<'a, Packet<Buf>>;
 }
 
-struct DynNetworkFunctionImpl<Buf: PacketBufferMut, NF: NetworkFunction<Buf>> {
+struct DynNetworkFunctionImpl<Buf: PacketBufferMut, NF: NetworkFunction<Buf> + 'static> {
     nf: NF,
     _marker: PhantomData<Buf>,
 }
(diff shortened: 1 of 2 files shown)


qmonnet:
> any `NetworkFunction` that is to be put into a `DynPipeline` also has to be bound by `'static`

I have a rough understanding of what the `'static` lifetime means, but I'm not sure I understand the exact consequences of the change for a `NetworkFunction`. Any change in particular that we should expect?

daniel-noland (approved):
I have no issue with this code in so far as it goes.  I'm not 100% sure I get the overall objective.  This is related to the need to introspect the pipeline from the CLI as I understand things, but I don't know the full story of that effort yet.

Fredi-raspall:
> I have no issue with this code in so far as it goes. I'm not 100% sure I get the overall objective. This is related to the need to introspect the pipeline from the CLI as I understand things, but I don't know the full story of that effort yet.

That's one possible use case. The general objective is:
* pipeline owns the NF's added to it, but loses their types as gets trait objects.
* Since pipeline is built to own the NFs', it is legit to be able to query it to provide access to them (access = some reference). This works (e.g. when using the nf_name as key for lookups). The problem is that this returns a ref to a dyn Networkfunc trait, which is a bit useless for callers if they cannot downcast to the original type. The PR is work to make that possible. Ie. to be able to downcast to the particular type that implements the NF trait.

mvachhar:
> > any `NetworkFunction` that is to be put into a `DynPipeline` also has to be bound by `'static`
> 
> I have a rough understanding of what the `'static` lifetime means, but I'm not sure I understand the exact consequences of the change for a `NetworkFunction`. Any change in particular that we should expect?

From the perspective of current users of the API, the `NetworkFunction` already had to be `'static` so it has no meaningful impact on users of the interface.

An object is bound by `'static' if either it has static lifetime (e.g., lives for the life of the program), or has only primitive data types, or all of its fields are 'static.  Essentially this means that for non-static-lifetime objects, an object has to own all its data and be able to destroy all fields when it is deallocated.  Note that this is a little tricky to interpret in the case that the object has reference counted pointers, but note that the object in that case owns the pointer, not the underlying object.

mvachhar:
> I have no issue with this code in so far as it goes. I'm not 100% sure I get the overall objective. This is related to the need to introspect the pipeline from the CLI as I understand things, but I don't know the full story of that effort yet.

Another way to look at this is this is how it should have been from the start.  Since `DynPipeline` owns its pipeline stages, there should be a way to get them back out if we need them, for whatever purpose.

Next steps will be to have methods to get references to the pipeline stages back out based on id or name.  @Fredi-raspall PR adds a name to each Pipeline stage, but I think a better approach there will be to use `add_stage_with_id` to give the stage an ID and then allow selection by type using turbo-fish and typeids.

I'll submit that change in a later PR after we get the Rust beta that has the upcast feature we need to actually implement that stuff.

PR #313, opened 2025-02-27
fix: use `produce` instead of `r#gen`

daniel-noland:
We can switch to a better method name now that https://github.com/camshaft/bolero/pull/270 is merged and bolero 0.13 is available.


qmonnet (approved):
Let's do that!

PR #322, opened 2025-03-01
Implement stateless NAT44 (not functional yet)

qmonnet:
Add a preliminary, not-yet-functional implementation of stateless NAT44.

### The following restrictions currently apply:

- Only stateless NAT is supported for now.

- Only NAT44 is supported for now.

- Either source or destination NAT are supported, only one at at time, by a given Nat object.

- PIFs mixing IPv4 and IPv6 endpoints or list of exposed IPs are not supported for now.

- For a given PIF used for NAT, the number of IP addresses covered by the full set of (VPC-internal) endpoint prefixes must be equal to the numer of IP addresses covered by the full set of externally-exposed IP prefixes; this is in order to make the 1:1 address mapping work.

Most importantly, **the VNI lookup has not been implemented yet**, which makes the code non-functional at the moment. A hardcoded VNI is passed in `.process_packet()`, so that we can test the rest of the code in a futher commit.

### Internals

Internally, stateless NAT perform a translation based on a 1:1 IP address mapping between two sets of IPs:

- The NAT processor looks up for the source and destination PIFs by looking up at the VNI and destination address of the packet: the VNI indicates the source VPC, and a (VNI, destination) combination corresponds to a single possible destination PIF from this VPC, attached to a single source PIF for the VPC.

- For source NAT, the NAT processor looks at the source PIF, searches the list of endpoints of this PIF for the source address of the packet, thus retrieving an index in the list of endpoint IPs; then it translates the source address into the address at the same index in the list of publicly-exposed IPs for the source PIF.

- For destination NAT, the NAT processor looks at the destination PIF, searches the list of publicly-exposed IP addresses for the destination of the packet, thus retrieving an indes in the list of exposed IPs; then it translates the destination address into the the address at the same index in the list of endpoint IPs for the destination PIF.

- The lists of endpoint or publicly-exposed IP addresses rely on the IpList interface implemented in a previous commit.

### Follow-ups to this PR include:

- Implementing the VNI lookup

- Proper error handling

- Concurrency safety

- Adding in-code documentation

- Adding configuration, moving fabric objects creation to configuration stage; moving part of the checks on fabric objects to configuration stage

- Optimise: Improve structures cross-referencing, retrieve references to relevant objects rather than names and suppress the lookup by name, etc.

- Some clean-up here and there, possibly collapsing some structs (nat/fabric.rs's PifTable or nat/mod.rs's GlobalContext are candidates)

- Lifting restrictions: IPv4/IPv6 support, stateful NAT

- Possibly rework the API: Offer different Nat objects instead of configuring with mode/destination?

Best reviewed per-commit.


Fredi-raspall:
> Most importantly, the VNI lookup has not been implemented yet, which makes the code non-functional at the moment. A hardcoded VNI is passed in .process_packet(), so that we can test the rest of the code in a futher commit.

I'm adding this as part of the routing functionality. We should sync up at some point.

dataplane/src/nat/fabric.rs
```diff
@@ -0,0 +1,419 @@
...
+    #[tracing::instrument(level = "trace")]
+    fn new() -> Self {
+        Self {
+            pifs: HashMap::new(),
+            endpoint_trie: PrefixTrie::new(),
+        }
+    }
+
+    #[tracing::instrument(level = "info")]
+    fn add_pif(&mut self, pif: Pif) -> Result<(), TrieError> {
```
Fredi-raspall, line 188: If Piftable is a global table containing all of the PIFs, I believe that the key should be (vpc, pifname), right? Otherwise, there may be name collisions. IMO, each VPC should be allowed to choose the pif names.

dataplane/src/nat/fabric.rs
```diff
@@ -0,0 +1,419 @@
...
+        if self.pifs.contains_key(&pif.name) {
+            return Err(TrieError::EntryExists);
+        }
+
+        for prefix in &pif.endpoints {
```
Fredi-raspall, line 193: I believe this should be ip's instead of endpoints? Endpoints are not guaranteed to be unique, whereas ip's should.

Fredi-raspall: ..or am I missing something?

dataplane/src/nat/fabric.rs
Fredi-raspall: Oh, now I see that pif table is per VPC. Ignore my comment!

dataplane/src/nat/fabric.rs
Fredi-raspall: Ok. Ignore this too, as I believe it's fine if iftable is per VPC

dataplane/src/nat/fabric.rs
```diff
@@ -0,0 +1,419 @@
...
+    #[tracing::instrument(level = "trace")]
+    pub fn find_prefix(&self, ip: &IpAddr) -> Option<&Prefix> {
+        self.iter_endpoints().find(|&prefix| prefix.covers_addr(ip))
+    }
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct PeeringPolicy {
+    name: String,
+    vnis: [Vni; 2],
```
Fredi-raspall, line 83: Shouldn't this be a pair of VPCs, instead of their VNIs?

dataplane/src/nat/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+            global_pif_trie: PrefixTrie::new(),
+            peerings: HashMap::new(),
+        }
+    }
+
+    #[tracing::instrument(level = "trace")]
+    fn insert_vpc(&mut self, vni: Vni, vpc: Vpc) {
+        vpc.iter_pifs().for_each(|pif| {
+            pif.iter_ips().for_each(|prefix| {
+                let _ = self.global_pif_trie.insert(prefix, pif.name().clone());
```
Fredi-raspall, line 48: Shouldn't this check that a pif hasn't been already added for the prefix / ip?
Else, a pif may overwrite the ips of another one?

dataplane/src/nat/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum NatMode {
+    Stateless,
+    #[allow(dead_code)]
+    Stateful,
+}
+
+#[derive(Debug)]
+pub struct Nat {
+    context: GlobalContext,
```
Fredi-raspall, line 138: I believe that the global context should not be owned by the NAT module but just be referred to it somehow? The VPCs, pifs, policies, etc.. pertain to a wider scope in general IMO.
Also, if we plan to support multiple NAT stages (since your code assumes that a NAT object has a single direction; I'm assuming we'd instantiate more than one), this means that the NAT cannot own the context. Else, we'd need to replicate the context for each one?

dataplane/src/nat/fabric.rs
```diff
@@ -0,0 +1,419 @@
...
+        self.peerings.iter()
+    }
+
+    #[tracing::instrument(level = "trace")]
+    pub fn add_endpoint(&mut self, endpoint: Prefix) {
+        self.endpoints.push(endpoint);
+    }
+
+    #[tracing::instrument(level = "trace")]
+    pub fn add_ip(&mut self, ip: Prefix) {
```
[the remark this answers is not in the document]
qmonnet: > I believe there should be a VPC table somewhere.
That table would include a trie of ips such that a given ip would only pertain to a Pif (without duplicates).

We have that slightly differently, as a global hash table for prefixes that tells, for each prefix, what is the corresponding VPC. This is `Nat.context.global_pif_trie`, that we update when creating a new VPC:

```rust
    fn insert_vpc(&mut self, vni: Vni, vpc: Vpc) {
        vpc.iter_pifs().for_each(|pif| {
            pif.iter_ips().for_each(|prefix| {
                let _ = self.global_pif_trie.insert(prefix, pif.name().clone());
            });
        });
        let _ = self.vpcs.insert(vni.as_u32(), vpc);
    }
```

> When a new pif would be added, before calling this, we should check that the ip is not in that trie and otherwise fail, right?

Yes, we should enforce this. But I've not tried to enforce these kind of checks at VPC, PIF, PeeringPolicy creation for now. I expect that these objects will be shared to some extent with other `NetworkFunction`s, and that their creation will be part of some config step that is not currently implemented. At the moment, the insert methods that I have are mostly to do tests, really.

So once we have a more robust configuration interface, yes, we should ensure that all exposed IPs are unique (and that VPC names are unique, and PIF names for a given VPC, and avoid having duplicate endpoints/IPs for a PIF, and so on), and we'll be able to remove some of the checks from the NAT logics. I don't plan to address it in the current PR, though

dataplane/src/nat/mod.rs
qmonnet: Yes, absolutely. See my previous comment - I expect all of the code for object creation to move to some config component, where we'll make sure to enforce such safety checks. Right now these were mostly implemented for testing.

Fredi-raspall (dismissed):
Nice work @qmonnet !
The scope of this PR is large and not only related to NAT.
Most of my comments pertain to the overall model.
We should discuss this.
cc @mvachhar

dataplane/src/nat/fabric.rs
qmonnet: Here the VNIs are keys to retrieve the VPCs, so indirectly, that's what we have, sort of. I suppose it would be more explicit to have VPC names, although I don't like referencing them by names a lot.

Ideally, I'd like to have a direct reference to the `Vpc` object here, but this requires some wrapping with refcounts/Cells and so on, as the VPC objects are already owned by the `GlobalContext` object. Same things for the PIFs, I'd like to get a direct reference to them from the `PeeringPolicy` and save the hashmap lookup. I expect to do more work in that direction, once we agree the overall structure is acceptable and this PR gets merged.

dataplane/src/nat/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+use net::ipv6::UnicastIpv6Addr;
+use net::vxlan::Vni;
+use std::collections::HashMap;
+use std::fmt::Debug;
+use std::net::IpAddr;
+
+#[derive(Debug)]
+#[allow(dead_code)]
+struct GlobalContext {
+    vpcs: HashMap<u32, Vpc>,
```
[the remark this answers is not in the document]
qmonnet: That's a good question. I think Sergey preferred to use names, I thought using just the VNIs might be more efficient than hashing strings every time. I feel like VPCs should have some sort of ID, and VNIs play this role nicely for now... I don't mind reverting to names if this is what we prefer, this is a small internal change that is easy to do.

See also the other reply on objects cross-referencing - I expect some movement on the way objects refer to one another anyway.

qmonnet:
Thanks a lot for the review! It's probably not clear enough from the PR description, but I expect all code related to the creation of VPCs, PIFs, PeeringPolicies to move out of the NAT submodule and to be shared more broadly with other components. This is why I didn't spend time ensuring unicity of names, or IP prefixes, etc.: I expect this code to move to some more global configuration step, and change a lot. Just like I didn't try to get the VNI lookup working, because I figured I'd get it as part of the metadata the `Nat` object would receive. But I consider implementing a shared context is beyond the scope of this Pull Request.

What I think matters the most in the PR is:

- The overall structure of the different objects, as used by the NAT. For example, validating that the `PeeringPolicy` refers to two VPCs (whether it refers to them through a direct Rust reference, or by name, or by VNI, is an implementation detail in the scope of this PR, and can certainly be adjusted when we move this to some shared config component).
- The overall logics for stateless NAT: what prefixes we look at for source NAT, destination NAT, how the mapping works in principle (then again, the internals of the `IpList` object are easy to modify if required).

Fredi, does this address your concerns?

dataplane/src/nat/mod.rs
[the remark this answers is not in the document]
qmonnet: I can use names as keys in that table, I've looked into it and it's a minimal change. But in the scope of this PR, it doesn't really make sense: The table is only used to look up for the source VPC for a packet, and this look up necessarily involves the VNI, as we don't know the name of the VPC from the packet.

So changing to name in that table requires either iterating on the hash table values to search for the VPC with the right VNI, or adding a separate table with a VNI -> VPC_name mapping to the global context.

My proposal would be to keep the VNI here for now, and adjust once we can pull the API objects from a global component external to the NAT module.

qmonnet:
> The self-hosted runner: lab-z6r4v-runner-797mc lost communication with the server.

PR #329, opened 2025-03-04
chore(dpdk): Fix clippy warning in docs

mvachhar:


qmonnet (changes requested):
Thank you! You beat me to it. And my bad, I forgot to run Clippy yesterday when fixing the rustdoc.

Given that we're clean from Clippy warnings at the moment, could you please add `#[deny(clippy::all)]` at the root of the crate to avoid reintroducing such warnings in the future?

qmonnet (approved):
Manish, I added the Clippy-deny as a separate commit - Please take a look and merge it if you're fine with it.

PR #330, opened 2025-03-04
feat: introduce ID crate

daniel-noland:
See docs in crate for details

I hope to make this useful in other crates shortly.


qmonnet (commented):
The code looks OK to me, but could you provide more details on the use case you have in mind, please?

mvachhar:
> The code looks OK to me, but could you provide more details on the use case you have in mind, please?

You can take a look at #320 where I will move from my internal version of this to a similar construct.  But perhaps a better description in the PR is warranted.

mvachhar (approved):
Let's give a better PR message for motivation and then I am good to merge this.  I'll probably add some stuff to this crate for mapping IDs to human readable strings.

qmonnet (approved):
OK sounds good, thanks a lot!

mvachhar:
> The motivation is to distinguish different types of IDs which we expect to use in configuration.
> 
> I am confident we will end up with multiple tables of configuration data, and keeping the IDs of VPCs compile time distinct from the IDs of nat rules can make things easier and less error-prone. A function which returns or accepts `Id<Vpc>` is clearer than one which just returns `Id`. `Id` of what?

In the future, we can put the motivation in the commit log.  I don't think it is worth holding up this PR just for that though.  Something to keep in mind for the future.

PR #335, opened 2025-03-07
doc(nat): Add some docs (rustdoc, comments) to the nat module

qmonnet:
Add rustdoc for most of nat's submodules, but not for the fabric.rs submodule, which is expected to change or disappear soon.

Note: The dataplane package is currently a binary, not a library, and its modules are private. As such, the doctest examples are not run on "cargo test", and the docs are not "expected" by rustdoc (so adding directive `#![deny(missing_docs)]` would have no effect).

### DIFF
--- a/dataplane/src/nat/iplist.rs
+++ b/dataplane/src/nat/iplist.rs
@@ -1,26 +1,91 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
+//! A data structure interface presenting a list of IP prefixes as a flat list
+//! of IP addresses.
+
 use iptrie::IpPrefix;
 use routing::prefix::Prefix;
 use std::fmt::Debug;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
+/// Type for an [`IpList`] object.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


dataplane/src/nat/iplist.rs
```diff
@@ -1,26 +1,91 @@
...
+/// |  10.1.255.255 |                              | 192.168.1.255 |
+/// |      10.0.5.0 |                              |     10.10.0.0 |
+/// |         ...   |                              |         ...   |
+/// |    10.0.5.255 |                              | 10.10.255.255 |
+/// |      10.2.0.0 |                              |      10.8.0.0 |
+/// |      10.2.0.1 |  10.2.0.1 maps to 10.8.1.0   |      10.8.0.1 |
+/// |      10.2.0.2 | ---------------------------> |      10.8.1.0 |
+/// |      10.2.0.3 |                              |      10.8.1.1 |
+/// |               |                              |               |
+/// +---------------+                              +---------------+
```
qmonnet, line 82: Interestingly, writing this piece of the doc made me realise that the mapping model is broken. By ignoring the subnet and broadcast addresses when establishing the mapping, we may end up with nonsensical translations such as this example, where we map to a network address.

This requires a fix that will come in its own Pull Request.

mvachhar (approved):

PR #336, opened 2025-03-07
ci: test

qmonnet:


.github/workflows/bump.yml
```diff
@@ -86,7 +86,6 @@ jobs:
         run: |
           cargo binstall -y cargo-edit # required to make `cargo upgrade` edit the Cargo.toml file
           cargo binstall -y cargo-deny
-          cargo binstall -y cargo-upgrades
```
qmonnet, line 89: @daniel-noland `cargo upgrade` still works (or at least doesn't throw an error) even if we don't install `cargo-upgrades`; I think `cargo-upgrades` would need to be called with `cargo upgrades` (note the `s`) anyway.

So I've got two questions:

1. Can we get rid of `cargo-upgrades`, that is seemingly unused in this workflow?
2. Where does `cargo upgrade` come from?? :thinking:

qmonnet:
See #338

PR #338, opened 2025-03-07
CI: In bump.yml, remove unused dep, rework cargo upgrades step to split PRs into individual upgrade commits

qmonnet:
- **ci(bump): Remove installation of cargo-upgrades**
- **ci(bump): Rework cargo upgrades**

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -103,20 +103,58 @@ jobs:
       - name: "cargo upgrade"
         id: upgrade
         run: |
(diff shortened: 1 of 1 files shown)


qmonnet:
Apart from some mysterious churn on the `windows-sys` version between commits, the `cargo upgrade` step seems to work well. This is ready for a try, as far as I can tell

daniel-noland (approved):
This is objectively better than what I cooked up so yeah.  LGTM

PR #340, opened 2025-03-09
Simple packet hash for ip traffic

Fredi-raspall:
@daniel-noland I was going to move Packet to a separate crate (having packet_metadata and this new packet_hash as submodules), but did not do that since I remembered you saying that you had that change in a separate branch.

### DIFF
--- a/dataplane/src/packet_hash.rs
+++ b/dataplane/src/packet_hash.rs
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+//
+//! Module to compute packet hashes
+
+use super::Packet;
+use ahash::AHasher;
+use net::buffer::PacketBufferMut;
+use net::headers::{Net, Transport, TryHeaders, TryIp, TryTransport};
+use std::hash::{Hash, Hasher};
+
+impl<Buf: PacketBufferMut> Packet<Buf> {
+    #[allow(unused)]
+    /// Computes a hash over a `Packet` object if it contains an ipv4 or ipv6 packet,
+    /// using invariant fields of the ip header and common transport headers,
+    /// if present, using the specified Hasher.
+    pub fn hash_ip<H: Hasher>(&self, state: &mut H) {
+        if let Some(ip) = self.headers().try_ip() {
+            match ip {
+                Net::Ipv4(ipv4) => {
+                    ipv4.source().hash(state);
+                    ipv4.destination().hash(state);
+                    ipv4.protocol().hash(state);
+                }
+                Net::Ipv6(ipv6) => {
+                    ipv6.source().hash(state);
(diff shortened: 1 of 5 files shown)


dataplane/src/packet_hash.rs
```diff
@@ -0,0 +1,59 @@
...
+use super::Packet;
+use net::buffer::PacketBufferMut;
+use net::headers::{TryHeaders, TryTransport};
+
+use ahash::AHasher;
+use net::headers::Net;
+use net::headers::Transport;
+use net::headers::TryIp;
+use std::hash::Hash;
+use std::hash::Hasher;
```
qmonnet, line 15: Formatting nit:

```suggestion
use ahash::AHasher;
use net::buffer::PacketBufferMut;
use net::headers::{Net, Transport, TryHeaders, TryIp, TryTransport};
use std::hash::{Hash, Hasher};
```

(Not tested.)

qmonnet (commented):
Looks good, thanks! Some questions/observations:

- Why ahash?

- Maybe add some unit test, for example simply trying to hash a packet and check that we obtain a hash within the bounds?

Fredi-raspall:
> Looks good, thanks! Some questions/observations:
> 
> * Why ahash?

It should be in principle significantly faster than the default hasher.

Fredi-raspall:
> * Maybe add some unit test, for example simply trying to hash a packet and check that we obtain a hash within the bounds?

That can be done, yes. However, we will need to build a number of packets. Isn't https://github.com/githedgehog/dataplane/pull/340/files#diff-b9135b759a2b2802c52b1e1943f1fea907c86f6e8c7ccf28586a4010d41cf959R57 
just enough for that ?

qmonnet:
Can't you simply reuse `build_test_ipv4_packet()` like in other dataplane tests?
It would be nice to have some test if it's straightforward to do, otherwise if generating the required boilerplate is much work then yes, I can live with your current code :slightly_smiling_face:

Fredi-raspall:
> Can't you simply reuse `build_test_ipv4_packet()` like in other dataplane tests? It would be nice to have some test if it's straightforward to do, otherwise if generating the required boilerplate is too much work then yes, I can live with your current code 🙂

 `build_test_ipv4_packet()` has the addresses hardcoded and won't be of much help here. I believe I have a variant of it that allows setting distinct addresses and the like in a branch that has not yet been merged. I can add such tests when those test utils are merged if that's fine with you ?

qmonnet:
Sure, fine by me, thank you!

qmonnet (approved):

dataplane/src/packet_hash.rs
Fredi-raspall: @qmonnet I have manually cleaned up the imports since you said the above wasn't tested.
Thanks!

PR #342, opened 2025-03-10
ci: Check Git repository status to avoid omitting Cargo.lock updates

qmonnet:
To avoid contributors forgetting to commit the Cargo.lock updates, let's add a step to the sterile.yml workflow, to check whether building or testing the changes resulted in an updated Cargo.lock file. To do that, we check the result value from `git diff --quiet`.


qmonnet:
Seems to work as intended. With the first commit in the PR only, I get:

image

Now adding a second commit with the Cargo.lock updates.

qmonnet:
`dev.yml` seems to fail on the per-commit build, precisely because we're missing the update in commit 2 of the Pull Request.

mvachhar (approved):

PR #345, opened 2025-03-11
Packet Hash test

Fredi-raspall:


qmonnet:
Converting to draft, we'll likely go with #347 instead

Fredi-raspall:
@mvachhar I am going to close this in favor of https://github.com/githedgehog/dataplane/pull/345 which contains roughly the same, but with the changes on top of the necessary move of Packet out of the dataplane crate.

qmonnet:
Fredi, can we delete this branch?

PR #346, opened 2025-03-12
fix: Move Packet to net

daniel-noland:
`Packet` (et. al.) need to live in net to avoid circular dependencies with, for example, the routing crate.

This logic was created in dataplane but never truly belonged there in the first place.

### DIFF
--- a/net/src/packet/mod.rs
+++ b/net/src/packet/mod.rs
@@ -1,14 +1,25 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
-use crate::packet_meta::{DoneReason, PacketMeta};
-use net::buffer::PacketBufferMut;
-use net::eth::EthError;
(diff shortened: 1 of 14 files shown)


Fredi-raspall (approved):
Looks good to me @daniel-noland !

net/src/packet/mod.rs
```diff
@@ -50,7 +66,12 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
         self.mbuf.take()
     }
 
-    pub(crate) fn reserialize(mut self) -> Buf {
+    /// Reserialize the packet into a buffer (consuming self).
+    ///
+    /// # Panics
+    ///
+    /// This method should never panic baring programmer error.
```
qmonnet, line 73: > baring programmer error

This reads like the programmer using this function could make a mistake and trigger a `panic`, but I think you need it shouldn't panic unless we did a mistake in the function/class? So maybe “baring implementation error”? [Not blocking]

qmonnet (approved):
Looks good, and makes sense indeed. Thanks!

PR #347, opened 2025-03-12
Pr/fredi/hash test rebase

Fredi-raspall:
This is https://github.com/githedgehog/dataplane/pull/345 rebased on https://github.com/githedgehog/dataplane/pull/346


dataplane/src/nat/mod.rs
```diff
@@ -393,10 +393,11 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for Nat {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::pipeline::test_utils::build_test_ipv4_packet;
     use iptrie::Ipv4Prefix;
     use net::buffer::TestBuffer;
     use net::headers::TryIpv4;
+    #[cfg(test)]
+    use net::packet::test_utils::build_test_ipv4_packet;
```
qmonnet, line 400: This is already under `#[cfg(test)]`, I don't think we need to add a second `#[cfg(test)]`?

dataplane/src/pipeline/test_utils.rs
```diff
@@ -6,16 +6,6 @@
...
 use crate::pipeline::{DynNetworkFunction, nf_dyn};
 use net::buffer::TestBuffer;
-use net::eth::Eth;
-use net::eth::ethtype::EthType;
-use net::eth::mac::{DestinationMac, Mac, SourceMac};
-use net::headers::{Headers, Net};
-use net::ipv4::Ipv4;
-use net::ipv4::addr::UnicastIpv4Addr;
-use net::packet::{InvalidPacket, Packet};
-use net::parse::DeParse;
```
qmonnet, line 16: And that's how you know that `net` was the right place for that function :slightly_smiling_face:

net/src/packet/test_utils.rs
```diff
@@ -0,0 +1,117 @@
...
+use std::default::Default;
+use std::net::Ipv4Addr;
+use std::str::FromStr;
+
+/// Builds a test packet with the given TTL value.
+///
+/// The packet is an IPv4 packet with a source and destination IP address of 1.2.3.4.
+/// The Ethernet source and destination MAC addresses are 0x02:00:00:00:00:01 and 0x02:00:00:00:00:02
+/// respectively.
+///
```
qmonnet:
```suggestion
```

net/src/packet/test_utils.rs
```diff
@@ -0,0 +1,117 @@
...
+    let mut udp = Udp::empty();
+    udp.set_source(UdpPort::new_checked(sport).expect("Bad src port"));
+    udp.set_destination(UdpPort::new_checked(dport).expect("Bad dst port"));
+    headers.transport = Some(Transport::Udp(udp));
+
+    let mut buffer: TestBuffer = TestBuffer::new();
+    headers.deparse(buffer.as_mut()).unwrap();
+
+    Packet::new(buffer).unwrap()
+}
```
qmonnet, line 118: How about something like:

```suggestion
#[allow(unsafe_code)]
pub fn build_test_udp_ipv4_packet(
    src_ip: &str,
    dst_ip: &str,
    sport: u16,
    dport: u16,
) -> Packet<TestBuffer> {
    build_test_udp_ipv4_frame(
        Mac([0x2, 0, 0, 0, 0, 1]),
        Mac([0x2, 0, 0, 0, 0, 1]),
        src_ip,
        dst_ip,
        sport,
        dport)
}
```

net/src/packet/hash.rs
```diff
@@ -52,3 +52,120 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
...
+
+    // compute ip hash using ahash hasher
+    fn hash_ip_packet<Buf: PacketBufferMut>(packet: &Packet<Buf>) -> u64 {
+        let mut hasher = AHasher::default();
+        packet.hash_ip(&mut hasher);
+        hasher.finish()
+    }
+    // Builds a vector of packets.
+    // Note: If this function is changed, the fingerprint file may
+    // need to be updated.
```
qmonnet, line 77:
```suggestion
    // Note: If this function is changed, the fingerprint file may
    // need to be updated.
    //
    // See instructions in the comment for test_ahash_detect_changes().

**Review by qmonnet — CHANGES_REQUESTED** [review]
Some small suggestions, but looks good overall, thanks!

File: net/src/packet/test_utils.rs (hunk 3 above)
Comment (qmonnet) [reply net/src/packet/test_utils.rs]: Actually I think you fixed that one in the last commit

File: dataplane/src/nat/mod.rs (hunk 1 above)
Comment (Fredi-raspall) [reply dataplane/src/nat/mod.rs:400]: true, it's not needed. Will remove

File: dataplane/src/pipeline/pipeline.rs [hunk 6]
```diff
@@ -188,8 +188,10 @@ mod test {
 
     use crate::pipeline::dyn_nf::DynNetworkFunctionImpl;
     use crate::pipeline::sample_nfs::DecrementTtl;
-    use crate::pipeline::test_utils::{DynStageGenerator, build_test_ipv4_packet};
+    use crate::pipeline::test_utils::DynStageGenerator;
     use crate::pipeline::{DynNetworkFunction, DynPipeline, NetworkFunction, StageId};
+    #[cfg(test)]
```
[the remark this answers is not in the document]
Fredi-raspall: It's not needed. It was a leftover

qmonnet (approved):
Looks good, thanks!

qmonnet:
We enforce stricter Clippy and rustdoc checks on the net module it seems, so CI complains. You can address these or temporary lift the restrictions in the relevant submodules.

Fredi-raspall:
> We enforce stricter Clippy and rustdoc checks on the net module it seems, so CI complains. You can address these or temporary lift the restrictions in the relevant submodules.

I think these are nonsense tbh.

qmonnet:
Thanks for fixing it anyway. This seems ready to go!

daniel-noland:
> I think these are nonsense tbh.

I'm really confused by that.

The dataplane isn't allowed to panic so we should only call unwrap and friends if we think that the error is unreachable.

What is the alternative?

mvachhar (approved):
The consistent hashing thing is a bigger lift than I thought.  I'm going to approve this for now, then we can address consistent hashing which will involve reworking a bunch of this code.

PR #350, opened 2025-03-13
Fix address mapping for stateless NAT

qmonnet:
The previous design for mapping addresses for stateless NAT had an issue. By using a representation of a flattened list of all endpoints/public-IP prefixes, and only ensuring that the origin and target address lists are of the same size, we risk mapping regular unicast addresses to potential network or broadcast addresses of a subnet, if the subnet structure is different in the two sets.

This PR addreses the issue by making sure we simply translate from one prefix to a prefix of the same size. It also extend the revamped IpList struct to support exclusion prefixes as defined in the current API proposal for the Gateway.

### DIFF
--- a/dataplane/src/nat/mod.rs
+++ b/dataplane/src/nat/mod.rs
@@ -41,7 +41,7 @@ mod iplist;
 mod prefixtrie;
 
 use crate::nat::fabric::{PeeringPolicy, Pif, Vpc};
-use crate::nat::iplist::{IpList, IpListType};
+use crate::nat::iplist::IpList;
 use crate::nat::prefixtrie::PrefixTrie;
 use crate::pipeline::NetworkFunction;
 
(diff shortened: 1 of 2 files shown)


dataplane/src/nat/mod.rs
```diff
@@ -245,9 +245,9 @@ impl Nat {
     /// From the destination IP address contained in `net`, finds the
     /// destination PIF for the packet.
     #[tracing::instrument(level = "trace")]
-    fn find_dst_pif(&self, net: &Net) -> Option<&Pif> {
+    fn find_dst_pif(&self, dst_ip: &IpAddr) -> Option<&Pif> {
```
Fredi-raspall, line 229: I'm a bit confused now. Are we going to keep the pif abstractions? cc @mvachhar

qmonnet: Not that I know of. Note that `find_dst_pif()` is not added in this PR. This code was written at a time when we considered keeping the PIF abstraction. We'll need to adjust it when we've finalised the data model, but this PR does not attempt to address that.

mvachhar:
Given the API changes, is this PR still relevant?

qmonnet:
Not as much as I initially hoped. I'd like to merge it anyway, this is still a small step in the right direction when compared to the current version of the code.

mvachhar (approved):
I do wonder if a library like [this](https://docs.rs/cidr/latest/cidr/struct.Ipv4Cidr.html) or something similar would help us encapsulate the prefix and containment logic here instead of writing it ourselves, though that may not interact well with the trie.  Just a thought.

qmonnet:
I started with that library, but I don't think it'll work with the LPM map we're using. I switched to `iptrie` to align with the routing code. `iptrie` offers some encapsulation for the prefixes, in a way similar to `cidr`, but it doesn't offer all methods that `cidr` has. It does provide some logics for containments/coverage, we just reuse it in our `Prefix` abstraction that wraps around it.

PR #352, opened 2025-03-17
ci: Disable stable toolchain for dev workflow

qmonnet:
We don't use the "stable" toolchain these days, given that we require Rust features that are in 1.86 (the current beta). Running the stable workflow in CI just burns resources and generates some noise in the discussion thread for the Pull Requests. Disable it.

This commit should be revert right before we revert commit 643de7afc7fc ("fix(justfile): Use beta release of rust, not stable").

Note: 1.86 is due to become stable on 3rd April (https://releases.rs/docs/1.86.0/).

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -75,9 +75,12 @@ jobs:
       fail-fast: false
       matrix:
         rust:
-          - # failures on stable block release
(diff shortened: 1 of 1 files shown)


daniel-noland (approved):
:+1:

PR #353, opened 2025-03-17
build: Add "just clippy" target

qmonnet:
Add a `just clippy` target to the justfile to help run Clippy. Running Clippy on the codebase is easy: `cargo clippy` should do. However, this may not run Clippy the way one should expect:

- If run without `just`, this may not pass all required flags to cargo, and may result Clippy running with different options. One consequence would be some undesired updates to the Cargo.lock file, or some Clippy reports that don't show as expected.

- By default, `cargo clippy` or even `just cargo clippy` will not run Clippy for all features, and in particular, they will not run on unit tests. This may hide Clippy reports.

In CI, we already make sure we run Clippy on tests since commit 550c75136cf9 ("ci(clippy): Run Clippy on all targets"). This commit adds a `clippy` target to the justfile to help users doing the same locally, just like we do in CI, without having to remember to pass all the necessary arguments to Clippy.

### DIFF
--- a/justfile
+++ b/justfile
@@ -393,6 +393,12 @@ mdbook *args="build":
       {{ _doc_env_container }} \
       {{ args }}
 
+# Serve rustdoc output locally (using port 8000)
+[script]
+rustdoc-serve:
+    echo "Launching web server, hit Ctrl-C to stop."
+    python -m http.server -d "target/{{ target }}/doc"
+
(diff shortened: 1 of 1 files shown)


qmonnet:
Add a second commit to add a just target to serve the generated rustdoc locally.

Fredi-raspall (approved):
Looks good to me @qmonnet !

PR #354, opened 2025-03-17
chore: Add missing license info in Cargo.toml files

qmonnet:
To avoid ambiguity in terms of licensing, explicitly provide a license parameters in the Cargo.toml of all packages in the repository. This should also help FOSSA, which currently struggles to determine the license for these crates.

The license is set to Apache-2.0, as for the packages that already stated a license.

Relates to #332


daniel-noland:
This is very strange.  I swear I already did this but I clearly did not.

Good catch.  Approved

mvachhar (approved):
LGTM

PR #355, opened 2025-03-17
Packet utils

Fredi-raspall:
Adds several getters for packet fields frequently retrieved, using in the results generalized types like std::net::IpAddr or Mac, (instead of DestinationMac or SourceMac) so that consumers need not depend on or distinguish those types. The methods are meant for code reuse, readability and ergonomics.

### DIFF
--- a/net/src/packet/utils.rs
+++ b/net/src/packet/utils.rs
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Packet higher-level methods to allow for code reuse
+
+use std::net::IpAddr;
+
+use crate::eth::Eth;
+use crate::eth::ethtype::EthType;
+use crate::eth::mac::Mac;
+use crate::headers::Net::{Ipv4, Ipv6};
+use crate::headers::{TryEth, TryIp};
+use crate::ip::NextHeader;
+use crate::packet::Packet;
+use crate::packet::PacketBufferMut;
+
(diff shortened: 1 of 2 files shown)


net/src/packet/utils.rs
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+//
+//! Packet higher-level methods to allow for code reuse
```
qmonnet, line 4:
```suggestion
// SPDX-License-Identifier: Apache-2.0
// Copyright Open Network Fabric Authors

//! Packet higher-level methods to allow for code reuse
```

net/src/packet/utils.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    /// Get the source mac address of a [`Packet`]
+    /// Returns None if the packet does not have Ethernet header
+    pub fn eth_source(&self) -> Option<Mac> {
+        self.try_eth().map(|eth| eth.source().inner())
+    }
+
+    /// Get the source mac address of a [`Packet`]
+    /// Returns None if the packet does not have Ethernet header
+    pub fn eth_type(&self) -> Option<EthType> {
```
qmonnet, line 32: The comment is a copy-paste from the previous method, it needs to be updated for `eth_type()`.

qmonnet (changes requested):
One comment to update, looks good otherwise. Thanks!

qmonnet (approved):
Looks good, thank you! This is good to merge as far as I'm concerned.

PR #360, opened 2025-03-20
Move the pipeline code to separate crate

Fredi-raspall:
This is needed to avoid circular dependencies.

### DIFF
--- a/pipeline/src/lib.rs
+++ b/pipeline/src/lib.rs
@@ -2,10 +2,12 @@
 // Copyright Open Network Fabric Authors
 
 #![allow(rustdoc::private_doc_tests)]
+#![deny(clippy::all)]
+#![deny(clippy::pedantic)]
 
 //! # Pipeline Building Blocks
 //!
-//! This module provides the building blocks for constructing pipelines of network functions.
+//! This crate provides the building blocks for constructing pipelines of network functions.
 //! There are two main methods provided for linking network functions together in sequence:
 //!
 //! - `StaticChain`: A trait for statically chaining network functions together.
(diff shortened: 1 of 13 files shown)


pipeline/src/lib.rs
```diff
@@ -102,6 +105,7 @@
 mod dyn_nf;
 #[allow(clippy::module_inception)]
```
mvachhar, line 106: Is this still required since we moved the crate?  It might be, not sure.

pipeline/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 
 //! # Pipeline Building Blocks
```
mvachhar, line 6: I think we need all `#[deny(clippy:all), deny(clippy:pendantic]` since we are no longer inheriting it.

mvachhar (changes requested):

pipeline/src/lib.rs
Fredi-raspall: I don't think it's required. Removing...

Fredi-raspall:
I am a bit confused about why the pipeline is now failing, after having fixed the doc tests locally.
```
just cargo test --package=pipeline --doc
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.04s
   Doc-tests pipeline

running 3 tests
test pipeline/src/lib.rs - (line 26) ... ok
test pipeline/src/lib.rs - (line 78) ... ok
test pipeline/src/lib.rs - (line 57) ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

@daniel-noland are we running the doc test with release builds ?

qmonnet:
We do run with `release` in CI, in the `sterile.yml` workflow:

```
just debug_justfile="" rust="beta" profile=release target=x86_64-unknown-linux-gnu \
    sterile cargo doc
```

I believe the reasoning is to run the doctests along with the regular tests, with the same profile?

Anyway, it doesn't work at the moment, this is a [known issue](https://github.com/githedgehog/dataplane/commit/5ed94136918f78e0ce402458667a678e3a134dd1). I'd recommend disabling doctests like we do for other crates for now; we really need to fix these at some point.

qmonnet:
> I am asking since if I do build with --release, then it also fails locally, but I guess that's not something we plan to do for doc tests, right?

Thinking again about this, I think you're right, it makes little sense generating the docs with the `release` profile especially if doctests are broken. I'm trying to remove that step in https://github.com/githedgehog/dataplane/pull/364.

mvachhar (approved):
The changes look good, but the doctest failure is mysterious.  I believe it is triggered by these changes but not caused by them so I'll approve this for merge and we can merge once we figure out what is causing the doctest failure on `--release`.

qmonnet (approved):

PR #361, opened 2025-03-21
chore(dataplane): fix command line

Fredi-raspall:
The command line was throwing an error when requesting --help Enabling the default features at the top level Cargo.toml seems to fix the problem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ bindgen = { version = "0.71.1", default-features = false, features = [] }
 bolero = { version = "0.13.0", default-features = false, features = [] }
 bytes = { version = "1.10.1", default-features = false, features = [] }
 # std feature currently globally required for clap
-clap = { version = "4.5.32", default-features = false, features = ["std"] }
+clap = { version = "4.5.32", default-features = true, features = [] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -42,7 +42,7 @@ bindgen = { version = "0.71.1", default-features = false, features = [] }
 bolero = { version = "0.13.0", default-features = false, features = [] }
 bytes = { version = "1.10.1", default-features = false, features = [] }
 # std feature currently globally required for clap
-clap = { version = "4.5.32", default-features = false, features = ["std"] }
+clap = { version = "4.5.32", default-features = true, features = [] }
```
qmonnet, line 45: Do you need to set `default-features = true` here if you already specify the required features in `dataplane/Cargo.toml`?

qmonnet (approved):
One question below, but either way this looks good, thank you!

Cargo.toml
Fredi-raspall: Apparently yes. I tried setting the features only in the dataplane crate and it did not work.

qmonnet: Ok thanks

PR #363, opened 2025-03-21
Kernel Driver

Fredi-raspall:
Rides on https://github.com/githedgehog/dataplane/pull/369

* Requires privileges
* If you wan to give it a try:
```
sudo ./target/x86_64-unknown-linux-gnu/debug/dataplane --driver kernel --interface any
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ bytes = { version = "1.10.1", default-features = false, features = [] }
 clap = { version = "4.5.32", default-features = true, features = [] }
 ctrlc = { version = "3.4.5", default-features = false, features = [] }
 derive_builder = { version = "0.20.2", default-features = false, features = ["default"] }
+default-net = { version = "0.22.0", default-features = false, features = [] }
 doxygen-rs = { version = "0.4.2", default-features = false, features = [] }
 dyn-iter = { version = "1.0.1", default-features = false, features = [] }
 etherparse = { version = "0.17.0", default-features = false, features = [] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ dpdk-sysroot-helper = { path = "./dpdk-sysroot-helper" }
 dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", version = "1.0.0" }
 errno = { path = "./errno", package = "dataplane-errno" }
 id = { path = "./id", package = "dataplane-id" }
-net = { path = "./net" }
+net = { path = "./net", features = ["test_buffer"]}
 pipeline = { path = "./pipeline" }
 routing = { path = "./routing" }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ pipeline = { path = "./pipeline" }
 routing = { path = "./routing" }
 
 # External
+afpacket = { version = "0.2.3", default-features = false, features = [] } 
 ahash = { version = "0.8.11", default-features = false, features = [] }
 arbitrary = { version = "1.4.1", default-features = false, features = [] }
 arc-swap = { version = "1.7.1" , default-features = false, features = [] }
--- a/dataplane/Cargo.toml
+++ b/dataplane/Cargo.toml
@@ -6,6 +6,7 @@ publish = false
 license = "Apache-2.0"
 
 [dependencies]
+afpacket = { workspace = true }
 clap = { workspace = true, features = ["std", "derive", "usage"] }
 ctrlc = { workspace = true, features = ["termination"] }
 dpdk = { workspace = true }
--- a/dataplane/Cargo.toml
+++ b/dataplane/Cargo.toml
@@ -13,7 +13,7 @@ id = { workspace = true }
 iptrie = { workspace = true }
 dyn-iter = { workspace = true }
 mio = { workspace = true, features = ["os-ext", "net"] }
-net = { workspace = true, features = ["serde"] }
+net = { workspace = true, features = ["serde", "test_buffer"] }
 ordermap = { workspace = true, features = ["std"] }
 pipeline = { workspace = true }
 routing = { workspace = true }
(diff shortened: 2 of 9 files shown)


mvachhar:
I think there are 2 possible approaches here.  Each worker gets its own pipeline but we explicitly pass in the shared state when we construct the pipeline stage, in which case the shared state object has to manage concurrent access, or we allow the pipeline to be shared and every pipeline stage has to deal with concurrency, even if private data per thread would be enough, the stage would have to allocate that private state. 

I have a preference for the first approach, and if a stage is mostly shared state (e.g., routing) a thin wrapper that just calles the process packets method on the shared router object is probably fine.  What do others think?

Fredi-raspall:
This PR still lacks a commit to be able to pass a pipeline for the DPDK driver.

deny.toml
```diff
@@ -46,6 +46,7 @@ allow = [
   "LGPL-3.0",
   "LPL-1.02",
   "MIT",
+  "MirOS",
```
qmonnet, line 49: Ack. I'm not a lawyer, but it seems safe to use.

dataplane/src/drivers/dpdk.rs
```diff
@@ -20,8 +20,10 @@ use net::packet::Packet;
 use pipeline::sample_nfs::Passthrough;
 use pipeline::{self, DynPipeline, NetworkFunction};
 
+/*
 #[global_allocator]
 static GLOBAL_ALLOCATOR: RteAllocator = RteAllocator::new_uninitialized();
+ */
```
qmonnet: Would be nice to provide some context on why we temporary disable this (instead of just removing it).

dataplane/Cargo.toml
```diff
@@ -15,7 +15,7 @@ iptrie = { workspace = true }
 default-net = { workspace = true }
 dyn-iter = { workspace = true }
 mio = { workspace = true, features = ["os-ext", "net"] }
-net = { workspace = true, features = ["serde"] }
+net = { workspace = true, features = ["serde", "test_buffer"] }
```
qmonnet, line 18: What do you need this for, given that this comes as the last commit in the PR? :thinking:

Cargo.toml
```diff
@@ -48,6 +49,7 @@ bytes = { version = "1.10.1", default-features = false, features = [] }
 # std feature currently globally required for clap
 clap = { version = "4.5.32", default-features = true, features = [] }
 ctrlc = { version = "3.4.5", default-features = false, features = [] }
+default-net = { version = "0.22.0", default-features = false, features = [] }
```
qmonnet, line 52: This is the crate causing the issues in CI. Not sure how to best address them yet.

qmonnet (approved):
Looks good to me, nice work! I've got a couple questions below, but nothing that's blocking.

Manish, regarding your comment: No strong opinion, I can see the simplicity behind going for one pipeline per worker when compared with doing concurrency management for every state. Routing or NAT tables will have to be shared though, right?

> I have a preference for the first approach, and if a stage is mostly shared state (e.g., routing) a thin wrapper that just calles the process packets method on the shared router object is probably fine.

I'm not sure I understand what exactly what you mean for this wrapper?

Cargo.toml
qmonnet: And I don't understand why this passes with the `beta` toolchain but fails on `nightly`, by the way

Cargo.toml
```diff
@@ -33,11 +33,12 @@ dpdk-sysroot-helper = { path = "./dpdk-sysroot-helper" }
 dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", version = "1.0.0" }
 errno = { path = "./errno", package = "dataplane-errno" }
 id = { path = "./id", package = "dataplane-id" }
-net = { path = "./net" }
+net = { path = "./net", features = ["test_buffer"]}
```
qmonnet, line 36: I don't know if this one is required if you use the feature in `dataplane/Cargo.toml` already, although I already asked you about something similar in a previous PR and you said it was necessary so maybe it is here as well. Just give a quick try to see if it's removable, if you haven't already.

Cargo.toml
qmonnet: It doesn't seem we can downgrade `thiserror` for the other crates using it. So if there's no alternative to `default-net` to find the interfaces, we should probably mark `thiserror` as an exception. @daniel-noland another suggestion?

dataplane/src/drivers/dpdk.rs
qmonnet: Maybe just add a comment about that in the code?

Fredi-raspall: I have commented the issue in the commit.

Cargo.toml
qmonnet: The difference seems to come from the output from `cargo metadata --format-version 1 --manifest-path Cargo.toml`. With `nightly`, this command prints info about the older version of `thiserror`, whereas `beta`'s `cargo` does not.

qmonnet: In the meantime:

```diff
diff --git i/deny.toml w/deny.toml
index 4c1e6aceed02..9f3bd2444279 100644
--- i/deny.toml
+++ w/deny.toml
@@ -11,7 +11,8 @@ no-default-features = false
 feature-depth = 1
 
 [advisories]
-ignore = []
+# Next release of netlink-packet-utils will drop "paste" and fix this advisory
+ignore = ["RUSTSEC-2024-0436"]
 [licenses]
 unused-allowed-license = "allow"
 allow = [
@@ -200,6 +201,14 @@ reason = "needed for test tracing.  Remove this exception when https://github.co
 name = "regex-syntax"
 reason = "needed for test tracing.  Remove this exception when https://github.com/tokio-rs/tracing/pull/3033 lands"
 
+[[bans.skip]]
+name = "thiserror"
+reason = "haven't found an easy workaround.  Next netlink-packet-utils release should fix this."
+
+[[bans.skip]]
+name = "thiserror-impl"
+reason = "haven't found an easy workaround.  Next netlink-packet-utils release should fix this."
+
 # This section is considered when running `cargo deny check sources`.
 # More documentation about the 'sources' section can be found here:
 # https://embarkstudios.github.io/cargo-deny/checks/sources/cfg.html
```

PR #370, opened 2025-03-24
style: Add clippy.toml to allow unwrap(), expect(), panics in tests

qmonnet:
Instead of manually allowing unwrap(), expect(), and panics in every test module, introduce a clippy.toml file at the root of the repository, using the relevant configuration options to allow these functions inside of test modules.

With this configuration file, most macros are now unnecessary; clean them up.


daniel-noland (approved):
nice :)

PR #372, opened 2025-03-24
justfile: Add compile-env images handling facilities

qmonnet:
- **build(justfile): Warn user when compile-env image is missing/outdated**
- **build(justfile): Add target to prune old compile-env image**

I know the check is barely readable; I decline all responsibility for `just --fmt`'s output.


daniel-noland (approved):
Yeah, as you said, that is a bit ugly but it is an improvement for sure.

PR #374, opened 2025-03-24
Prep for vxlan: move packet to net

daniel-noland:
Packet, as a concept, is not specific to the dataplane.


daniel-noland:
pr created by mistake.  Ignore

PR #376, opened 2025-03-24
Cli binary and protocol with dataplane

Fredi-raspall:
~~NB: rides on https://github.com/githedgehog/dataplane/pull/363~~
NB: rides on https://github.com/githedgehog/dataplane/pull/465

* Adds a cli binary to issue commands to the dataplane process over a unix socket.
* This PR implements the binary and defines the 'protocol' / wire format with the dataplane over the unix socket.
* At the moment, data is returned as strings., but this can be easily changed to json or binary. Deferring this until the complete data model is defined and we can derive those if needed.

**Build**
```
# just cargo build --package=cli --bin cli  --features=cli (old)

# new build in this branch
just cargo build --package=dataplane-cli
```
Once in cli
```
help
```
... shows the available commands. Many are not implemented.

**Give it a try!**
The cli can be started to see the syntax of the commands, even if the dataplane does not handle them.
The command completion requires some re-work.


cli/src/cmdtree.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.children
+            .values_mut()
+            .for_each(|c| c.set_depth(depth + 1));
+    }
+    fn add(&mut self, mut cmd: Node) {
+        if cmd.name.is_empty() {
+            //if a node has no name, skip it and adopt
+            //its children. This is just to ease the
+            //creation of the tree by creating
+            //unnamed commands that will not show up
```
qmonnet, line 86:
```suggestion
            // If a node has no name, skip it and adopt
            // its children. This is just to ease the
            // creation of the tree by creating
            // unnamed commands that will not show up.
```

cli/src/bin/cli.rs
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+    }
+}
+
+fn execute_remote_action(
+    action: CliAction,       /* action to perform */
+    args: &RequestArgs,      /* action arguments */
+    terminal: &mut Terminal, /* this terminal */
+    sock: &UnixDatagram,     /* socket to talk to dataplane */
```
qmonnet: Nit: Rust-style comments for consistency? Here and elsewhere in the file

cli/src/cliproto.rs
```diff
@@ -0,0 +1,172 @@
...
+            result: Err(error),
+        }
+    }
+}
+
+enum_from_primitive! {
+#[allow(unused)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+
+pub enum CliAction {
```
qmonnet, line 126: Missing indent

qmonnet (commented):
I've glanced over the changes, although I haven't looked at each chunk in details. It looks good, no red flags, great work! It's self-contained in its own crate anyway so low risk to break something when we merge it.

Would be great to have more tests, but a number of them will likely require starting the datapath and run some sort of end-to-end test, which we don't really have any infra for, yet, so we can probably keep that for later.

I'll try to build and run this CLI tomorrow to get a better understanding on the features.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 
-members = [
+members = [ "cli",
```
qmonnet, line 3: I note you added `cli` to `members` but not to `default-members`, which means tests won't be exercised on a simple `cargo test` from the root of the repository. Is this Probably something we should change?

cli/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "cli"
+version = "0.1.0"
+edition = "2021"
```
qmonnet, line 4: We've switched to 2024 everywhere else in the repo

```suggestion
edition = "2024"
```

cli/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "cli"
+path = "src/bin/cli.rs"
+required-features = ["cli"]
+
+[dependencies]
+colored = "3.0.0"
+bincode2 = "2.0.1"
+enum_primitive = "0.1.1"
+log = { version = "^0.4.25", features = ["serde"]}
+rustyline = { version = "8.0.0", features = ["with-fuzzy"] }
```
qmonnet, line 18: Let's manage versions from the toplevel Cargo.toml?

qmonnet:
Note: Looking at the `help` output, I'm not entirely clear why some command names appear in blue and bold (in my terminal, at least), and the others with a regular font.

<details><summary>screenshot</summary>

image
</details>

Fredi-raspall:
> Note: Looking at the `help` output, I'm not entirely clear why some command names appear in blue and bold (in my terminal, at least), and the others with a regular font.
> 
> screenshot

I wanted to highlight distinct facts in the output, but I believe I did not finish it. Let me redo.

cli/src/cmdtree.rs
qmonnet: Doesn't have to be doc comments, we do usually leave a space after the `//` marker - at least I can't find any other comment instance without that space.

qmonnet:
> There are some "aliases" or shortcuts, that do not show up in the help. If you want to quit, you can type "quit" or "exit" (these are shown), but you can also use "q".

I found `quit` and `exit`, and it's OK :slightly_smiling_face:. I just find it so much more convenient to exit from a shell session with CTRL+D, as a habit.

sergeymatov (approved):

Cargo.toml
Fredi-raspall: There are not tests at the moment in the CLI

qmonnet:
Note: Rebase required

PR #377, opened 2025-03-25
vxlan encap / decap

daniel-noland:
> [!NOTE]
> ~~This is now on rebased top of #383 which should go in first~~ mission complete

This patch series is aimed at two major goals

1. implement vxlan encap / decap
2. unify the packet and pipeline and net functionality to the point where we can implement the next set of tasks (i.e. interface manager and routing crates)

### DIFF
--- a/net/Cargo.toml
+++ b/net/Cargo.toml
@@ -14,6 +14,7 @@ test_buffer = []
 [dependencies]
 ahash = { workspace = true }
 arbitrary = { workspace = true, features = ["derive"], optional = true }
+arc-swap = { workspace = true }
 arrayvec = { workspace = true }
 bolero = { workspace = true, features = ["alloc", "arbitrary", "std"], optional = true }
 derive_builder = { workspace = true, features = [] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -885,6 +885,7 @@ version = "0.0.1"
 dependencies = [
  "ahash",
  "arbitrary",
+ "arc-swap",
  "arrayvec",
  "bolero",
  "derive_builder",
--- a/dataplane/src/drivers/dpdk.rs
+++ b/dataplane/src/drivers/dpdk.rs
@@ -12,7 +12,7 @@ use dpdk::mem::{Mbuf, Pool, PoolConfig, PoolParams, RteAllocator};
 use dpdk::queue::rx::{RxQueueConfig, RxQueueIndex};
 use dpdk::queue::tx::{TxQueueConfig, TxQueueIndex};
 use dpdk::{dev, eal, socket};
-use tracing::{info, trace, warn};
+use tracing::{error, info, trace, warn};
 
 use crate::CmdArgs;
 use net::buffer::PacketBufferMut;
@@ -113,7 +113,13 @@ fn start_rte_workers(devices: &[Dev], setup_pipeline: &(impl Sync + Fn() -> DynP
                 });
 
                 let pkts_out = pipeline.process(pkts);
-                tx_queue.transmit(pkts_out.map(Packet::reserialize));
+                let buffers = pkts_out.filter_map(|pkt| match pkt.serialize() {
+                    Ok(buf) => Some(buf),
+                    Err(e) => {
+                        error!("{e:?}");
+                        None
+                    }
+                });
             }
         });
     });
--- a/dataplane/src/drivers/kernel.rs
+++ b/dataplane/src/drivers/kernel.rs
@@ -180,8 +180,14 @@ impl DriverKernel {
                         if let Some(oif) = &meta.oif {
                             /* lookup outgoing interface and xmit packet */
                             if let Some(outgoing) = kiftable.get_mut_by_index(oif.get_id()) {
-                                let mut out = pkt.reserialize();
-                                outgoing.sock.write_all(out.as_mut());
+                                match pkt.serialize() {
+                                    Ok(out) => {
+                                        outgoing.sock.write_all(out.as_ref());
+                                    }
+                                    Err(e) => {
+                                        error!("Failure serializing packet: {e:?}");
+                                    }
+                                }
                             } else {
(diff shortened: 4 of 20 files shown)


qmonnet:
I know that the PR is not ready yet, but I took a look and thought I'd comment at least on the decision to move the pipeline to the `net` package. It seems like a good idea to keep it as a separate crate?

How about we make the VXLAN structs only operate on a one packet:

```rust
impl VxlanDecap {
    pub fn run<Buf: PacketBufferMut> (&mut self, mut packet: Packet<Buf>) -> Option<Packet<Buf>> {
        match packet.vxlan_decap() {
            ...
        }
}
```

and

```rust
impl VxlanEncap {
    pub fn run<Buf: PacketBufferMut> (&mut self, packet: Packet<Buf>) -> Option<Packet<Buf>> {
        #[allow(clippy::expect_used)] // clearly impossible to fail here
        let mbuf_len = NonZero::new(...)
        ...
    }
```

Then we could have `pipeline/src/vxlan.rs`:

```rust

use crate::NetworkFunction;
use net::buffer::PacketBufferMut;
use net::packet::Packet;
use net::vxlan::{VxlanDecap, VxlanEncap};

impl<Buf: PacketBufferMut> NetworkFunction<Buf> for VxlanDecap {
    fn process<'a, Input: Iterator<Item = Packet<Buf>> + 'a>(
        &'a mut self,
        input: Input,
        input.filter_map(|packet| self.run(packet))
    }
}

impl<Buf: PacketBufferMut> NetworkFunction<Buf> for VxlanEncap {
    fn process<'a, Input: Iterator<Item = Packet<Buf>> + 'a>(
        &'a mut self,
        input: Input,
        input.filter_map(|packet| self.run(packet))
    }
}
```

I've pushed that to branch https://github.com/githedgehog/dataplane/tree/wip/qmonnet/vxlan-encap-decap, but I haven't managed to clean up the big resulting commit with all `packet` changes yet.

daniel-noland:
@qmonnet 

I'm open to making pipeline its own crate based on conversation with @Fredi-raspall.

Given that vxlan encap / decap does not need to be / shouldn't be a pipeline stage then I think it is fine.

That said, I also think it is fine for pipeline to be a part of net.  I don't see any particular advantage to making it another crate, but maybe I'm missing something

qmonnet:
> I'm open to making pipeline its own crate based on conversation with @Fredi-raspall.
> 
> Given that vxlan encap / decap does not need to be / shouldn't be a pipeline stage then I think it is fine.

OK sounds good, thanks!

> That said, I also think it is fine for pipeline to be a part of net. I don't see any particular advantage to making it another crate, but maybe I'm missing something

I see it as a different logical block, it's probably OK to move it to `net` if it really makes things easier, but otherwise I believe it makes more sense to keep it separate, if only to better understand how the repo is structured.

Fredi-raspall:
> > I'm open to making pipeline its own crate based on conversation with @Fredi-raspall.
> > Given that vxlan encap / decap does not need to be / shouldn't be a pipeline stage then I think it is fine.
> 
> OK sounds good, thanks!
> 
> > That said, I also think it is fine for pipeline to be a part of net. I don't see any particular advantage to making it another crate, but maybe I'm missing something
> 
> I see it as a different logical block, it's probably OK to move it to `net` if it really makes things easier, but otherwise I believe it makes more sense to keep it separate, if only to better understand how the repo is structured.

Isn't the pipeline already a separate crate ?

qmonnet:
Precisely, but this PR is currently moving it into `net`

daniel-noland:
@qmonnet 

This is now rebased to address concerns.

Let me know what ya think

net/Cargo.toml
```diff
@@ -14,11 +14,14 @@ test_buffer = []
 [dependencies]
 ahash = { workspace = true }
 arbitrary = { workspace = true, features = ["derive"], optional = true }
+arc-swap = { workspace = true }
 arrayvec = { workspace = true }
 bolero = { workspace = true, features = ["alloc", "arbitrary", "std"], optional = true }
 derive_builder = { workspace = true, features = [] }
+dyn-iter = { workspace = true }
```
Fredi-raspall, line 21: I think dyn-iter is not  needed ?

Fredi-raspall (approved):
Looks good to me @daniel-noland . Nice work!
I would have opened a smaller PR with the Option Eth header, though.
There's several PRs outstanding and we'll have to adjust them.

Fredi-raspall:
@daniel-noland I have rebased your branch on top of latest main and added a fixup commit to be squashed by you if you think it's fine. I've also pushed your original branch as it was at https://github.com/githedgehog/dataplane/tree/origin/pr/daniel-noland/vxlan-rebase-original

Fredi-raspall:
> Silly question, I see `net/src/vxlan/encap.rs` but no longer the decap equivalent. I take it we only need the `vxlan_decap()` function in [5b4bf02](https://github.com/githedgehog/dataplane/commit/5b4bf021e80cc21ae75113b0023c7fac31557a1d), and no longer the dedicated submodule? (Makes sense I guess since the encap submodule has no code, only errors and struct needed to use the function in mod.rs) Just checking I understand correctly.
> 
> Looks good to me, although CI complains about some commits not building successfully.

No question is silly. The vxlan_decap is in packet/mod.rs

Fredi-raspall:
> Looks good to me @daniel-noland . Nice work! I would have opened a smaller PR with the Option Eth header, though. There's several PRs outstanding and we'll have to adjust them.

Oh, boy, now I see https://github.com/githedgehog/dataplane/pull/383
That was a fast merge! Ignore my comment on the option Eth.

qmonnet:
Thank you Daniel, and Fredi for pushing this forward! Are we good to merge this one?

Fredi-raspall:
> Thank you Daniel, and Fredi for pushing this forward! Are we good to merge this one?

Thank Daniel, I just rebased! 
I think it is good to go.

PR #380, opened 2025-03-26
ci: Fix summary step for dev workflow

qmonnet:
The summary step fails to detect failures at the Clippy or "build-each-commit" steps, because it's not looking for the conditions of these steps at the end of the run. Update the condition to ensure that we look at these results.


qmonnet:
Seems to work as expected, `build-each-commit` makes the `summary` step fail. Now removing test commits...

daniel-noland (approved):
sigh.

LGTM.  Nice work.

PR #383, opened 2025-03-28
Make eth optional

daniel-noland:
It turns out that (upcoming) vxlan encap / decap benefit from making eth an optional field.
Specifically, at one point in the pipeline we simply don't know the ethernet header we want to use yet.

It is also entirely possible that we will use layer 3 tunneling for other protocols in the future so making eth required is not generally wise.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -392,6 +392,41 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


qmonnet (approved):
Looks good to me

PR #391, opened 2025-04-07
Dataplane configuration

Fredi-raspall:


Cargo.toml
```diff
@@ -6,7 +6,7 @@ members = [
 	"dpdk-sys",
 	"dpdk-sysroot-helper",
 	"errno",
-	"id",
+	"id", "mgmt",
```
qmonnet, line 9: Nit: Can we have it on separate lines, please?

mgmt/src/config/routing/bgp.rs
```diff
@@ -0,0 +1,510 @@
...
+    pub strict_capability_match: bool,
+    pub dont_capability_negotiate: bool,
+    pub allow_as_in: bool,
+    pub extended_link_bandwidth: bool,
+    pub next_hop_self: bool,
+    pub remove_private_as: bool,
+    pub rr_client: bool,
+    pub default_originate: bool,
+
+    /* AFs */
```
qmonnet, line 129: Nit: I'd expend to “address families”

`-E2MANYACRONYMS`

mgmt/src/rpc/overlay/vpcpeering.rs
```diff
@@ -115,4 +115,31 @@ impl VpcPeeringTable {
...
+            let name2 = peering.vpc2.as_ref().map(|m| m.name.as_str());
+            if name1.is_none() || name2.is_none() {
+                false
+            } else {
+                name1 == Some(vpc) || name2 == Some(vpc)
+            }
+        })
+    }
+    /// Produce iterator of [`VpcPeering`]s between the two provided vpcs.
+    /// In principle there should be one peering at the most between two vpcs ?
```
qmonnet, line 132: I don't think we can assume that? For any two given VPCs, we could have different peerings for different groups of endpoints from each VPC?

qmonnet (approved):
Looks good to me (with a couple of non-blocking nits), awesome work!

mgmt/src/rpc/overlay/vpcpeering.rs
Fredi-raspall: I was wondering if, _between two specific VPCs_, the API would give us everything in the form of a single peering.

PR #392, opened 2025-04-07
Interface-manager part 2

daniel-noland:
Cleanup finally finished.

Don't merge until #419 lands

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1249,6 +1249,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 
 [[package]]
+name = "rekon"
+version = "0.1.0"
+
+[[package]]
 name = "routing"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -521,7 +521,7 @@ dependencies = [
  "libc",
  "memalloc",
  "netlink-packet-core",
- "netlink-packet-route",
+ "netlink-packet-route 0.17.1",
  "netlink-sys",
  "once_cell",
  "system-configuration",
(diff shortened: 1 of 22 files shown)


irekon/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-//! Network interface management tools for the dataplane
+// Network interface management tools for the dataplane
```
qmonnet: Looked like a good idea to have a module doc line, though :thinking:

irekon/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "irekon"
+version = "0.1.0"
+edition = "2024"
```
qmonnet, line 4: Missing `license`.

(And this commit needs to be squashed with the previous ones, when we introduce `irekon` and when adding the different dependencies.)

vpc-manager/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "dataplane-vpc-manager"
+version = "0.1.0"
+edition = "2024"
```
qmonnet, line 4: Missing `license`.

.cargo/config.toml
```diff
@@ -5,3 +5,6 @@ LIBCLANG_PATH = { value = "compile-env/lib", relative = true, force = true }
 
 [build]
 target = "x86_64-unknown-linux-gnu"
+
+[target.x86_64-unknown-linux-gnu]
+runner = ["/usr/bin/sudo", "-E"]
```
qmonnet, line 10: Why is this necessary?

> This is sadly needed to run (upcoming) interface-manager tests

This lacks precision. What do we need perms for?

Cargo.toml
```diff
@@ -7,19 +7,25 @@ members = [
 	"dpdk-sysroot-helper",
 	"errno",
 	"id",
+	"irekon",
 	"net",
 	"pipeline",
+	"rekon",
 	"routing",
+	"vpc-manager",
```
qmonnet, line 15: We'll have `mgmt` from Fredi's PR soon. This PR brings three more new crates - are we sure we need all of them to be independent crates?

qmonnet (commented):
I do understand that this is still in progress. I flagged a few things I noticed, but I know you probably plan to address a chunk of them already.

Apart from the VPC manager, there are no tests in the PR. Do you plan to add more?

qmonnet:
image

`multi_index_map` currently pulls `proc-macro-error` that gets flagged as unmaintained.

Cargo.lock
```diff
@@ -1257,12 +1481,35 @@ dependencies = [
  "tracing-subscriber",
 ]
 
+[[package]]
+name = "rtnetlink"
+version = "0.16.0"
+source = "git+https://github.com/rust-netlink/rtnetlink.git#2423645faed3fdad55eb8ea9fe7ea8b0649e6b26"
```
qmonnet, line 1487: `cargo deny` also complains about this. I'm not sure how this ended up here, did you set the source manually?

qmonnet:
And I guess these:

image

Can be fixed by adding `publish = false` for the new crates?

net/src/route/mod.rs
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Common abstractions used in routing.
```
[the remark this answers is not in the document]
qmonnet: OK I see you use it in net/src/interface/vrf.rs so we probably don't want it in `routing`. I wonder if `net` is the right place for the `interface` submodule, though :thinking:

Fredi-raspall:
@daniel-noland , nice work, I still need to go through some parts of it to understand it better. Some early feedback, given that the PR is in DRAFT:
* please, open separate PRs for refactors / cleanups not directly related to the present PR. I think this is useful (and takes little time) because it does not add noise to the PR, allows having a separate discussion and helps those clean-ups / refactors make it sooner to the base; so, if there are other outstanding PRs, rebases are easier. No need to change this in this PR, just a recommendation for subsequent ones.
* I'm not too sure that the interface manager code should live in `net`. Maybe there's a reason, but wouldn't it make more sense to move it out of there? I'm fine if it has, as dependencies, other smaller crates, but I think the interface manager is pretty unrelated to all the Packet abstraction and parsing?
* Similarly the route submodule in net: I think it is unrelated to net. Also, shouldn't it be part of the VRF interface? Is it used elsewhere?
* I see you have a number of definitions for Interface, etc... which are identical to the ones in routing. I wish we'd had the time to unite those. 
* Regarding the vpc_manager crate, is there any chance to make it test module? I think it is great to validate the whole interface creation and reconciliation work (and should be by all means kept), but it may clash with other work aiming at configuring the GW as a whole. I think that this is indeed the reason for it? (Correct me if I am wrong).  Otherwise,  I'd : 1) not make it a crate 2) just for clarity not use the term VPC there but s/t like "Sample" something. Finally, can we make the `reconcile_test` test be such that it does not leave any interfaces when done?

net/src/route/mod.rs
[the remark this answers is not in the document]
qmonnet: This part has moved to the part-1 PR.

interface-manager/src/netns.rs
```diff
@@ -0,0 +1,119 @@
...
+        .spawn(move || {
+            #[allow(clippy::expect_used)] // the inability to swap to the other netns is fatal
+            #[allow(unsafe_code)] // uses external linux FFI
+            unsafe { swap_thread_to_netns(&netns_str) }.expect("failed to swap to netns");
+            #[allow(clippy::unwrap_used)] // the inability to start tokio is fatal
+            let tokio_runtime = tokio::runtime::Builder::new_current_thread()
+                .enable_io()
+                .enable_time()
+                .build()
+                .expect("failed to build tokio runtime");
```
qmonnet, line 53: `unwrap()` no longer used, it's `expect()` now :slightly_smiling_face:

rekon/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "rekon"
+version = "0.1.0"
+edition = "2024"
+license = "Apache-2.0"
```
qmonnet, line 5:
```suggestion
[package]
name = "rekon"
version = "0.1.0"
edition = "2024"
license = "Apache-2.0"
publish = false
```

Cargo.toml
```diff
@@ -70,6 +69,7 @@ ipnet = { version = "2.11.0", default-features = false, features = [] }
 iptrie = { version = "0.10.3", default-features = false, features = [] }
 mio = { version = "1.0.3", default-features = false, features = [] }
 multi_index_map = { version = "0.12.1", default-features = false, features = [] }
+netdev = { git = "https://github.com/githedgehog/netdev", branch = "pr/daniel-noland/bump-rtnetlink", default-features = false, features = [] }
```
qmonnet, line 72: The upstream PR you created and mentioned in the commit log has been merged, so this change can probably be reverted?

Cargo.lock
```diff
@@ -692,6 +723,16 @@ version = "2.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
+[[package]]
+name = "fixin"
+version = "0.0.1"
+source = "git+https://github.com/githedgehog/fixin?branch=main#9afddc5d845e3d7c769560be659a6cb0fbd148e2"
```
qmonnet, line 706: Why is this necessary?

Cargo.toml
daniel-noland: lovely.  Will do

interface-manager/src/netns.rs
daniel-noland: good catch

Cargo.toml
daniel-noland: I like the idea of merging vpc-manager with mgmt

Cargo.lock
daniel-noland: The test-runner.sh launches a container which only has libc.  Proc macro crates require the rust standard library tho and this breaks our test runner.  The only easy answer was to move fixin to another repo (I can't specify the test runner on a per crate basis sadly)

rekon/README.md
```diff
@@ -0,0 +1,7 @@
+# Rekon
+
+Rekon is a [facade] for an abstract reconciliation engine.
```
qmonnet, line 3: My French screams at the absence of cedilla (“façade”). I'll try to deal with it :face_with_peeking_eye:

qmonnet (approved):
I'm still not convinced that we need so many crates (we had mentioned merging the `vpc-manager` with `mgmt`, for example), but this is something we can revisit later if necessary.

Awesome work!

daniel-noland:
> I'm still not convinced that we need so many crates (we had mentioned merging the `vpc-manager` with `mgmt`, for example), but this is something we can revisit later if necessary.
> 
> Awesome work!

I'm completely with you on this one.  I want to do a lot of cleanup and de-dupliation after this demo is done

PR #401, opened 2025-04-14
net: Fix serde traits derivation (only available when "serde" feature is enabled)

qmonnet:
The use of dependency crate `serde` in the `net` package is optional, based on the `serde` feature for the package. As a consequence, we cannot always derive the serde traits without taking precautions inside of the package.

This issue has crept in in spite of the CI checks, because CI does not build every package independently. We observed the build failure with:

    $ just cargo build -p net

We then found the relevant commit that introduced the issue by bissecting, and observed we could fix the issue, on top of that commit, by enabling `serde` by running instead:

    $ just cargo build -p net -F serde

However, the issue doesn't appear at all by running only:

    $ just cargo build

This is likely because when building from the root of the repository, we must address requirements from all packages, including the `dataplane` package that requires the `net` package to be built with its `serde` feature, thus masking the issue.

Anyway, fix the bug by making traits derivation conditional to the use of the `serde` feature in ethtype.rs, then in interface/mod.rs.

### DIFF
--- a/net/src/eth/ethtype.rs
+++ b/net/src/eth/ethtype.rs
@@ -17,10 +17,9 @@ use etherparse::EtherType;
 /// 2. Permit the implementation of the `TypeGenerator` trait on this type
 ///    to allow us to property test the rest of our code.
 #[repr(transparent)]
-#[derive(
-    Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, serde::Serialize, serde::Deserialize,
-)]
-#[serde(from = "u16", into = "u16")]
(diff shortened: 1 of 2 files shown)


qmonnet:
@daniel-noland Do you have an idea on how to check this in CI to avoid this sort of bugs to make it into the tree again? I suppose we could build each package independently, turning each feature on/off to check that everything's alright, but it would probably add quite some overhead to the CI runs. Any other suggestion?

Fredi-raspall (approved):
LGTM @qmonnet , thanks!

PR #402, opened 2025-04-15
fix: rework id crate

daniel-noland:
I realized that the requirement to use a type alias hack in the id crate is fundamentally stemming from failing to put the `?Sized` bound on the phantom type.

This commit reworks that crate and significantly reduces its complexity.

### DIFF
--- a/id/src/lib.rs
+++ b/id/src/lib.rs
@@ -1,13 +1,17 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
-//! A "typed" [UUID] crate.
+//! A typed identifier.
 //!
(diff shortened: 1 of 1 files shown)


daniel-noland:
The idea for this just hit me over the weekend and so I cooked it up before I forgot about it

qmonnet (commented):
I must admit I'm a bit lost, trying to review this PR.

1. I don't understand the main change, would you mind summarising what's happening here, please?
2. Is the `fixin/Cargo.lock` related at all to the change?
3. Are the `serde` changes related, too?
4. Why do you enforce inlining of several functions, rather than leaving it to the compiler to decide?

daniel-noland:
Sorry about sleep deprived Dan's commits.  This should be ready now

mvachhar (approved):
Glad that ugly abstract type hack is gone.

PR #403, opened 2025-04-15
fix(net): serde should not be optional

daniel-noland:
I originally had dreams of making the net crate fully support no-std and limited resource environments. As a result, I made things optional which, in retrospect, have no need to be. Serde is one of those things.

So here ya go, with appologies to @qmonnet who just fixed this feature I'm removing.


qmonnet (approved):
Looks all in order, thanks!

> So here ya go, with appologies to @qmonnet who just fixed this feature I'm removing.

No worries at all :slightly_smiling_face:. I saw yesterday's PR as a quick fix anyway, I'm happy to see this additional complexity go away; I agree we don't need it for now. We can always revisit in the future if necessary.

PR #412, opened 2025-04-15
Test runner

daniel-noland:
Cargo automatically runs this script for every unit test (this applies to nextest as well).
The script has two main responsibilities:

1. It runs `setcap` on the _test binary_ to elevate the test's _permitted_ capabilities.
   This action _does not_ cause the tests to run with these capabilities active by default.
   That would involve setting the _effective_ capabilities for the test binary (which we don't do).
   Instead, assigning the _permitted_ capabilities allows the use of the `caps` crate to allow us to request elevated
   permissions for specific sections of test code.

   The purpose of these elevated privileges is to allow the tests to create and destroy virtual network interfaces and
   network namespaces (as is required for integration testing).

2. It bind mounts the (setcap modified) test binary, the project directory, and a few other files into a (read-only)
   docker container (which executes the test).  This docker container contains _only_ libc and libgcc_s (to better
   simulate our deployment environment and discourage faulty assumptions about what will be available at runtime).

   The purpose of this container is to
   * minimize the damage a faulty test might do
   * make sure that we aren't relying on resources only available on the developer's machine in the tests (test like
     we are in prod).

Hopefully, this process also requires us to carefully think about what parts of our code require which privileges (and
to document these requirements carefully).  I'm lookin' at you, future me :)


daniel-noland:
This is ready for review but can't acutally land until https://github.com/githedgehog/dpdk-sys/pull/112 hits main and the versioning hack in this PR is removed

daniel-noland:
> Looks very good to me @daniel-noland , thanks!

Thanks!

I fixed a few typos but will merge when CI finishes.

PR #413, opened 2025-04-16
feat: add back nextest config

daniel-noland:
nextest is a more modern test runner / reporting tool than the one which ships with cargo. nextest is significantly faster than cargo test and can report as junit et al.

To run nextest tests you can call

```
just cargo nextest run
```

It was used before, but I removed it in the name of making the justfile less complex.

Since then the justfile has been refactored to the point where nextest is very low cost to support, so we may as well add it back.  I like the faster tests.

Note: we can't currently support nextest in the sterile build env as I have yet to add it to the compile-env.  I may do this later if folks want it.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,133 @@
+# This is the default config used by nextest. It is embedded in the binary at
+# build time. It may be used as a template for .config/nextest.toml.
+
+[store]
+# The directory under the workspace root at which nextest-related files are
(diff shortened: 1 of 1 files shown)


mvachhar (approved):

qmonnet:
> Note: we can't currently support nextest in the sterile build env [...]

@daniel-noland ... but could it be added to dev.yml at least? This way we'd at least have some non-reg for it? Although I suppose the test suite is strictly the same as for `cargo test`?

PR #418, opened 2025-04-16
feat: dynamically link libc for production binaries

daniel-noland:
Static linking is my preference for containerized binaries, but it is somewhat in the way in regard to fuzz testing and the use of sanitizers (which is objectively more important in this case). Specifically, fuzz tests want the sanitizers on and many sanitizers require a dynamically linked libc.

Given that we now have a perfectly serviceable libc container anyway, we may as well use that for our production container. It is what we are testing against after all.


daniel-noland:
turns out this in't needed just yet

PR #419, opened 2025-04-17
Interface-manager part 1

daniel-noland:
Part one of the long-awaited interface manager pr.

I expect a part two on top of this shortly.

### DIFF
--- a/net/src/interface/vtep.rs
+++ b/net/src/interface/vtep.rs
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use crate::ipv4::UnicastIpv4Addr;
+use crate::vxlan::Vni;
+use derive_builder::Builder;
+use multi_index_map::MultiIndexMap;
+use serde::{Deserialize, Serialize};
+
+/// Vtep (vxlan device) specific properties
+#[derive(
+    Builder,
+    Clone,
+    Debug,
+    Deserialize,
+    Eq,
+    Hash,
+    MultiIndexMap,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    Serialize,
+)]
+#[multi_index_derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct VtepProperties {
+    /// The VNI associated with this vtep.
+    /// This value can be `None` in the event of an "external" vtep.
+    #[multi_index(hashed_unique)]
+    pub vni: Option<Vni>,
+    /// The local ip address to be associated with this vxlan device.
+    /// I.e., the source ip address to be used by encapsulating packets.
+    /// This value can be `None` if the vtep did not have a local value specified when it was
+    /// created.
+    #[builder(default)]
+    pub local: Option<UnicastIpv4Addr>,
+    /// The TTL of the vtep.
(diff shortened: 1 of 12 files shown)


net/src/interface/vtep.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// The local ip address to be associated with this vxlan device.
+    /// I.e., the source ip address to be used by encapsulating packets.
+    /// This value can be `None` if the vtep did not have a local value specified when it was
+    /// created.
+    #[builder(default)]
+    pub local: Option<SourceIpv4Addr>,
+    /// The TTL of the vtep.
+    /// This value can be `None` if the use did not specify the value when the VTEP was created or,
+    /// for whatever reason, the value is not available in the netlink message describing the vtep.
+    #[builder(default = Some(0))]
```
Fredi-raspall, line 39: Could it be that Mac has been forgotten here?

Fredi-raspall: I'd make it optional, if included.

net/src/ipv4/addr.rs
```diff
@@ -6,6 +6,39 @@
...
+pub struct SourceIpv4Addr(Ipv4Addr);
+
+impl SourceIpv4Addr {
+    /// Map an unqualified [`Ipv4Addr`] to a [`UnicastIpv4Addr`].
+    ///
+    /// # Errors
+    ///
+    /// Returns the supplied address back in the [`Err`] case if it is not a unicast address.
+    pub fn new(ip: Ipv4Addr) -> Result<SourceIpv4Addr, Ipv4Addr> {
+        if ip.is_multicast() || ip.is_unspecified() {
```
Fredi-raspall, line 28: In theory the unspecified ip address 0.0.0.0 can appear in some packets (e.g. DHCP).
I don't know if forbidding them here may cause trouble depending on the use of `SourceIpv4Addr` ?

Fredi-raspall (changes requested):
@daniel-noland looks good to me. The only comments I have relate to SourceIpv4Addr and the MAC for the vxlan interface.

net/src/ipv4/addr.rs
daniel-noland: That is a really good point.

I know that linux  forbids 0.0.0.0 as the source address of a vtep

net/src/interface/vtep.rs
daniel-noland: Mac is part of the parent structure

daniel-noland:
@Fredi-raspall

I addressed the thing about SourceIpv4.  That was a very good point.  That type was not sound.

I also added one commit at the end to clarify how mac is handled.

net/src/route/mod.rs
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Common abstractions used in routing.
```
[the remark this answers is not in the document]
qmonnet: Discussed during the meeting - I understand we're fine with this for now, and we'll try to refactor it One Day (TM). So I'm OK with the `route` submodule.

qmonnet (approved):
Looks good to me, thanks!
(I'd recommend squashing the last two commits into the relevant prior commits before merging.)

PR #421, opened 2025-04-17
Build fixups

daniel-noland:
Random build fixups to make cargo-deny happy.

This blocks the interface manger part 2 PR


Cargo.toml
```diff
@@ -41,7 +41,7 @@ routing = { path = "./routing" }
 
 # External
 afpacket = { version = "0.2.3", default-features = false, features = [] }
-ahash = { version = "0.8.11", default-features = false, features = [] }
+ahash = { git = "https://github.com/githedgehog/aHash", branch = "pr/daniel-noland/bump-zero-copy", default-features = false, features = ["no-rng"] }
```
mvachhar, line 44: Why do we need to bump aHash to get a new version of zerocopy?

mgmt/src/rpc/overlay/vpc.rs
```diff
@@ -12,6 +12,8 @@ use std::collections::BTreeSet;
 
 use crate::rpc::overlay::VpcManifest;
 use crate::rpc::overlay::VpcPeeringTable;
+#[cfg(doc)]
```
Fredi-raspall, line 15: why is this needed?

Fredi-raspall: @daniel-noland , sorry I forgot to review this. I'm not quite sure about the motivation, but you can go ahead and merge if it's blocking you.

Cargo.toml
daniel-noland: The zerocopy bump is related to the cargo-deny requirement that we don't have duplicate versions of dependencies linked into our production artifacts.
Some other crates picked the new version of zerocopy but ahash hadn't updated.  This bounced our build

mgmt/src/rpc/overlay/vpc.rs
daniel-noland: because this import is only used in the doc comments

daniel-noland (commented):
Hopefully I have clarified.

mvachhar (approved):

PR #423, opened 2025-04-22
More management changes

Fredi-raspall:


mgmt/src/models/internal/device/settings.rs
```diff
@@ -23,28 +23,19 @@ pub enum PacketDriver {
 
 #[derive(Clone, Debug)]
 pub struct DeviceSettings {
-    hostname: Option<String>,
+    hostname: String,
```
qmonnet: Why make it mandatory?

qmonnet (approved):
Looks good, thank you!

mgmt/src/models/internal/device/settings.rs
Fredi-raspall: It simplifies the code and we can always default it to something if needed.

sergeymatov (approved):

PR #427, opened 2025-04-23
Management config processing logic

Fredi-raspall:
**N.B:** this now targets https://github.com/githedgehog/dataplane/pull/465 after rebasing on latest main

### DIFF
--- a/mgmt/src/frr/renderer/routemap.rs
+++ b/mgmt/src/frr/renderer/routemap.rs
@@ -185,7 +185,7 @@ mod tests {
 
         let mut rmap_table = RouteMapTable::new();
         rmap_table.add_route_map(rmap);
-        println!("{:#?}", rmap_table);
+        println!("{rmap_table:#?}");
         println!("{}", rmap_table.render(&()));
     }
 }
--- a/mgmt/src/frr/renderer/routemap.rs
+++ b/mgmt/src/frr/renderer/routemap.rs
@@ -1,0 +1,191 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Config renderer: route maps
+
+use crate::frr::renderer::builder::{ConfigBuilder, MARKER, Render};
+use crate::models::internal::routing::routemap::*;
+use std::fmt::Display;
+
+/* Impl Display */
+impl Display for MatchingPolicy {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            MatchingPolicy::Deny => write!(f, "deny"),
+            MatchingPolicy::Permit => write!(f, "permit"),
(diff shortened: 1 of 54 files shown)


qmonnet:
I pushed a commit on top of your branch to fix the reports from Clippy.

mgmt/src/frr/renderer/vrf.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn render_vrf_bgp(&self) -> ConfigBuilder {
+        let mut cfg = ConfigBuilder::new();
+        self.iter().for_each(|vrf| cfg += vrf.render_vrf_bgp());
+        cfg
+    }
+}
+
+#[cfg(test)]
+#[allow(dead_code)]
+mod tests {
```
qmonnet, line 76: Is the `#[allow(dead_code)]` required here?

mgmt/src/frr/renderer/routemap.rs
```diff
@@ -0,0 +1,191 @@
...
+    type Output = ConfigBuilder;
+    fn render(&self, _: &Self::Context) -> ConfigBuilder {
+        let mut cfg = ConfigBuilder::new();
+        self.values().for_each(|plist| cfg += plist.render(&()));
+        cfg
+    }
+}
+
+#[cfg(test)]
+#[allow(dead_code)]
```
qmonnet, line 143: Same question (and for your other `tests` submodules in general)

mgmt/src/frr/renderer/routemap.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let entry = RouteMapEntry::new(40, MatchingPolicy::Deny);
+        rmap.add_entry(entry);
+        rmap
+    }
+
+    #[test]
+    fn test_route_map_render() {
+        let rmap = build_test_route_map();
+        //println!("{}", rmap.render(&()));
```
qmonnet, line 184: Uncomment or delete?

qmonnet (approved):
This PR looks good to me. I trust you blindly with the rendering of the FRR config for BGP or OSPF :see_no_evil:

I note how you trigger the config build from mgmt/src/processor/proc.rs, I'll probably need to hook somewhere around there for the NAT config.

Trivial questions/nits below are not blocking.

daniel-noland (approved):
I actually have a lot of thoughts about this PR and what we might do a little differently, but I think we should merge this now and I can address some of those changes over the course of the vpc/interface-manager integration

Nice work.

qmonnet:
> I actually have a lot of thoughts about this PR and what we might do a little differently

@daniel-noland This sounds somewhat concerning to me. The Pull Request has remained open for nearly three weeks, why not voice these thoughts during the review process? Sounds like there's something we're not doing right for reviews, here. Is there something we could have done differently to get your input?

PR #429, opened 2025-04-24
Fix Clippy reports (nightly)

qmonnet:
- **style(routing): Fix Clippy reports**
- **build: Allow passing arguments to "just clippy"**

### DIFF
--- a/justfile
+++ b/justfile
@@ -401,7 +401,7 @@ push-container: build-container
 
 # Run Clippy like you're in CI
 [script]
-clippy: (cargo "clippy" "--all-targets" "--all-features" "--" "-D" "warnings")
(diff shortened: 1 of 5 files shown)


sergeymatov:
@qmonnet does this a result of clippy fix?

qmonnet:
@sergeymatov Yes, Clippy has been complaining about these in CI, in the job with the nightly toolchain (although I couldn't reproduce locally, but I haven't tried super hard).

qmonnet:
You can see the Clippy reports here, for example: https://github.com/githedgehog/dataplane/actions/runs/14644622746/job/41095468884?pr=430

sergeymatov (approved):
I'm totally ok with this.

PR #430, opened 2025-04-24
CI: Remove comments on Nightly job failures

qmonnet:
I don't like the comments generated by the CI workflow when the job with the Nightly toolchain fails, I find them disruptive when trying to follow a conversation on a GitHub Pull Request page. Plus given that the job is optional, we barely ever click the links from the comments anyway. A failed status check draws more attention to the actual issue, in my opinion.

Motivation:
- My objective here is to remove the creation of the comments when an optional step fails
- To get some feedback on failure instead, we remove the `continue-on-error` directives: I expect this to make the job using the Nightly toolchain fail on errors (which should show up in the list of status checks at the bottom of the page).

Will a job failure block the CI or the merge queue?
- The individual jobs for the dev.yml workflow are not marked as required, only the last `Summary` step is required
- The job with Nightly still has an `optional` attribute in the Matrix, and as a consequence its status is ignored when building the `Summary`, so I don't expect it to block CI/merge queue.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -66,11 +66,6 @@ jobs:
   build:
     needs: [ check_changes ]
     if: "${{ needs.check_changes.outputs.devfiles == 'true' }}"
-    permissions:
-      issues: "write"
-      pull-requests: "write"
-      contents: "write"
-      id-token: "write"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


qmonnet:
Clippy reports on nightly build are not related to this PR and should be addressed with https://github.com/githedgehog/dataplane/pull/429.

But the failure does show that:

1. We get failures again for the nightly build
2. It doesn't seem to block the merge

So, all good!

qmonnet:
Ping @daniel-noland, can you take a quick look please?

mvachhar (approved):

PR #438, opened 2025-05-06
test(test-runner): Per-package opt-in for test runner

qmonnet:
The test runner that makes unit tests run inside of a Docker container tend to make it harder to run and debug tests. Only use the container for packages that do require specific configuration or elevated privileges, not for the other packages.

Code from Daniel.


qmonnet:
I'll merge this tomorrow if nobody objects.

daniel-noland (approved):
LGTM

PR #439, opened 2025-05-06
chore(dependabot): Make Dependabot ignore gateway_config updates

qmonnet:
Updates for gateway_config require manual adjustments in our code for now, we need someone to take care of them. Attempts from the bot to autmatically bump the version number (or commit SHA) are unlikely to be correct, so let's discharge Dependabot of the task.


daniel-noland (approved):
This suggests an issue with repository layout but for now it seems reasonable

Issue #442, opened 2025-05-06
Validate gRPC objects


qmonnet:
Validate objects that we receive (in the `mgmt` package) from the gRPC server.

They should be checked for soundness, and also from the perspective of different components as necessary. For example, static NAT requires the same amount of available IP addresses in the public and private sides of a `VpcExpose` object.

PR #445, opened 2025-05-08
mgmt: Validate VpcPeering & friends

qmonnet:
This Pull Request implements some validation for `VpcExpose`, `VpcManifest`, `VpcPeering`, and `VpcTable` objects received from the gRPC server.

Fixes: #442

Please refer to individual commit descriptions for details.

### DIFF
--- a/mgmt/src/models/external/overlay/vpc.rs
+++ b/mgmt/src/models/external/overlay/vpc.rs
@@ -170,4 +170,21 @@ impl VpcTable {
     pub fn clear_ids(&mut self) {
         self.ids.clear();
     }
+    /// Validate the [`VpcTable`]
+    pub fn validate(&self) -> ConfigResult {
+        for vpc in self.values() {
+            let mut peers = BTreeSet::new();
+            // For each VPC, loop over all peerings
+            for peering in &vpc.peerings {
+                // Check whether we have duplicate remote VPCs between peerings.
+                // If we fail to insert, this means the remote VPC ID is already in our set,
+                // and we have a duplicate peering: this is a configuration error.
+                if (!peers.insert(peering.remote_id.clone())) {
+                    return Err(ConfigError::DuplicateVpcPeerings(peering.name.clone()));
+                }
+            }
+            peers.clear();
+        }
+        Ok(())
+    }
 }
--- a/mgmt/src/models/external/overlay/vpcpeering.rs
+++ b/mgmt/src/models/external/overlay/vpcpeering.rs
@@ -141,6 +141,27 @@ impl VpcManifest {
             ..Default::default()
         }
     }
+    fn validate_expose_collisions(&self) -> ConfigResult {
+        // Check that prefixes in each expose don't overlap with prefixes in other exposes
+        for (index, expose_left) in self.exposes.iter().enumerate() {
+            // Loop over the remaining exposes in the list
+            for expose_right in self.exposes.iter().skip(index + 1) {
+                validate_overlapping(
+                    &expose_left.ips,
+                    &expose_left.nots,
+                    &expose_right.ips,
+                    &expose_right.nots,
+                )?;
+                validate_overlapping(
+                    &expose_left.as_range,
+                    &expose_left.not_as,
+                    &expose_right.as_range,
+                    &expose_right.not_as,
+                )?;
+            }
+        }
+        Ok(())
+    }
     pub fn add_expose(&mut self, expose: VpcExpose) -> ConfigResult {
         self.exposes.push(expose);
         Ok(())
@@ -152,6 +173,7 @@ impl VpcManifest {
         for expose in &self.exposes {
             expose.validate()?;
         }
+        self.validate_expose_collisions()?;
         Ok(())
     }
 }
@@ -229,4 +251,123 @@ impl VpcPeeringTable {
             .values()
             .filter(move |p| p.left.name == vpc || p.right.name == vpc)
     }
+}
+
+// Validate that two sets of prefixes, with their exclusion prefixes applied, don't overlap
+fn validate_overlapping(
+    prefixes_left: &BTreeSet<Prefix>,
+    excludes_left: &BTreeSet<Prefix>,
+    prefixes_right: &BTreeSet<Prefix>,
+    excludes_right: &BTreeSet<Prefix>,
+) -> Result<(), ConfigError> {
+    // Find colliding prefixes
+    let mut colliding = Vec::new();
+    for prefix_left in prefixes_left.iter() {
+        for prefix_right in prefixes_right.iter() {
+            if prefix_left.covers(prefix_right) || prefix_right.covers(prefix_left) {
+                colliding.push((*prefix_left, *prefix_right));
+            }
+        }
+    }
+    // If not prefixes collide, we're good - exit.
+    if colliding.is_empty() {
+        return Ok(());
+    }
+
+    // How do we determine whether there is a collision between the set of available addresses on
+    // the left side, and the set of available addresses on the right side? A collision means:
+    //
+    // - Prefixes collide, in other words, they have a non-empty intersection (we've checked that
+    //   earlier)
+    //
+    // - This intersection is not fully covered by exclusion prefixes
+    //
+    // The idea in the loop below is that for each pair of colliding prefixes:
+    //
+    // - We retrieve the size of the intersection of the colliding prefixes. This is easy, because
+    //   they're "prefixes", so if they collide we have necessarily one that is contained within the
+    //   other, and the size of the intersection is the size of the smallest one.
+    //
+    // - We retrieve the size of the union of all the exclusion prefixes (from left and right sides)
+    //   covering part of this intersection (which we know is the smallest of the two colliding
+    //   prefixes). The union of the exclusion prefixes is the set of non-overlapping exclusion
+    //   prefixes that cover the intersection of allowed prefixes, such that if exclusion prefixes
+    //   collide, we always keep the largest prefix.
+    //
+    // - If the size of the intersection of colliding allowed prefixes is bigger than the size of
+    //   the union of the exclusion prefixes applying to them, then it means that some addresses are
+    //   effectively allowed in both the left-side and the right-side set of available addresses,
+    //   and this is an error. If the sizes are identical, then all addresses in the intersection of
+    //   the prefixes are excluded on at least one side, so it's all good.
+    for (prefix_left, prefix_right) in colliding {
+        // If prefixes collide, there's necessarily one prefix that is contained inside of the
+        // other. Find the intersection of the two colliding prefixes, which is the smallest of the
+        // two prefixes.
+        let intersection_prefix = if prefix_left.covers(&prefix_right) {
+            &prefix_right
+        } else {
+            &prefix_left
+        };
+
+        // Retrieve the union of all exclusion prefixes covering the intersection of the colliding
+        // prefixes
+        let mut union_excludes = BTreeSet::new();
+
+        // Consider exclusion prefixes from excludes_left
+        'outer: for exclude_left in excludes_left.iter().filter(|exclude| {
+            exclude.covers(intersection_prefix) || intersection_prefix.covers(exclude)
+        }) {
+            for exclude_right in excludes_right.iter().filter(|exclude| {
+                exclude.covers(intersection_prefix) || intersection_prefix.covers(exclude)
+            }) {
+                if exclude_left.covers(exclude_right) {
+                    // exclude_left contains exclude_right, and given that exclusion prefixes in
+                    // list excludes_right don't overlap there's no exclusion prefix containing
+                    // exclude_left. We want to keep exclude_left as part of the union.
+                    union_excludes.insert(*exclude_left);
+                    continue 'outer;
+                } else if exclude_right.covers(exclude_left) {
+                    // exclude_left is contained within exclude_right, don't keep it as part of the
+                    // union. Process next exclusion prefix from list excludes_left.
+                    continue 'outer;
+                }
+            }
+            // No collision for this exclude_left, add it to the union
+            union_excludes.insert(*exclude_left);
+        }
+        // Consider exclusion prefixes from excludes_right
+        'outer: for exclude_right in excludes_right.iter().filter(|exclude| {
+            exclude.covers(intersection_prefix) || intersection_prefix.covers(exclude)
+        }) {
+            for exclude_left in excludes_left.iter().filter(|exclude| {
+                exclude.covers(intersection_prefix) || intersection_prefix.covers(exclude)
+            }) {
+                if exclude_right.covers(exclude_left) {
+                    // exclude_right contains exclude_left, and given that exclusions prefixes in
+                    // list excludes_left don't overlap there's no exclusion prefix containing
+                    // exclude_right. We want to keep exclude_right as part of the union.
+                    union_excludes.insert(*exclude_right);
+                    continue 'outer;
+                } else if exclude_left.covers(exclude_right) {
+                    // exclude_right is contained within exclude_left, don't keep it as part of the
+                    // union. Process next exclusion prefix from list excludes_right.
+                    continue 'outer;
+                }
+            }
+            // No collision for this exclude_right, add it to the union
+            union_excludes.insert(*exclude_right);
+        }
+
+        let union_size = union_excludes
+            .iter()
+            .map(|exclude| exclude.size())
+            .sum::<u128>();
+        if union_size < intersection_prefix.size() {
+            // Some addresses at the intersection of both prefixes are not covered by the union of
+            // all exclusion prefixes, in other words, they are available from both prefixes. This
+            // is an error.
+            return Err(ConfigError::OverlappingPrefixes(prefix_left, prefix_right));
+        }
+    }
+    Ok(())
 }
--- a/mgmt/src/models/external/overlay/vpcpeering.rs
+++ b/mgmt/src/models/external/overlay/vpcpeering.rs
@@ -142,8 +142,16 @@ impl VpcManifest {
         }
     }
     pub fn add_expose(&mut self, expose: VpcExpose) -> ConfigResult {
-        expose.validate()?;
         self.exposes.push(expose);
+        Ok(())
+    }
+    pub fn validate(&self) -> ConfigResult {
+        if self.name.is_empty() {
+            return Err(ConfigError::MissingIdentifier("Manifest name"));
+        }
+        for expose in &self.exposes {
+            expose.validate()?;
+        }
         Ok(())
     }
 }
@@ -164,6 +172,8 @@ impl VpcPeering {
     }
     pub fn validate(&self) -> ConfigResult {
         debug!("Validating VPC peering '{}'...", &self.name);
+        self.left.validate()?;
+        self.right.validate()?;
         Ok(())
     }
     /// Given a peering fetch the manifests, orderly depending on the provided vpc name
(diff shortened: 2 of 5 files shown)


Fredi-raspall:
@qmonnet I would strongly recommend rebasing this work on https://github.com/githedgehog/dataplane/pull/441
Otherwise you're going to be fixing nits already fixed and add changes to code that has  been significantly modified.

Fredi-raspall (changes requested):
I believe it is not useful to review this PR until rebased.

qmonnet:
I'm not rebasing on the top of three different PRs. Rebases in any of the three will cause me to need new rebases. (Although, there doesn't seem to be that many conflicts between your code and mine - mostly the changes and renaming of `ApiError`, I think).

If you prefer your sets to go in first - totally understandable - I'll wait for them to get merged, and then rebase.

qmonnet:
Oh, you did rebase :exploding_head:
Thank you!

mgmt/src/grpc/test.rs
```diff
@@ -569,21 +569,55 @@ mod tests {
                     original_entry.vpc
                 );
 
+                fn get_filtered_ips_count(ips: &[gateway_config::PeeringIPs]) -> usize {
```
Fredi-raspall, line 572: I don't understand the use of this. Also, shouldn't the validation be in the configuration oustide the gRPC. IIRC we discussed that its better to do that there and let the gRPC remain shallow. The rationale is: if ever the "transport" changes or another one is added, the validation would  remain if not done here.

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -53,11 +198,49 @@ impl VpcManifest {
...
+            validate_overlapping(
+                &expose.as_range,
+                &expose.not_as,
+                &other_expose.as_range,
+                &other_expose.not_as,
+            )?;
+        }
+        Ok(())
+    }
+    pub fn add_expose(&mut self, new_expose: VpcExpose) -> ConfigResult {
```
Fredi-raspall, line 219: I would not do validation in object builders but later (at least that's what we've done so far). The reason is that we may want to  dump the whole config when rejecting it. So, IMO, we should assemble the config, completely, and then validate it. This also reduces the scope of each method and the number of places where validations are done.

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -35,8 +36,152 @@ impl VpcExpose {
         self.not_as.insert(prefix);
         self
     }
+    fn fixup(mut self) -> Result<Self, ConfigError> {
```
Fredi-raspall, line 39: I have some concerns about this, philosophical and practical.
I don't think we should be fixing users' configs (that's philosophical). If we fix a "typo", okay, but here you're actually augmenting the configuration. That creates implicit behavior that I believe we should avoid. So, IMO, we should just reject a config instead of fixing it.
The other reasons I'm concerned about this are:
* if we fixup a config, there's no way to tell if the config included the 0.0.0.0/0 or we added it. IMO, we should treat the config as a manifest. If something was lacking, complain. Else, mismatches between what we get and what we have (implicit behaviors) can be very confusing.
* Adding a 0.0.0.0/0 to the list of Ips semantically means "expose everything", which is the exact opposite of having left the list of ips empty. (Another discussion is whether we should accept an empty list of ips. IMO, we should not reject that. The GW is not going to make anything useful but it won't harm and we may find some use cases there such as  temporarily not allowing traffic to a VPC without unconfiguring it --- I'd be as less restrictive as possible).
*  "doping" the config with a 0.0.0.0/0 fixup may have undesired side effects (now or in the future) on the configuration of other parts of the system (e.g. routing) and, especially if done silently, add issues hard to detect / diagnose since they may not lead to bugs but change the intended behavior of the system in subtle (or not) ways.
* If this fixup() really helps for NAT, then I would "dope" the NAT-specific state. However, the same considerations apply wrt to implicit behaviors.

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -106,9 +291,43 @@ impl VpcPeeringTable {
...
+            }
+        }
+        Ok(())
+    }
     /// Add a [`VpcPeering`] to a [`VpcPeeringTable`]
     pub fn add(&mut self, peering: VpcPeering) -> ConfigResult {
         peering.validate()?;
+        self.validate_peering_addition(&peering)?;
+
+        // First look for an existing entry, to avoid inserting a duplicate peering
```
Fredi-raspall, line 326: The duplicate check is already performed in line 331.

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -106,9 +291,43 @@ impl VpcPeeringTable {
...
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
     /// Add a [`VpcPeering`] to a [`VpcPeeringTable`]
     pub fn add(&mut self, peering: VpcPeering) -> ConfigResult {
         peering.validate()?;
+        self.validate_peering_addition(&peering)?;
```
Fredi-raspall, line 324: The same comments apply here for the two validations: unless the input config would be glaringly wrong (e.g. with a duplicate peering id name, which would not a allow a later validation -- in this case because a peering would replace another one) , I would defer it and concentrate it in a single place. This can sound inefficient, but it actually isn't necessarily:  validate_peering_addition() method seems to require the whole set of peerings so far. If you would be validating once having added all of the peerings that would be ~ O(n²) ?. But, the current code has the same complexity.

Fredi-raspall (changes requested):

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: Right, I added them “early” to avoid building a broken object, but you make a good point, I'll change this.

mgmt/src/grpc/test.rs
qmonnet: > I don't understand the use of this.

This is related to the “fixup” step where we add a root prefix in the case of a set of addresses that is only defined by exclusion prefixes (see also commit description). The test attempts to compare the size of the "rules" (`expose` entries) before and after translating the gRPC object into a config object. With the “fixup” step in place, this number differs, because of the addition of the `0.0.0.0/0` prefix. We have no easy way to tell whether we added a prefix or not, so I adjusted the check to discard root-level prefixes on both sides when counting.

> Also, shouldn't the validation be in the configuration oustide the gRPC. IIRC we discussed that its better to do that there and let the gRPC remain shallow. The rationale is: if ever the "transport" changes or another one is added, the validation would remain if not done here.

I'm not adding any more validation here, I'm just adjusting the existing test so it doesn't break. Moving the tests is beyond the scope of this PR.

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: Agree, I'll change, just like for `add_expose()`.

Agree on complexity as well, in both cases we end up comparing the same number of prefixes, so there shouldn't be any particular impact.

mgmt/src/models/external/overlay/vpcpeering.rs
Fredi-raspall: You're absolutely right. I wonder if then we should change the code after the check, in particular wrt to the return value. It should probably succeed always and the Err be unreachable?

qmonnet: It should, at least as long as there's only one thread reading from the table. I hesitated to remove the other check, I could mark it as `unreachable!()` so that we fail loudly if the assumption ever turns out to no longer be valid

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: Discussed during today's meeting. My understanding, including the list of cases described above, are incorrect. When there's no allowed prefixes, then there's no exposed IP addresses, period.

As a consequence, it does not make sense to keep this fixup step because 1) if we wanted to apply it, we don't know the set of valid addresses in the VPC space that would “apply”, and 2) in fact, it's not necessary at all. I'll remove it and revert the related changes.

Thanks for the detailed feedback!

mgmt/src/models/external/overlay/vpcpeering.rs
Fredi-raspall: Either option is fine. You could actually remove the extra check you added:
* if everything is ok (no duplicate), the code will just do an insertion (instead of lookup + instertion)
* in the bad case (duplication), true, the insertion will overwrite the previous element. However, we'll be discarding the config. So, the most optimal option in terms of code and performance may be just inserting and complaining afterwards.

qmonnet: Unless you're opposed to it I think I prefer to keep the check.

I understand we currently reject the config on error, but we could imagine someone changing or reusing the code elsewhere and do something like “try to insert this rule, else fallback to that...” and we'll get the rule inserted anyway as a side effect. That's the case in my tests by the way, where I try to insert several rules successively, sometimes expecting an error, but not expecting rules to be inserted in case of error.

Performance-wise, you're correct, although I think the difference is minor, and this happens at config time not in the hot path, so it shouldn't matter.

qmonnet:
OK, we should be good for a second review

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -187,6 +281,144 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_validate_expose() {
```
daniel-noland, line 500: two things

1. why not make these distinct tests?
2. why not use bolero?

daniel-noland, line 500: two things

1. why not make these distinct tests?
2. why not use bolero?

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -144,10 +144,21 @@ impl VpcManifest {
         }
     }
     pub fn add_expose(&mut self, expose: VpcExpose) -> ConfigResult {
-        expose.validate()?;
```
daniel-noland, line 57: is it your intention to make it possible to construct invalid configurations?

Why remove the validation check here?

daniel-noland (approved):
Very nice work!

I especially appreciate the design comments.

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: This is based on Fredi's feedback from my first version:

- https://github.com/githedgehog/dataplane/pull/445#discussion_r2079843706
- https://github.com/githedgehog/dataplane/pull/445#discussion_r2079899667

The gist is that we prefer having validation in a single place, and that, yes, it's acceptable to build a broken config that we will reject at validation time, if only to show it back to the user.

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: 1. Not sure what you mean, you propose to move each of the different case to a separate function? They felt close enough (all about testing the validation for `VpcExpose`) that I could group them in a single function, but if you think it's better to have them separated into many smaller functions, I guess that's also a possibility.

2. Because I'm not familiar with bolero and haven't looked into it (or even thought about it, for that matter :face_with_peeking_eye:). I can, if you think it's worth it.

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -125,4 +257,489 @@ impl VpcPeeringTable {
...
+    //   other, and the size of the intersection is the size of the smallest one.
+    //
+    // - We retrieve the size of the union of all the exclusion prefixes (from left and right sides)
+    //   covering part of this intersection (which we know is the smallest of the two colliding
+    //   prefixes). The union of the exclusion prefixes is the set of non-overlapping exclusion
+    //   prefixes that cover the intersection of allowed prefixes, such that if exclusion prefixes
+    //   collide, we always keep the largest prefix.
+    //
+    // - If the size of the intersection of colliding allowed prefixes is bigger than the size of
+    //   the union of the exclusion prefixes applying to them, then it needs that some addresses are
```
qmonnet, line 358:
```suggestion
    //   the union of the exclusion prefixes applying to them, then it means that some addresses are
```

mgmt/src/models/external/overlay/vpcpeering.rs
daniel-noland: Ok.

I can live with that.

It might be good to have a validated type which wraps this one in that case to express the static requirement that you can't feed an invalid config to the dataplane.

mgmt/src/models/external/overlay/vpcpeering.rs
daniel-noland: Oh, I won't reject the PR on this basis.

I just think that bolero would be notably easier and more effective than this path.

Perhaps I should do a tutorial on how to use it.

qmonnet: I think you did a presentation on it at some point? I need to find the link again. If I'm getting confused and you haven't, then yes, I'd gladly watch/read it!

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: Yes, I can look into it as a follow-up.

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -126,3 +126,150 @@ impl VpcPeeringTable {
             .filter(move |p| p.left.name == vpc || p.right.name == vpc)
     }
 }
+
+#[cfg(test)]
```
Fredi-raspall, line 435: @qmonnet, I believe that some of the tests added here overlap with existing tests in https://github.com/githedgehog/dataplane/blob/pr/qmonnet/nat-validation/mgmt/src/models/external/overlay/tests.rs 
Did you double-check?

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -125,4 +257,489 @@ impl VpcPeeringTable {
             .values()
             .filter(move |p| p.left.name == vpc || p.right.name == vpc)
     }
+
+    fn vpcs(&self) -> HashSet<&String> {
```
Fredi-raspall, line 261: I believe this adds unnecessary code and steps.
As part of the overlay validation (https://github.com/githedgehog/dataplane/blob/pr/qmonnet/nat-validation/mgmt/src/models/external/overlay/mod.rs#L52) we collect all of the peerings where a VPC participates and add them to the VPC itself (see https://github.com/githedgehog/dataplane/blob/f42cd90d2f47e464b45a96c71ddecf4df5bb246e/mgmt/src/models/external/overlay/vpc.rs#L80). Can't you just validate the peerings once collected in the VPC to avoid this ?

mgmt/src/models/external/overlay/vpcpeering.rs
```diff
@@ -125,4 +257,489 @@ impl VpcPeeringTable {
...
+        let mut vpcs = HashSet::new();
+        self.0.values().for_each(|p| {
+            vpcs.insert(&p.left.name);
+            vpcs.insert(&p.right.name);
+        });
+        vpcs
+    }
+    fn validate_peering_collisions(&self) -> ConfigResult {
+        // For each VPC, check that exposes in manifests for this VPC in the different peerings do
```
Fredi-raspall, line 270: I am not sure to understand this requirement. Why would it be bad to expose the same set of prefixes over distinct peerings?

Fredi-raspall (changes requested):

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: There's some overlap in that we're testing closely related components, but in `test.rs` we test things build, in `vpcpeering.rs` we test they're valid - that's the distinction we talked about before. So I'm not sure the tests are plain redundant. `test_manifest_expose()` added here looks closely to the tests for `add_expose()` in `test.rs`, but it also checks the result of `get_peering_manifests()` at the end. It also seems that all cases returning errors in `test.rs` or `vpcpeering.rs` are distinct.

mgmt/src/models/external/overlay/vpcpeering.rs
qmonnet: I can, but this means that we need to pass the `vpc_table` from the outer `Overlay` to the `validate()` function for the `peering_table()`. I wasn't sure that was the best approach, that's why I had recomputed the set separately.

mgmt/src/models/external/overlay/vpcpeering.rs
Fredi-raspall: Hum. I believe that in test.rs we also test things like:
  * duplicated vpc names
  * that peerings refer to existing VPCs.
  * we test Overlay::validate()
  * the filtered lookups to find the peerings that a VPC participates in.

In any case, can you move the tests there so that they are in a single place and we can see what the tests cover altogether. The reason to add a separate module is not to bloat the functional code with large tests.

mgmt/src/models/external/overlay/vpcpeering.rs
Fredi-raspall: Between two VPCs, there should only exist one peering. You probably mean exposes here, right ?

qmonnet: > Between two VPCs, there should only exist one peering.

I didn't take that into account, hence the checks.

If there's just one peering max between any pair of VPCs, then indeed I need to do fewer checks here; but I think I need to add a check elsewhere to make sure that we only have one peering at most, I don't remember seeing a check for that

mgmt/src/models/external/overlay/vpcpeering.rs
Fredi-raspall: Thanks!

mgmt/src/models/external/overlay/vpcpeering.rs
Fredi-raspall: We should double-check with @Frostman and @sergeymatov , but I'd say that only one peering is needed.

I don't think we check that either atm. However, the check should be easy to do once the VPC peerings for a given VPC have been collected: there should not be two peerings with the same remote.name or remote_id. That approach requires checking the VPC's Peering objects (instead of the VpcPeering ones in the vpcpeering table), but that is somewhat more aligned with the idea of soon ceasing to use the VpcPeerings (and table), which exist because the API expresses peerings and VPCs separately, in favor of the Peering objects within VPCs.

qmonnet: Sergei confirmed we only have one peering at most. I'll change my checks (see also [my comment](https://github.com/githedgehog/dataplane/pull/445#discussion_r2091577312) on the discussion above).

qmonnet:
I've incorporated Fredi's feedback I think, this is ready for another review.

Fredi-raspall (changes requested):
I'd see if we can make the validations less heavy and ponder whether all are required. In general, I'd not "parent" users. We should just reject stuff that will cause failures and not forbid configs which, while seemingly useless, are legitimate.
Requesting @mvachhar a second opinion on the complexity / performance cost of the validations. Other than that, looks good to me @qmonnet !

mgmt/src/models/external/overlay/vpc.rs
```diff
@@ -170,4 +170,24 @@ impl VpcTable {
     pub fn clear_ids(&mut self) {
         self.ids.clear();
     }
+    /// Validate the [`VpcTable`]
```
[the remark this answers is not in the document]
qmonnet: Good idea. I agree that we care little about the name of the first peering here, I'll reuse your snippet. Thanks!

mvachhar (approved):
Approved, but please see issue #554 to have a validated and unvalidated set of types here in the future.   The unvalidated types make it much easier to test the gRPC conversion where we want to make sure the data gets translated but aren't as worried about higher level semantics.

PR #447, opened 2025-05-08
fix: bump rdma-core to 57.0

daniel-noland:
Upstream rdma-core changes the library names.


daniel-noland:
closed in favor of #450

PR #449, opened 2025-05-09
feat(routing): Implement From<&str> trait for Prefix

qmonnet:
Make it possible to create prefixes by passing a single string with the address/mask notation, for example:

```rust
let prefix_v4: Prefix = "1.2.3.0/24".into();
```

The objective is to write prefixes faster, mostly when building context for tests.

### DIFF
--- a/routing/src/prefix.rs
+++ b/routing/src/prefix.rs
@@ -182,6 +182,19 @@ impl<'a> From<&'a Prefix> for &'a Ipv6Prefix {
         }
     }
 }
+/// Only for testing. Will panic with badly formatted prefix strings
+#[cfg(any(test, feature = "testing"))]
+impl From<&str> for Prefix {
+    fn from(s: &str) -> Self {
+        if let Ok(p) = Ipv4Net::from_str(s) {
+            Prefix::IPV4(Ipv4Prefix::from(p))
+        } else if let Ok(p) = Ipv6Net::from_str(s) {
+            Prefix::IPV6(Ipv6Prefix::from(p))
+        } else {
+            panic!("Not a valid IP prefix")
+        }
+    }
+}
 
 impl Display for Prefix {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


Fredi-raspall:
> Make it possible to create prefixes by passing a single string with the address/mask notation, for example:
> 
> ```rust
> let prefix_v4: Prefix = "1.2.3.0/24".into();
> ```
> 
> The objective is to write prefixes faster, mostly when building context for tests.

I belive this functionality exists already in the routing branch to be integrated :disappointed: 
Let's see how to best proceed. I'd suggest that unless this blocks you, you keep it locally in your working branch until we can merge the other one.

routing/src/prefix.rs
```diff
@@ -142,6 +142,17 @@ impl From<Ipv6Prefix> for Prefix {
         Self::IPV6(value)
     }
 }
+impl From<&str> for Prefix {
```
daniel-noland, line 145: can we either

1. add `#[cfg(test)]` to this or
2. put this in a `tests` module?

I understand wanting this in tests, but panic is not an ok answer to invalid input on a day to day basis and we should not allow this type of logic into the rest of the dataplane.

daniel-noland (changes requested):

routing/src/prefix.rs
qmonnet: I'm fine with making them test-only, this was my use case anyway. There are some other surrounding implementations of the `From` trait that can `panic` as well, I can move them at the same time if it doesn't break anything.

qmonnet: (Assuming I come back to this PR, not sure based on what Fredi said.)

daniel-noland: That's fair.

Generally, I don't like impls of `From` or `Into` which can panic.

Another option is to use a macro for situations like this

qmonnet: How does the macro prevent the panic? You do have to parse the string into a valid prefix at one point or another?

Fredi-raspall: I agree with @daniel-noland with panicking in prod code. I've pushed 3 commits to the management cleanup PR (which should be merged soon) that address a very similar thing. Initially the routing crate had a testing-only converting function (annotated with `#[cfg(test)]`. Later on, the annotation was dropped (I believe by me) and as such it was used by gRPC prod code which was panicking when the input configuration contained garbage prefixes. The first commit is https://github.com/githedgehog/dataplane/pull/441/commits/4f939261055330b7fae04502572ae2de0453465b

I've annotated it back with `#[cfg(test)]`. However, to allow using it from other crates, I've had to define a feature flag (`'testing'`) in the routing crate and enable it wherever it is needed.

daniel-noland: > How does the macro prevent the panic? You do have to parse the string into a valid prefix at one point or another?

The `#[cfg(test)]` macro means that the panic implementation simply does not exist in production code.  It is fine to panic in the test :)

daniel-noland: > How does the macro prevent the panic? You do have to parse the string into a valid prefix at one point or another?

The `#[cfg(test)]` macro means that the panic implementation simply does not exist in production code.  It is fine to panic in the test :)

qmonnet: Aaah OK I didn't understand what you suggested, I thought you meant to implement the trait or function as a macro somehow. Got it now :slightly_smiling_face:

qmonnet:
The trait implementation is now under the `testing` feature gate introduced for the crate by Fredi.

daniel-noland (approved):
Lovely.  Thanks

PR #450, opened 2025-05-09
fix: bump rdma-core

daniel-noland:
The upstream version of rdma-core changes the names of the shipped libraries.

This requires a minor change to our linking in build.rs for dpdk-sys.


qmonnet (approved):
Looks good

PR #456, opened 2025-05-12
feat(mgmt): add grpc server addr option

sergeymatov:


qmonnet (commented):
Would make sense to me to separate the address and port as two separate arguments, no?

sergeymatov:
@qmonnet I would say yes if we have address as mandatory, and port could be a default one (like it's used in for ssh and other cli tools). Having it combined is more native to server listening usage, but it's my subjective opinion.

qmonnet (approved):
OK fine, let's go with just one address then

dataplane/src/main.rs
```diff
@@ -59,11 +59,12 @@ fn main() {
...
-    let grpc_address = "[::1]:50051".parse().expect("Bad grpc address");
-
-    if let Err(e) = start_mgmt(grpc_address) {
+    // Get the gRPC address from command line args
+    let grpc_addr = args.get_grpc_address();
+    info!("Starting gRPC server on address: {}", grpc_addr);
+    if let Err(e) = start_mgmt(grpc_addr) {
         error!("Failed to start management service: {e}");
-        std::process::exit(0);
+        std::process::exit(0); // TODO - should we exit with an error code?
```
daniel-noland, line 67: personally I would call `panic!("Failed to start management service: {e}")` here

dataplane/src/main.rs
```diff
@@ -80,7 +81,7 @@ fn main() {
         }
         other => {
             error!("Unknown driver '{other}'. Aborting...");
-            std::process::exit(0);
+            std::process::exit(0); // TODO - same as above
```
daniel-noland, line 84: again, prefer to call `error!` and then `panic!`

dataplane/src/args.rs
```diff
@@ -30,7 +31,16 @@ pub(crate) struct CmdArgs {
     driver: Option<String>,
     #[arg(long, value_name = "name of kernel interface")]
     interface: Vec<String>,
+
+    // gRPC server address
+    #[arg(
+        long,
+        value_name = "gRPC server address",
+        default_value = "[::1]:50051"
```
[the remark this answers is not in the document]
Fredi-raspall: I believe this is just a temporary default. @sergeymatov is looking into using Unix sockets as transport, in which case we'll be on the safe side.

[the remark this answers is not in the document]
daniel-noland: Ah, that is much better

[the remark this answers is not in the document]
sergeymatov: This is sane default, stands as equivalent of `localhost` or `127.0.0.1` just in ipv6 signature

dataplane/src/main.rs
sergeymatov: switched to `panic!`

dataplane/src/main.rs
sergeymatov: added `panic!`

dataplane/src/args.rs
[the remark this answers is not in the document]
Fredi-raspall: True, that's ::1 (not ::0)

Fredi-raspall (approved):
Looks good to me @sergeymatov , thanks!

dataplane/src/main.rs
```diff
@@ -61,11 +61,12 @@ fn main() {
     /* parse cmd line args */
     let args = CmdArgs::parse();
 
-    let grpc_address = "[::1]:50051".parse().expect("Bad grpc address");
-
-    if let Err(e) = start_mgmt(grpc_address) {
+    // Get the gRPC address from command line args
+    let grpc_addr = args.get_grpc_address();
+    info!("Starting gRPC server on address: {}", grpc_addr);
```
Fredi-raspall, line 66: @sergeymatov I would remove this log, since the gRPC server is actually not starting at this point but later and there is already a log for that.

Issue #180, opened 2025-01-27
NAT: Support stateless NAT44


qmonnet:
I think we'll do stateful first, but my understanding is that we also need support for stateless NAT44 at some point.

mvachhar:
I'm ok with this, though isn't stateful nat more complex than stateless nat44?  Perhaps it is better to get something simple out first, then do the more complex thing?  Or does the design of the stateful nat affect how we do stateless NAT?

qmonnet:
My understanding is that stateless is easier than stateful. In stateless mode we can skip most (all?) of what the NAT session manager needs to do, we also don't need to check whether an IP/port/protocol tuple is free given that we use a translation algo that is not supposed to produce collisions.

I thought we wanted to roll out stateful immediately, although I can't remember where I get this impression from. If that is not the case then I'm happy to start with stateless.

mvachhar:
First priority is to get some meaningful functionality, IMHO.  We can then
get more complex functionality rolled out.  But yes, I suspect stateful NAT
will be the more common use case.  But stateless NAT seems so much easier,
that I feel it is better to just do it first.

qmonnet:
Copy that 🙂

qmonnet:
Mostly addressed with #322. We miss the VNI lookup for it to be functional (and then there will also be a lot of follow-up work). VNI will most likely be passed to the NAT module as metadata.

qmonnet:
See #470

qmonnet:
Fixed with #599. It needs some improvements and refinements, but the initial version should now be working.

PR #470, opened 2025-05-14
New static NAT implementation

qmonnet:
This is a new implementation for static NAT.

The major changes for this new implementation are:

- Use the actual user API objects to build the NAT tables, instead of the peering interfaces and other objects that we considered using in past representations.

- Match addresses from a list of prefixes with exclusion prefixes to a similar list of prefixes with exclusion prefixes (rather than doing a simple prefix-to-prefix match, which would not correspond to the current API).

- Restructure the lookups for NAT rules: use two successive tables for source NAT lookup instead of retrieving and "following" a peering to figure out what VPC's table should be checked.

- Construction of the NAT rules have been moved to the mgmt crate, rather than living next to the NAT processing logics in the dataplane crate.

The code changes a lot since the previous implementation. As a consequence, the PR doesn't split very well into incremental changes; instead I'm re-adding new definitions as separate submodules, then removing most of the legacy code, and then at last I update the core of the logics in dataplane's nat/mod.rs, based on the new definitions introduced in the early commits.

After this Pull Request, the main follow-up items are:

- #445: Objects validation
- VNI lookup
- Calling the functions to build the table as part of the configuration build-up in the mgmt package
- Protecting the tables against concurrent access, likely with left-right


mvachhar:
Does this close any issues?  If so we should mention them in the PR so it auto closes them.

qmonnet:
I know, but this PR alone doesn't close an issue. The related one is #180, but we'll need the follow-ups listed above to have the complete feature working.

mgmt/src/models/internal/nat/table_extend.rs
```diff
@@ -0,0 +1,242 @@
...
+mod tests {
+    use super::*;
+    use crate::models::internal::nat::tables::NatTables;
+    use iptrie::{Ipv4Prefix, Ipv6Prefix};
+    use net::vxlan::Vni;
+    use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+    use std::str::FromStr;
+
+    #[test]
+    fn test_fabric() {
```
mvachhar, line 195: Fuzz testing would be really nice to validate the correctness of some of these optimizations.

mvachhar (approved):

mgmt/src/models/internal/nat/table_extend.rs
qmonnet: I suspect this optimisation will go away, or change, at some point. But other than that, I agree. I just haven't found the time to get into how fuzzing works, yet. I'll look into it at some point.

PR #478, opened 2025-05-15
feat(mgmt): no import routemap in peeringless VPC

Fredi-raspall:
When building the configuration of FRR, do not add an import route- map if the VPC has no peerings. In principle, it should not harm. However, such a route-map is useless and we speculate it may confuse frr-reload.


Fredi-raspall:
> Looks good!
> 
> Should there be a comment above the check to remember why we skip the addition if there's no peering?

I don't think it's needed.

PR #482, opened 2025-05-15
docs: Remove outdated design-docs

qmonnet:
The docs are, in good part, outdated; we're not using or maintaining them at the moment. Let's remove the design-docs directory, and related mentions in justfile, .gitignore, workflows.


qmonnet:
Will need the docs workflow requirement to be unset on the repo before we merge

daniel-noland (approved):
I have to agree with this PR.  No docs are better than wrong docs, and we just didn't put any energy into documentation for this project.

It makes me a little sad to be honest.  To be blunt, I think this represents a lost opportunity and a culture fail.  That said, I could have tried harder to keep this up to date and didn't.

Truthfully, I am now disinclined to write documentation going forward as I don't believe that it will be valued :/

Still, the only way forward is forward.

qmonnet:
> It makes me a little sad to be honest. To be blunt, I think this represents a lost opportunity and a culture fail. That said, I could have tried harder to keep this up to date and didn't.

I don't see it the same way. My impression is that we tried to introduce this documentation very early, probably too early in the process. Is there such a thing as “too early”? I don't know, but we've created the docs at a stage where the code, and the overall design itself, were still moving a lot, and where we've been having strong incentives to make progress on the code base itself, not so much on the accompanying design docs. We should maybe have considered the docs we had more in terms of “initial reflection material”, and not so much as “a description of the moving software parts”. I don't know.
 
> Truthfully, I am now disinclined to write documentation going forward as I don't believe that it will be valued :/

Cheer up! :slightly_smiling_face:. I think it was good to have these docs initially. We'll need to get back to it at some point. We'll need user-facing docs of course, but we'll also need a description of the architecture so we can keep track ourselves, and onboard new contributors to the project. Doesn't mean it has to be you only, and hopefully this will be at a time when we can afford more cycles keeping the docs up-to-date.

> Still, the only way forward is forward.

The alternative is to keep the current version of the docs in the repo with the hope that we can spend some time updating it in the near future. I'm not opposed to that, I opened the PR as much as a way to trigger a discussion about our docs as a clean-up operation.

PR #489, opened 2025-05-16
feat(mgmt): ensure name consistency

Fredi-raspall:
Make sure that there is no inconsistency on name cases by ensuring upper case by construction.

### DIFF
--- a/mgmt/src/frr/renderer/prefixlist.rs
+++ b/mgmt/src/frr/renderer/prefixlist.rs
@@ -50,7 +50,7 @@ impl Render for PrefixList {
     type Output = ConfigBuilder;
     fn render(&self, _: &Self::Context) -> ConfigBuilder {
         let mut config = ConfigBuilder::new();
-        let pfx = format!("ip prefix-list {}", self.name.to_uppercase());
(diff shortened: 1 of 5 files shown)


qmonnet (commented):
I'm not sure I understand the motivation for the second commit. As I understand it, this means that:

- When config passes validation, we log the validated VPC table and peering table, which is fine
- When config is invalid, we don't long (the requested versions of) these tables anymore.

Aren't we losing information in the logs in case of invalid config? What was the issue with displaying the requested config and then telling whether it was valid or not?

Fredi-raspall:
> I'm not sure I understand the motivation for the second commit. As I understand it, this means that:
> 
> * When config passes validation, we log the validated VPC table and peering table, which is fine
> * When config is invalid, we don't long (the requested versions of) these tables anymore.
> 
> Aren't we losing information in the logs in case of invalid config? What was the issue with displaying the requested config and then telling whether it was valid or not?

I don't think we were logging it yet if it was wrong. I just moved the display of the peering table (pretty formatted) because since we empty it, it was showing no peerings because we were logging the contents after having cleared it.

qmonnet (approved):
OK makes sense in that case, thanks for the clarification!

PR #490, opened 2025-05-16
Pr/fredi/grpc move

Fredi-raspall:
Small reorg to keep main small as that will help rebases, pushing functionality outside of it.


Fredi-raspall:
@sergeymatov I am not sure if we could get rid of one of GrpcAddress or ServerAddress.
No big deal, but if we built ServerAddress directly from the cmd line args, we would get rid of one type.

sergeymatov (approved):

qmonnet (approved):

PR #493, opened 2025-05-19
style(cli): Fix Clippy report

qmonnet:
Clippy from the nightly toolchain complains that we can group the two conditions into one, in a check in the completion code for the CLI. Let's make Clippy happy.

Link: https://github.com/githedgehog/dataplane/actions/runs/15113369097/job/42477859103?pr=492


Fredi-raspall:
@qmonnet , listening to clippy complaints is recommended. However, I wonder if we should just silent these in particular.
There are cases where, definitely, collapsing if's is fine. In some others, I don't think it make sense and makes the code harder to read and maintain. I wonder if, some of the changes you propose could be avoided with just one `#![allow(clippy::collapsible_if)]` at the top, as compared to cluttering the whole code with the annotations.

qmonnet:
The macros all over the code are temporary; we should be able to get rid of them as soon as we switch to Rust 1.88 (current beta). I figured that having the macros in the code would help find the code to look at, at that time.

I thought it would be OK to just punt the issue for this time: Whether we chose to collapse the `if`s or to ignore the warning more globally can be decided on a case-by-case basis, once we have support for the `if let ...` chains.

mvachhar:
I take it back, I won't mark this as don't merge, I'll wait for @Fredi-raspall to click approve/merge when he is ready.

Fredi-raspall (approved):

Issue #491, opened 2025-05-16
Fix bump.yml workflow


qmonnet:
`bump.yml` has been failing for a couple days already.

The workflow fails to update the `ahash` dependency, currently pointing to our own fork.

Apparently, there's an `--offline` option that is passed somewhere when we run `cargo upgrade`, and that is the cause for this failure; but I haven't found where it comes from yet.

qmonnet:
We also need to fix the workflow with regards to the addition of the CLI as a separate binary, with some dependencies not handled in the toplevel Cargo.toml file:

```console
$ cargo upgrade --incompatible=allow --dry-run 2>/dev/null
name     old req compatible latest new req
====     ======= ========== ====== =======
netdev   0.34.0  0.34.1     0.34.1 0.34.1 
smallvec 1.14.0  1.15.0     1.15.0 1.15.0 
name old req compatible latest new req
==== ======= ========== ====== =======
log  ^0.4.25 0.4.27     0.4.27 ^0.4.27
```

We get two files processed. I couldn't find a combination of options to prevent that. We should either move the dep declarations for the CLI to the toplevel Cargo.toml, or add some more post-processing to the output from the dry run.

qmonnet:
I think `cargo upgrade` calls `cargo update --offline` to pull the Git dependencies: https://github.com/killercup/cargo-edit/blob/42e15faea5e99b630df5d77fc4f4b4ce70ed0fec/src/bin/upgrade/upgrade.rs#L549 but I'm not sure why, or why it fails, or why I can't reproduce locally.

qmonnet:
Local reproducer:

```console
$ docker run --rm -it ubuntu:24.04
root@5d16f982a0ae:/# sudo apt update
root@5d16f982a0ae:/# apt install --yes git rustup
root@5d16f982a0ae:/# rustup default stable
root@5d16f982a0ae:/# git clone https://github.com/githedgehog/dataplane.git
root@5d16f982a0ae:/# cd dataplane/
root@5d16f982a0ae:/dataplane# cargo update --offline ahash
error: Unable to update https://github.com/githedgehog/aHash?branch=pr%2Fdaniel-noland%2Fbump-zero-copy

Caused by:
  can't checkout from 'https://github.com/githedgehog/aHash': you are in the offline mode (--offline)
```

qmonnet:
https://github.com/killercup/cargo-edit/issues/939

qmonnet:
[...]

PR #495, opened 2025-05-20
ci: Reorder steps in bump.yml workflow to fix upgrades from Git

qmonnet:
We have an issue in the bump.yml workflow. It fails to update recursive dependencies for the project when these dependencies are passed as links to Git repositories (for example, a branch name and an URL to a GitHub repository).

I'm not sure what happens exactly, but there seems to be an issue with the `cargo upgrade` command: it expects some local index to be populated when running the upgrade for the Git-based dependency, and calls `cargo update` under the hood with the `--offline` option, but this command fails.

We've opened a GitHub issue for the upstream `cargo-edit` project (providing `cargo upgrade`), but in the meantime, let's re-order the commands we run: if we do the `cargo update` first, this should populate whatever index `cargo upgrade` seems to rely on when we try to bump the dependencies in Cargo.toml.

See https://github.com/githedgehog/dataplane/issues/491 for more context.

Fixes: #491

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -107,6 +107,13 @@ jobs:
           git config user.email '<41898282+github-actions[bot]@users.noreply.github.com>'
           BASE="$(git rev-parse HEAD)"
 
+          # Run "cargo update"
+          cargo update
+          if ! git diff --quiet; then
+              git add Cargo.lock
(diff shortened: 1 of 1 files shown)


qmonnet:
This will (hopefully :crossed_fingers:) fix the bump.yml workflow, but doesn't touch the rest of the code. I'm merging.

PR #497, opened 2025-05-20
Relax configuration constraints in gRPC

Fredi-raspall:
Admit incomplete configurations that do not have underlay or overlay so that empty configs can be pushed to wipe out the system. Except for syntax, config validation should be moved out of the gRPC handlers.


qmonnet:
dev.yml workflow with nightly toolchain is expected to fail until we merge #493 or some equivalent fix.

PR #498, opened 2025-05-20
fix: temporary fix to bgp neigh update-source

Fredi-raspall:
**Temporary** fix to bgp neigh update-source

The fix hard-codes the name of the loopback interface and should be replaced by a proper method that uses the information coming from the gRPC config model.

See: https://github.com/githedgehog/gateway-proto/compare/pr/fredi/augment-proto-bgp-update-source?expand=1


Fredi-raspall:
@sergeymatov I am assigning this to you as well as the implementation of the update source in agent.
+cc @Frostman

qmonnet (approved):
Looks OK, thanks

Fredi-raspall:
Marking this as no-merge since the work-around will not work with the current config in Fabric.

Fredi-raspall:
I am closing this because the good fix is in https://github.com/githedgehog/dataplane/pull/528

PR #500, opened 2025-05-20
ci: Fix bump.yml workflow for CLI

qmonnet:
We recently adjusted our bump.yml workflow to account for upgrades in the CLI repository (cli), but forgot to "git add" the related Cargo.toml file, making the workflow fail when toplevel Cargo files are not modified (because there's nothing to commit - the change is in cli/Cargo.toml). Let's add it.

Fixes: e3522a0e664d (ci: Fix generation of list of packages to update)


qmonnet:
This time, I'm hopeful this will fix the bump.yml workflow for real. No impact on the codebase outside of the bump.yml workflow file, I'm merging this straight away.

PR #506, opened 2025-05-21
minor additions vpc manager integration

Fredi-raspall:

PR #507, opened 2025-05-21
Fix gRPC over unix

Fredi-raspall:
Note this targets https://github.com/githedgehog/dataplane/pull/506


mgmt/Cargo.toml
```diff
@@ -17,6 +17,7 @@ bolero = ["dep:bolero", "interface-manager/bolero", "id/bolero", "net/arbitrary"
 [dependencies]
 # internal
 gateway_config = { workspace = true }
+hyper = { version = "0.14", features = ["server", "http1", "tcp", "stream"] }
```
qmonnet, line 20: Please set the version in the toplevel Cargo.toml, as for the other dependencies

qmonnet (changes requested):
(see my comment above)

mgmt/src/processor/proc.rs
```diff
@@ -327,28 +377,38 @@ async fn start_grpc_server_unix(
...
         Err(e) => {
             error!("Failed to bind UNIX socket: {}", e);
             return Err(e);
         }
     };
 
     // Set socket permissions if needed
-    match std::fs::set_permissions(socket_path, std::fs::Permissions::from_mode(0o660)) {
-        Ok(_) => debug!("Socket permissions set to 0660"),
+    match std::fs::set_permissions(socket_path, std::fs::Permissions::from_mode(0o666)) {
```
qmonnet, line 392: Why does it need to be 0666 and not 0660?

qmonnet (approved):
Thanks!

qmonnet:
I think the failure for dev.yml with nightly is because this is base on an older branch, which doesn't include the fix yet. Looks unrelated to this PR.

sergeymatov (commented):
Interesting. So the tokio does not provide an accepter for Unix socket stream for GRPC? Seems quite obvious that it's should catching events

mgmt/src/processor/proc.rs
Fredi-raspall: For testing, user may belong to "other". I don't think this poses  a security threat atm. We can revisit later...

Fredi-raspall:
> Interesting. So the tokio does not provide an accepter for Unix socket stream for GRPC? Seems quite obvious that it's should catching events

I don't know. For sure the prior was not working because nobody was calling accept()

PR #509, opened 2025-05-21
Fix Vxlan code + small additions

Fredi-raspall:


Fredi-raspall:
Marking as dont-merge until I double check that these are the good fixes and @daniel-noland approves.

qmonnet (approved):
Changes look OK to me

net/src/headers.rs
```diff
@@ -319,6 +319,11 @@ impl Headers {
         Headers::default()
     }
 
+    /// Add / Replace Ethernet header
+    pub fn set_eth(&mut self, eth: Eth) {
```
daniel-noland, line 323: technically this method is unsafe in the sense that you can set an ethernet header with an ethertype which doesn't match the lower layers of the frame.  Will think about a good way to fix this.

net/src/ipv4/mod.rs
```diff
@@ -257,6 +267,42 @@ impl Ipv4 {
...
+
+    /// Set the length _of the payload_ of the ipv4 packet.
+    ///
+    /// This method will adjust the total length of the header to account for options and the length
+    /// of this header.
+    ///
+    /// This method _will not_ update the checksum of the header.
+    /// # Errors
+    ///    This method returns [`Ipv4LengthError`] if the value is too big
+    pub fn set_payload_len(&mut self, payload_len: u16) -> Result<(), Ipv4LengthError> {
```
daniel-noland, line 297: again, technically an unsafe method.  Invariants pushed to the caller.

net/src/packet/mod.rs
```diff
@@ -74,6 +74,11 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
         &self.payload
     }
 
+    /// Add / Replace Ethernet header
+    pub fn set_eth(&mut self, eth: Eth) {
```
daniel-noland, line 78: technically unsafe

net/src/packet/mod.rs
```diff
@@ -133,10 +138,7 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
     /// ```
     pub fn vxlan_decap(&mut self) -> Option<Result<Vxlan, ParseError<EthError>>> {
         match self.headers.try_vxlan() {
-            None => {
```
daniel-noland, line 136: why are we removing the log here?

net/src/packet/mod.rs
```diff
@@ -161,7 +163,11 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
 
     /// Encapsulate the packet in the supplied [`Vxlan`] [`Headers`]
     ///
-    /// The supplied [`Headers`] will be validated to ensure they form a VXLAN header.
+    /// * The supplied [`Headers`] will be validated to ensure they form a VXLAN header.
+    /// * If the supplied headers describe an IPv4 encapsulation, then the IPv4 checksum will be
+    ///   updated.
+    /// * The IPv4 / IPv6 headers will be updated to correctly describe the length of the packet.
+    /// *
```
daniel-noland, line 170: stray `*`

net/src/packet/mod.rs
```diff
@@ -311,4 +320,18 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
     pub fn get_meta_mut(&mut self) -> &mut PacketMeta {
         &mut self.meta
     }
+
+    /// Wraps a packet in an `Option` depending on the metadata:
+    /// If [`Packet`] is to be dropped, returns `None`. Else, `Some`.
+    pub fn enforce(self) -> Option<Self> {
```
daniel-noland, line 326: I'm not sure I follow the logic for the naming here.  Why `enforce`?

net/src/headers.rs
daniel-noland: Having thought about it some more, I don't think we can fix this safety issue without requiring the user to go through a strongly typed builder API on every packet manipulation.

The actual fix here should be to make the API contract of the packet / headers types free form and then create a validated type wrapper which does not allow mutations without consuming the structure back into the mutable form.

That will prevent us from sending (or accepting) total nonsense while still giving us the latitude to actually manipulate packets.

That said, wherever possible, we should still forbid creating nonsense packets; moving errors closer to their actual cause point is just good practice anyway.

Fredi-raspall: I would say it's the caller's responsibility to set the ether-type. I mean, IMO, the net crate should provide the machinery to build packets, but can't be foreseeing all possible combinations and, if it ought to, it should probably be after the fact maybe when deparsing(), or via some validation method to be called explicitly?

net/src/ipv4/mod.rs
Fredi-raspall: Yes, this is probably a distinct case. Here consistency matters a bit more. There's distinct ways we could enforce correctness. Either with validation or by "encapsulation"; i.e. somewhat emulate a stack by wrapping layers. But this requires a significant refactor and APIs? E.g. you could have something like `packet.eth_encap( Params )`, with packet being anything above L2 (ARP, IPv4, IPv6). This would get the packet, automatically determine ethertype and use params to set src/dst macs. Similarly you could have something like  `packet.ip_encap( Params )` (where packet would be assumed to contain L4 stuff maybe) and the params contain the fields that would make sense for the user to set. Protocol, header-length, packet length, and so on could be set automatically. So you could do things like:
```
udp_packet.ip_encap( l3-params ).eth_encap( l2-params )
```

net/src/packet/mod.rs
Fredi-raspall: We call vxlan_decap for every packet destined to the gateway, but the gateway may receive packets that are not vxlan (control-plane, icmp, etc..). So the prior log would show for every packet sent to it.

net/src/packet/mod.rs
Fredi-raspall: How do you want to proceed? You want me to annotate all those methods as unsafe?

net/src/packet/mod.rs
Fredi-raspall: When we iterate over a burst of packets, we pervasively use `filter_map` which requires returning `Some(packet) `or `None` to indicate if the packet should be returned or not. The routing stages (as currently defined) annotate in the packet metadata what to do next: 1) no annotation, 2) some drop condition (there are plenty) 3) whether the packet can be delivered etc. These annotations happen inside the pipeline functions. The enforce is just some sugar syntax to, based on those, determine if Some(packet) / None is to be returned. So, most stages end up calling packet.enforce(). This allows hiding all that logic to the stages and allowing to change it without needing to change them. That's the explanation. Given this purpose, do you have suggestions for the naming? You're the native speaker here!

Fredi-raspall:
> I think we need to consider some API issues here, but my only actual hard concern is the naming of the `enforce` method. Given that this is easily refactored later, I'm ok with approving this.
> 
> That said, we should talk through my thoughts on a validated packet type at some point

I can easily change the naming for `enforce` if we find a better name. Not that I like it either....

PR #512, opened 2025-05-23
Remove direct dependency on 'hyper' crate

qmonnet:
This all started with yet another breakage of the build.yml workflow. The workflow complained that bumping dependencies failed: it tried to bump the `hyper` crate from version 0.14 to 1.6, and failed, because version 1 of the crate introduced breaking changes. In particular, the `hyper::server` submodule is gone, meaning that we could no longer pull the `Accept` trait.

As it turns out, we do not need the Accept trait to be implemented for the UnixAcceptor to use it with the 'tonic' crate, we only need the `Stream` trait (which we have).

Refactor `poll_next()` and `poll_accept()`: given that the former would only wrap around the latter, inline the code from `poll_accept()` into `poll_next()`, and drop the implementation of the `Accept` trait.

After that, we can simply drop the direct dependency on the `hyper` crate.

Link: https://hyper.rs/guides/1/upgrading/
Link: https://docs.rs/tonic/latest/tonic/transport/server/struct.Router.html#method.serve_with_incoming

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -180,8 +180,8 @@ dependencies = [
  "axum-core",
  "bytes",
  "futures-util",
- "http 1.3.1",
- "http-body 1.0.1",
+ "http",
+ "http-body",
  "http-body-util",
  "itoa",
  "matchit",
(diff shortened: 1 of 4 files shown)


Fredi-raspall (approved):
Looks good to me @qmonnet , thanks!
I tested it and it works.

PR #513, opened 2025-05-23
Update IP checksum when decrementing TTL

Fredi-raspall:
Targets https://github.com/githedgehog/dataplane/pull/509

### DIFF
--- a/net/src/ipv4/mod.rs
+++ b/net/src/ipv4/mod.rs
@@ -201,6 +201,7 @@ impl Ipv4 {
             return Err(TtlAlreadyZero);
         }
         self.0.time_to_live -= 1;
+        self.update_checksum();
         Ok(())
     }
 


qmonnet (changes requested):
Why update the checksum when decrementing the TTL, but not for the other operations touching the header? Shouldn't it be the responsibility of the caller to update the checksum after calling `decrement_ttl()`?

Another thing: it would likely be more efficient to adjust the checksum incrementally, just accounting for the change in the TTL, rather than having `etherparse` recompute the whole checksum (https://www.rfc-editor.org/rfc/rfc1624).

mvachhar:
> Why update the checksum when decrementing the TTL, but not for the other operations touching the header? Shouldn't it be the responsibility of the caller to update the checksum after calling `decrement_ttl()`?
> 
> Another thing: it would likely be more efficient to adjust the checksum incrementally, just accounting for the change in the TTL, rather than having `etherparse` recompute the whole checksum (https://www.rfc-editor.org/rfc/rfc1624).

I suggest we merge as is and create 2 new issues, one for incremental checksum update (which may not matter due to HW offload) and another to fix who the caller is for checksum update.

qmonnet (approved):
OK then

PR #523, opened 2025-05-24
Set bridge mac

daniel-noland:
This PR sets the bridge (and vtep + vrf) MAC to a consistent value, as well as adding infrastructure to allow specifying the VTEP ip more logically.

Doing so requires a bit of plumbing, but not a lot of new logic.

### DIFF
--- a/mgmt/src/models/internal/routing/evpn.rs
+++ b/mgmt/src/models/internal/routing/evpn.rs
@@ -3,16 +3,21 @@
 
 //! Dataplane configuration model: EVPN
 
-use net::eth::mac::Mac;
-use std::net::IpAddr;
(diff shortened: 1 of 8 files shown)


Fredi-raspall:
I've tried the PR and it has the issues with the blank config. Otherwise it works good!
I've pushed a couple of commits where we move things around without changing the code but such that we can apply empty configs and where the validation happens earlier. @daniel-noland , please have a look and modify and squash as you see fit.

We should sync on the configuration API wrt to the vtep, as discussed. The inference from the loopback is not convincing to me (unless we couple it with the bgp config). Else, we should just require vtep interfaces. IMO, the latter is preferable  as it is very explicit  and would allow us to set the VTEP mac. cc @sergeymatov

PR #532, opened 2025-05-28
Some Cleanup of the gRPC converter code in mgmt

mvachhar:
Started cleanup of the mgmt converter code.

I've added clippy checks and fixed issues, but not in a way that will change the interface.
I've removed all hard-coded constants
I've moved code to use Rust pattern matching instead of length checks + indexing for known-length vectors

There is still much more to do, but this is a good start at getting rid of some of the bigger issues.


Fredi-raspall (approved):
LGTM @mvachhar !

PR #537, opened 2025-05-30
fix: fix toolchain linking issue

Fredi-raspall:
Forcing to use compile-env/bin/rustc does not seem to suffice to fix the dynamic linking issues of the toolchain, when invoked via system cargo. Work around by instead invoking the cargo included in the compile environment.

### DIFF
--- a/justfile
+++ b/justfile
@@ -9,6 +9,9 @@ set dotenv-required := true
 set dotenv-path := "."
 set dotenv-filename := "./scripts/rust.env"
 
+# set cargo binary. Just will just fail ;-) if the compile-env does not exist. That's actually good
+cargo_binary := "compile-env/bin/cargo"
+
 # enable to debug just recipes
 
 debug_justfile := "false"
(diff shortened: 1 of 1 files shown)


Fredi-raspall:
I'm aware that this is quick and hacky, but it does the job. @daniel-noland, I am assigning this to you to add any extra sanities you see fit.

qmonnet (approved):
Seems to work for me just as well after the patch, although I'll let Daniel take a look.

It means we need additional parameters to try alternate cargo versions, for example if I want to run locally with nightly:

```console
$ cargo --version                        
cargo 1.89.0-nightly (64a124607 2025-05-30)

$ just cargo --version                   
cargo 1.86.0 (adf9b6ad1 2025-02-28)

$ just cargo_binary=cargo cargo --version
cargo 1.89.0-nightly (64a124607 2025-05-30)
```

(We could always add a `lcargo` `just` target to use local cargo if it becomes annoying, in the future.)

daniel-noland:
closing in favor of (already merged) #557

PR #538, opened 2025-06-01
Fix bugs, extend tests, and refactor gRPC converter

mvachhar:
This PR does the following:

1. It has fuzz testing now for the converter
2. It has bug fixes/extensions for the problems the fuzzer found
3. The converter implementation has been moved from a single file to a directory
4. Updates to gateway-proto 0.2.3

Left to do in future PRs

1.  Move implementation to separate files
2. Remove `convert_*` functions and just implement in the `TryFrom` interfaces

### DIFF
--- a/mgmt/src/grpc/converter/old_impl.rs
+++ b/mgmt/src/grpc/converter/old_impl.rs
@@ -1,10 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
-#![deny(clippy::all, clippy::pedantic)]
-#![allow(clippy::missing_errors_doc)]
-#![allow(clippy::must_use_candidate)] // Do not want to remove pub methods yet
-
 use net::eth::mac::Mac;
 use net::vlan::Vid;
 use std::net::{IpAddr, Ipv4Addr};
--- a/mgmt/src/grpc/converter/old_impl.rs
+++ b/mgmt/src/grpc/converter/old_impl.rs
@@ -730,7 +730,13 @@ pub fn convert_interface_to_grpc(
 ) -> Result<gateway_config::Interface, String> {
     // Get IP address safely
     //let ipaddr = get_primary_address(interface)?;
-    let interface_addresses = interface_prefixes_to_strings(interface);
+    let interface_addresses = match &interface.iftype {
+        InterfaceType::Ethernet(_) | InterfaceType::Vlan(_) | InterfaceType::Loopback => {
+            interface_prefixes_to_strings(interface)
+        }
+        InterfaceType::Vtep(vtep) => vec![format!("{}/32", vtep.local.to_string())],
+        _ => vec![],
+    };
 
     // Convert interface type
     let if_type = match &interface.iftype {
--- a/mgmt/src/grpc/converter/old_impl.rs
+++ b/mgmt/src/grpc/converter/old_impl.rs
@@ -278,16 +278,14 @@ pub fn convert_interface_to_interface_config(
     // Convert interface type
     let grpc_if_type = gateway_config::config::IfType::try_from(iface.r#type)
         .map_err(|_| format!("Invalid interface type: {}", iface.r#type))?;
+    let mac = match &iface.macaddr {
+        Some(mac) => {
+            Some(Mac::try_from(mac.as_str()).map_err(|_| format!("Invalid MAC address: {mac}"))?)
+        }
+        None => None,
+    };
     let iftype = match grpc_if_type {
-        gateway_config::config::IfType::Ethernet => InterfaceType::Ethernet(IfEthConfig {
-            mac: match &iface.macaddr {
-                Some(mac) => Some(
-                    Mac::try_from(mac.as_str())
-                        .map_err(|_| format!("Invalid MAC address: {mac}"))?,
-                ),
-                None => None,
-            },
-        }),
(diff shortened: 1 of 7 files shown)


mgmt/src/models/external/overlay/vpc.rs
```diff
@@ -34,21 +34,22 @@ pub struct Peering {
...
+        if value.len() != ID_LEN {
             return Err(ConfigError::BadVpcId(value.to_owned()));
         }
         if !value.chars().all(|c| c.is_ascii_alphanumeric()) {
             return Err(ConfigError::BadVpcId(value.to_owned()));
         }
-        let chars: Vec<char> = value.chars().collect();
-        Ok(VpcId::new(chars[0], chars[1], chars[2], chars[3], chars[4]))
+        let mut chars = value.chars().take(ID_LEN);
+        Ok(VpcId::new([(); 5].map(|i| chars.next().unwrap_or('0'))))
```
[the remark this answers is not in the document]
mvachhar: What do others think here, I don't know if it is a good idea to panic the entire dataplane.

mgmt/src/grpc/converter/old_impl.rs
```diff
@@ -816,6 +812,10 @@ pub fn convert_bgp_neighbor_to_grpc(
     })
 }
 
+fn has_redistribute(redistribute: &[Redistribute], protocol: &Protocol) -> bool {
+    redistribute.iter().any(|r| r.protocol == *protocol)
+}
+
```
qmonnet, line 886: Nit: This is added in the first commit, but should it be added later? It's not used yet in the first commit.

mgmt/src/grpc/converter/old_impl.rs
```diff
@@ -306,8 +306,16 @@ pub fn convert_interface_to_interface_config(
...
-                    .map_err(|_| format!("Invalid local IP address for VTEP: {addr}")),
+                [addr] => {
+                    let (ip, mask) = parse_cidr(addr)
+                        .map_err(|_| format!("Invalid local IP address for VTEP: {addr}"))?;
+                    if mask == 32 {
+                        Ok(IpAddr::from_str(&ip)
+                            .map_err(|_| format!("Invalid local IP address for VTEP: {addr}")))
+                    } else {
+                        Err("VTEP interface requires a /32 IP address".to_string())
+                    }?
```
qmonnet, line 317: Should we use a checked type for the VTEP IP?

qmonnet (commented):
Something I don't really understand, why rename the file to mgmt/src/grpc/converter/old_impl.rs (which suggests it's legacy and will be deleted soon, not extended) but add new code to it in later commits?

mgmt/src/grpc/converter/old_impl.rs
mvachhar: Yep this should be moved, will do.

mgmt/src/grpc/converter/old_impl.rs
mvachhar: We should be using Prefix to parse this, but I didn't address this here, I just replicated the pattern that was already there.

I have one more cleanup PR that I want to do where I can address this.

mvachhar:
> Something I don't really understand, why rename the file to mgmt/src/grpc/converter/old_impl.rs (which suggests it's legacy and will be deleted soon, not extended) but add new code to it in later commits?

So, I intended to refactor this first then put in the bug fixes, but I decided, after moving the files, that it would be better to fix the bugs first, then split this into separate files.

Hence the name, old_impl.rs

I wanted the file split to also be in this PR but it took me longer than I wanted to get here an thought it better to merge the fixes and tested version.

I also am not sure it is worth refactoring to fix this particular issue at this time since I will split this into separate files and delete this file in the next day or so.

I hope this makes sense.

mgmt/src/models/external/overlay/vpc.rs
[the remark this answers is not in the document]
mvachhar: Ok, I'll use an unwrap where I can use the unreachable! Macro so it is clear that we expect this to be unreachable, with a comment as to why.

mgmt/src/grpc/converter/old_impl.rs
mvachhar: Fixed.

mgmt/src/models/external/overlay/vpc.rs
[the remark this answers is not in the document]
mvachhar: Done.

mgmt/src/grpc/converter/old_impl.rs
mvachhar: Nevermind, I pushed a fix to use Prefix to parse the addr in this case.

mvachhar: Oh, and as for a checked type for /32, I don't think that is generally needed.  The check is here because we are abusing the interface IP addresses for the VTEP interface.

We should fix this, but it isn't worth the time really because of the changes to the gRPC and how that percolates through the code.

qmonnet (approved):
> I also am not sure it is worth refactoring to fix this particular issue at this time since I will split this into separate files and delete this file in the next day or so.
> 
> I hope this makes sense.

It does, it might be worth mentioning in the PR in the first place :). I'm OK with that, as long as we do split the files soon and don't remain with a file called `old_impl.rs` for too long.

Looks good, thanks!

mgmt/src/grpc/converter/old_impl.rs
qmonnet: OK

PR #555, opened 2025-06-05
fix(routing): Move mac_address dependency to toplevel Cargo.toml

qmonnet:
Move the declaration of the mac_address dependency (with its version and features) from routing/Cargo.toml to the toplevel Cargo.toml, for consistency.

This issue was found because it broke the bump.yml GitHub workflow, given that it's the routing/Cargo.toml file that gets updated, and we don't "git add" it which results in no added changes to commit.

Link: https://github.com/githedgehog/dataplane/actions/runs/15457829853/job/43513283734
Fixes: 31a7771f2b9d


daniel-noland (approved):
LGTM.  Thanks

PR #558, opened 2025-06-05
chore(mgmt): Bump gateway-proto to 0.2.4

mvachhar:
Do not merge until [gateway-proto#29](https://github.com/githedgehog/gateway-proto/pull/29) is merged.

Now that dataplane uses Rust 1.87.0 we can use
gateway-proto 0.2.4 code that requires Rust 1.87.0


daniel-noland (approved):
Yup.

PR #559, opened 2025-06-06
Split gRPC converter into multiple files and remove convert_* functions

mvachhar:
This PR:

* Splits the gRPC converter into multiple files
* Removes almost all of the `convert_*` functions in favor of `TryFrom`
  * The one convert that remains is the top level function because it adds defaults.  We probably want to remove this at some point but I want to limit the blast radius of these changes.
  * The VRF commit still uses the first vrf as default if there is no VRF labeled default - this is broken IMHO but I don't want to break things for what is currently minor benefit.  Plus the fuzzer in gateway-proto will have to be fixed to rememdy this issue.
* Minor cleanups, elimination of constants, use of checked types, etc.
   * These are rolled into the file move commits since the problems were observed during the move
 * Adds parsing for `InterfaceAddress` types
   * `Prefix` is not the same as `InterfaceAddress` even though both look like `<ip addr>/<mask>` as Prefix zeros the low order bits, `InterfaceAddress` does not.  It may be the case that we should error parsing a `Prefix` that does not already have the low order bits as 0 and error if the `InterfaceAddress` does have the low order bits as 0 (except for /31 and /127 masks) but I do not do that now as to limit the blast radius these changes.
   * This distinction also revealed a bug in the value generators used for fuzzing so I will fix that next.


mgmt/src/models/internal/interfaces/interface.rs
```diff
@@ -87,6 +90,46 @@ impl InterfaceAddress {
...
+pub enum InterfaceAddressParseError {
+    #[error("Invalid interface address format: {0}")]
+    InvalidFormat(String),
+    #[error("Invalid IP address interface address: {0}")]
+    InvalidIpAddress(String),
+    #[error("Invalid mask length in interface address: {0}")]
+    InvalidMaskLength(String),
+}
+
+impl FromStr for InterfaceAddress {
```
Fredi-raspall, line 103: Also, why is this needed for?

Issue #181, opened 2025-01-27
NAT session manager: Create new flows into flow manager


qmonnet:
Once a new NAT session has been created, we need to ask the flow manager to create the resulting flows.

TO DO: Figure out how to create relevant flows in the case of stateless NAT, for which we don't record NAT session. I don't know how expressive are the flows we can generate (for stateless NAT, we don't have a 1:1 mapping for an existing connection, but instead we want to use a single algorithm to map automatically any packet).

Issue #175, opened 2025-01-27
NAT session manager: Synchronise NAT sessions upon creation


qmonnet:
New NAT sessions need to be synchronised with other Gateway instances once they're created.

Issue #174, opened 2025-01-27
NAT session manager: Create new session, assign IP/port number


qmonnet:
Implement creation of new NAT sessions.

For dynamic NAT, we need to use an efficient algorithm for IP/port allocation. From meeting's discussion, “it needs to be O(1), or maybe O(n) if n is small, but we don't want to walk all the hash table with NAT sessions to find whether a port is free”.

mvachhar:
O(log n) would also suffice for bounded and small n.

qmonnet:
Addressed with #697

qmonnet:
See follow-ups in #788

PR #560, opened 2025-06-06
Skeleton for stateful NAT

qmonnet:
Introduce a very simple skeleton for stateful NAT. There's nothing functional yet, this is just a very early skeleton for the logic. Committing this part first allows us to later create Pull Requests to focus on the more complex components (in particular, IP/port allocation and session management).

It turns out we don't share that much code with the stateless NAT portion of the pipeline stage; we may split stateless and stateful NAT into distinct pipeline stages in the future, this is yet to be determined (and doesn't matter much at this stage).

Follow-ups:
- #174
- #181
- #175


dataplane/src/nat/stateful/allocator.rs
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use std::collections::BTreeSet;
+use std::net::IpAddr;
+
+#[derive(Debug, Clone)]
+pub struct NatPool {
+    ips: BTreeSet<IpAddr>,
```
daniel-noland, line 9: why use a `BTreeSet` instead of a `HashSet`?

dataplane/src/nat/stateful/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+pub enum StatefulNatError {
+    #[error("other error")]
+    Other,
+}
+
+#[derive(Debug, Clone)]
+struct NatState {}
+
+#[derive(Debug, Clone)]
+struct NatTuple {
```
daniel-noland, line 24: Again, I would make this type generic over a sealed Ip trait

dataplane/src/nat/stateful/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+    src_ip: IpAddr,
+    dst_ip: IpAddr,
+    protocol: u8,
+    vni: Vni,
+}
+
+impl NatState {
+    fn new(net: &Net, pool: &allocator::NatPool) -> Self {
```
daniel-noland, line 32: Are we expecting to clone `net` and `pool`?  I would expect these to be transferred in as owned

dataplane/src/nat/stateful/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+        Self {}
+    }
+}
+
+impl Nat {
+    fn hash_tuple(net: &Net, vni: Vni) -> NatTuple {
```
daniel-noland, line 38: I'm not sure I understand the intent of this function.  Is it a lookup?  If so, what if it misses?  Don't we need to return `Option`

daniel-noland (changes requested):
I think this is a really strong start.  I propose we have a meeting at some point and sketch out some refinements before moving too much farther.

dataplane/src/nat/stateful/mod.rs
qmonnet: > Is it a lookup? If so, what if it misses? Don't we need to return `Option`

Yes, that's right. Another detail in the skeleton that can be adjusted when the proper implementation lands (but it looks better if it's in order now, so I'll fix it)

qmonnet: > I'm not sure I understand the intent of this function.

Looks like I moved this bit and forgot to rename it. This function is used to extract the tuple from the packet, so yes it's a lookup. `hash_` is not a good prefix here, I'll rename it, thank you for noticing :slightly_smiling_face:

dataplane/src/nat/stateful/mod.rs
qmonnet: Wait I replied too fast, why should they be passed as owned? They won't be part of the generated `NatState`, they'll just be used read-only to create the new state. So by reference should be the way to go? Otherwise, I've got to clone them before passing them to the `NatState::new()` method.

dataplane/src/nat/stateful.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl NatState {
+    fn new(net: &Net, pool: &NatPool) -> Self {
+        Self {}
+    }
+}
+
+#[derive(Debug, Clone)]
+struct NatPool {}
+
```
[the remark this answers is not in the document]
qmonnet: I added two commits, one making the allocator a trait, the other making the session manager (sessions table) a trait. Please take a look @daniel-noland, to validate that this is what you had in mind.

dataplane/src/nat/stateful/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[derive(Debug, Clone)]
+struct NatState {}
+
+#[derive(Debug, Clone)]
+struct NatTuple {
+    src_ip: IpAddr,
+    dst_ip: IpAddr,
+    protocol: u8,
+    vni: Vni,
```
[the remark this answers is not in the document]
qmonnet: As discussed yesterday, we now use a `VrfId` in the tuple.

For now, the VRF ID will be obtained from the source VNI.

dataplane/src/nat/stateful/allocator.rs
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use std::collections::BTreeSet;
+use std::net::IpAddr;
+
+#[derive(Debug, Clone)]
+pub struct NatPool {
```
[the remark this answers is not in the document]
qmonnet: I had to look it up, but yes, I think I understand how this makes sense. The last commit of the PR now implements that. Thanks!

PR #569, opened 2025-06-07
CI rework

daniel-noland:
This PR experienced quite a bit of mission creep to be honest.

See the commit log for more details, but the high level is

1. add support for test coverage
2. add support for codecov
3. rework CI workflows to optimize for "fast failure"
4. rename a badly named feature (i.e., net/arbitrary)


mvachhar (changes requested):
Please put the instructions on how to run this in the readme.md file?

scripts/rust.env
```diff
@@ -14,7 +14,7 @@ COVERAGE="-C instrument-coverage"
 FUZZ_SANITIZERS="-Z sanitizer=address,leak"
 OPTIMIZE_OFF="${DEBUG_ASSERTIONS_ON} ${OVERFLOW_CHECK_ON}"
 OPTIMIZE_ON="-C opt-level=3 ${LTO} ${DEBUG_ASSERTIONS_OFF} ${OVERFLOW_CHECK_OFF}"
-OPTIMIZE_FUZZ="-C opt-level=3 ${DEBUG_ASSERTIONS_ON} ${OVERFLOW_CHECK_ON} ${FUZZ_SANITIZERS} ${COVERAGE}"
+OPTIMIZE_FUZZ="-C opt-level=3 ${LTO} ${DEBUG_ASSERTIONS_ON} ${OVERFLOW_CHECK_ON}"
```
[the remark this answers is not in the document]
daniel-noland: About 25 seconds.  But I think I have it worked out to be functionally zero now that the pipeline is reworked.

That is, the longest run is release-sterile because

1. it also uses LTO
2. it pushes the actual container build

As a result the time for the other steps is only relevant in so far as it is important to make sure we fail fast if we are going to fail at all.

I tried to order the steps so that the most plausible failure points are as early as possible in the pipeline.

daniel-noland:
> It looks to me like the coverage tests bring the CI workflow from 3 min to 5 min (code coverage tests themselves took 1'47", uploading results from existing tests add just a few seconds). Is it worth it? Should we consider running the coverage tests in a separate job (would Codecov be able to understand it's from the same PR)?

Now that I have factored sterile.yml into dev.yml the longest pull is about the same as sterile.yml always was.  I don't see any path which optimizes away the need to build the release artifact, push the release artifact image, or run the release tests.

The only thing we could do to optimize this is to make a different job just for pushing the release artifact container, but that requires yet another compile-env pull.

Now that the pipeline is optimized for rapid failure, I don't think the added complexity is worth while.

scripts/rust.env
qmonnet, line 15: Just double-checking, is it expected that the `${FUZZ_SANITIZERS}` are not longer part of `OPTIMIZE_FUZZ`?

.github/workflows/dev.yml
```diff
@@ -285,14 +285,6 @@ jobs:
...
             cargo doc
 
-      - id: "build_each_commit"
-        continue-on-error: true
-        name: "build each commit"
-        if: "${{ github.event.pull_request || github.event.merge_group }}"
-        run: |
-          BASE="${{ github.event.pull_request.base.sha || github.event.merge_group.base.sha }}"
-          just build-sweep "${BASE}"
-
```
qmonnet: OK to remove, but I don't fully understand the explanations from commit description. How is this broken, and who is it not actionable?

.github/workflows/dev.yml
```diff
@@ -49,9 +49,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust:
-          - version: "stable"
-            optional: false
```
qmonnet, line 80: So this means we have no intent to get back to supporting several compiler versions? I thought this was fixable? I suppose it would require shipping several versions of the compiler in the compile-env, would that be an issue apart from the increased size?

.github/workflows/dev.yml
```diff
@@ -49,9 +49,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        profile:
+          - name: "debug"
+            sterile: ""
```
qmonnet, line 54: What's the point for the “not-sterile” environment? What are we expecting to catch with this run?

justfile
```diff
@@ -17,6 +17,13 @@ dpdk_sys_commit := shell("source ./scripts/dpdk-sys.env && echo $DPDK_SYS_COMMIT
 [private]
 _just_debuggable_ := if debug_justfile == "true" { "set -x" } else { "" }
 
+# Set to FUZZ to run the full fuzzer in the fuzz recipe
+_test_type := "DEFAULT"
+
+# comma delimited list of sanitizers to use with bolero
+
+sanitizers := "address,leak"
```
qmonnet: Nit: Remove blank line between comment and assignment

```suggestion
# comma delimited list of sanitizers to use with bolero
sanitizers := "address,leak"
```

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+If you wish to study coverage data locally, you can run 
+
+```shell
+just coverage
+cd ./target/nextest/coverage/html
+python3 -m http.server
+```
+
+And then open a web-browser to [http://localhost:8000](http://localhost:8000) to view coverage data.
```
qmonnet, line 43:
```suggestion
And then open a web-browser to <http://localhost:8000> to view coverage data.
```

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Fuzz testing (bolero)
+
+The dataplane project makes fairly extensive use of [fuzz testing](https://en.wikipedia.org/wiki/Fuzzing).
+We use the [bolero] crate for our fuzz tests.
+
+Running the test suite via `just cargo test` or `just cargo nextest run` will run the fuzz tests, but it does so under significant limitations.
```
qmonnet, line 50: Why are they “significant”? :slightly_smiling_face:

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. The tests (even the fuzz tests) are only run briefly.
+2. Coverage information and sanitizers are not enabled.
+3. A full fuzzing engine is not set up, so evolutionary feedback is not provided when the tests are run this way,
```
qmonnet, line 54: Nit: punctuation

```suggestion
3. A full fuzzing engine is not set up, so evolutionary feedback is not provided when the tests are run this way.
```

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+Using [libfuzzer](https://llvm.org/docs/LibFuzzer.html) or [afl](https://github.com/AFLplusplus/AFLplusplus) can remove these limitations in exchange for adding additional limits.
```
qmonnet, line 56: “additional limits” sounds a bit confusing, if they're additional it's unclear whether the initial ones are entirely lifted or not.

“... can lift these limitations, but comes with new restrictions instead:”?

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+1. The tests are run one at a time and can run for much longer (actually, they run as long as you want them to).
+2. Coverage information and sanitizers (as well as overflow checks and debug asserts) are enabled.
+3. Evolutionary feedback!
```
qmonnet, line 60: Are these limitations? :thinking:

testing.md
qmonnet, line 54: Note: I'd use unordered lists when there's no particular reason to order the items.

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+```shell
+just _test_type=FUZZ fuzz vxlan::test::mutation_of_header_preserves_contract -T 15min
+```
+
+> [!NOTE]
+> The fuzz tests are run with full optimizations and extensive debugging information, so expect a fairly long compile time.
+
+> [!NOTE]
+> Sorry about the need to pass in `_test_type=FUZZ`.  It requires significant refactoring of the `justfile` to avoid this requirement.
```
qmonnet, line 86: Not sure what this does in the docs. That's the way it is, no need to apologise! :slightly_smiling_face:

testing.md
```diff
@@ -0,0 +1,93 @@
...
+
+Fortunately, [nextest] ships with recent versions of the compile-env, so assuming you have already followed the instructions in the [README.md](./README.md), you should be able to run
+
+```shell
+just cargo nextest run
+```
+
+even if you have not installed [nextest].
+
+:warning: [nextest profiles] are not the same thing as [cargo profiles].
```
qmonnet, line 15: Use a “warning” admonition instead?

justfile
```diff
@@ -31,10 +31,6 @@ profile := "debug"
 [private]
 _container_repo := "ghcr.io/githedgehog/dataplane"
 
-# the rust channel to use (choose stable, beta, or nightly)
-
-rust := "stable"
```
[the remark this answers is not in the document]
daniel-noland: Not really.  The mechanism by which we would test with more toolchains would just change.

The reasonable way to do it would be to split the sysroot from the compile-env.  Then ship multiple compile-envs and run each in a branch.  It wouldn't be too bad

testing.md
daniel-noland: Fair enough

testing.md
daniel-noland: Ok.

testing.md
daniel-noland: They are the relative advantages.  The limitations are listed next.  Can reword tho

.github/workflows/dev.yml
daniel-noland: The non-sterile environment is there to ensure that the developer's workflow works normally.  I don't want to break normal IDE integrations.  The sterile environment is far too sparse to actually develop anything in.

.github/workflows/dev.yml
```diff
@@ -39,33 +39,7 @@ permissions:
...
-          persist-credentials: "false"
-          fetch-depth: "0"
-      - name: "Check code changes"
-        uses: "dorny/paths-filter@v3"
-        id: "changes"
-        with:
-          filters: |
-            devfiles:
-              - '!(README.md|LICENSE|.gitignore|.github/**)'
-              - '.github/workflows/dev.yml'
```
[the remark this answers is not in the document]
qmonnet: The point at the time was mostly for docs, indeed. But if it's true we're not changing these files often, we do change other CI files a bit more for example (looking at you, bump.yml). And I find it a bit frustrating to run the full CI suite in that case, but :shrug:

.github/workflows/dev.yml
qmonnet: OK thank you

daniel-noland:
> It looks to me like the coverage tests bring the CI workflow from 3 min to 5 min (code coverage tests themselves took 1'47", uploading results from existing tests add just a few seconds). Is it worth it? Should we consider running the coverage tests in a separate job (would Codecov be able to understand it's from the same PR)?

This should be addressed now.

The total build is longer, but it is also much more parallel this way.

I think it will make effective development more efficient in that if we are going to fail, we are likely to fail notably faster (and with more actionable data)

justfile
```diff
@@ -337,6 +344,17 @@ fake-nix refake="":
...
+# Run the full fuzzer / property-checker on a bolero test. Args are forwarded to bolero
+[script]
+list-fuzz-tests *args: (cargo "bolero" "list" ("--sanitizer=" + sanitizers) "--build-std" "--profile=fuzz" args)
+
+# Run the full fuzzer / property-checker on a bolero test. Args are forwarded to bolero
+fuzz test timeout="-T 60sec" *args="--engine=libfuzzer --engine-args='-max_len=65536'": (compile-env "just" "_test_type=FUZZ" "cargo" "bolero" "test" test "--build-std" "--profile=fuzz" ("--sanitizer=" + sanitizers) timeout args)
+
+# Run the full fuzzer / property-checker on a bolero test with the AFL fuzzer
+[script]
+fuzz-afl test: (fuzz test "" "--engine=afl" "--engine-args=-mnone")
```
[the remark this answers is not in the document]
mvachhar: Is it American Fuzzing Lop?  It is a strategy of picking values for the fuzz tester, if I am not mistaken.

scripts/test-runner.sh
```diff
@@ -116,13 +116,46 @@ for test_suite in "${WRAPPED_TEST_SUITES[@]}"; do
...
+  # project (LLVM is a C++ codebase).
+  # Unfortunately, the combination of bolero's RUSTFLAGS and the nix fenix rust overlay _do not_ set the rpath for
+  # libstdc++.so.6.
+  # As a result, a naive attempt to execute the test binary in the compile-env will result in a file not found error
+  # when the dynamic linker is unable to find libstdc++.so.6.
+  # Fortunately, this is relatively easy to fix; we need to patch the test binary to make sure it resolves to the
+  # exact libstdc++.so.6 file which it was liked against.
+  # If the compile-env is correct, then `/lib/libstdc++.so.6` will always be a symlink to the `/nix` store which
+  # contains the correct dynamic library.
+  patchelf --replace-needed libstdc++.so.6 "$(readlink -e /lib/libstdc++.so.6)" "${test_exe}"
```
mvachhar, line 138: This is really gross.  Is there really no better way with LD_LIBRARY_PATH or editing the rpath in the binary or something?  Perhaps we never care about running this binary elsewhere?

mvachhar (approved):

scripts/test-runner.sh
daniel-noland: Yeah, we don't care about running this binary elsewhere.  It is a one and done sterile build.

I just can't think of any other way to fix this.

LD_LIBRARY_PATH _should_ work, but I can't seem to make it work.  I wish I knew why

PR #586, opened 2025-06-13
Use gRPC configured mac address and ip address on VTEP

mvachhar:


mgmt/src/models/internal/interfaces/interface.rs
```diff
@@ -48,7 +48,7 @@ pub struct IfVtepConfig {
     pub mac: Option<Mac>,
     pub vni: Option<Vni>,
     pub ttl: Option<u8>,
-    pub local: IpAddr,
+    pub local: Ipv4Addr,
```
Fredi-raspall, line 51: I understand the intent of this change but not the strategy we're following here. We may need to support Ipv6 IIRC. Wouldn't it be better then to leave the internal implementation "dual" and simply forbid Ipv6 when processing the config? Otherwise, this change will require us changing the implementation again to support IPv6. Recall that FRR does not support Ipv6 atm but there's some work going on in that direction.

mgmt/src/processor/confbuild.rs
```diff
@@ -204,55 +204,56 @@ fn build_vpc_internal_config(
...
-fn get_loopback_interface_config(
-    external: &ExternalConfig,
-) -> Result<&InterfaceConfig, ConfigError> {
-    external
+struct VtepInfo {
+    ip: UnicastIpv4Addr,
+    mac: SourceMac,
+}
+
+fn get_vtep_info(external: &ExternalConfig) -> Result<VtepInfo, ConfigError> {
```
Fredi-raspall, line 212: Wouldn't it be easier to simply validate the vtep interface in https://github.com/githedgehog/dataplane/blob/main/mgmt/src/models/internal/interfaces/interface.rs#L167

... to avoid needing to iterate over all of the interfaces to do that validation?
The `InterfaceConfig::validate()` is executed once per interface from https://github.com/githedgehog/dataplane/blob/main/mgmt/src/models/external/gwconfig/mod.rs#L34

That is the place where I would add the restriction for Ipv4 only (okay and unicast) so that both the gRPC conversion and the rest of code can be dual version.

Fredi-raspall: Also, the building of the configuration is fallible. However, it logically occurs after having validated it. So, if we want to retain that structure, I'd move the validation there.

mgmt/src/processor/confbuild.rs
```diff
@@ -204,55 +204,56 @@ fn build_vpc_internal_config(
...
-        .ok_or(ConfigError::MissingParameter("loopback interface address"))
+            _ => unreachable!(),
+        },
+        _ => Err(ConfigError::TooManyInstances("Vtep interfaces", 1)),
+    }
 }
 
 /// Build the Vtep config
 fn build_vtep_config(external: &ExternalConfig, internal: &mut InternalConfig) -> ConfigResult {
     // TODO: The vtep configuration should come from the external config, but I am deriving it from
```
Fredi-raspall, line 250: I think these comments could be removed.

Fredi-raspall (changes requested):

mgmt/src/models/internal/interfaces/interface.rs
mvachhar: We should be using the type system to help us avoid errors and we should catch the error as early as possible so that we don't have to constantly recheck.  Since we have to do some checks in gRPC -> mgmt anyway we might as well build the best types possible.

By using the type system in this way, we don't have to reason globally about the code, we can know what has been validated and what hasn't.

Having said this, we are where we are with this right now and fixing this to generate all the right types unravels a ton of code and error checking which is not worthwhile right now.

As for, "we might want IPv6 Vteps", when FRR supports IPv6 underlay we can fix this.  There has been progress on that front, but all the VXLAN stuff uses `in_addr` everywhere in FRR and that has to be unraveled.  It is going to be a while, so I see no issue with IPv4 use here.

THe later section of code in mgmt should perform semantic validation that is beyond what we encode in the types.  That validation function should return an immutable config called `ValidatedConfig` or something like that.  However, all that is just not worth it right now which is why the commit message says this is a tiny step in the that direction.

(For example, I tried to move from `Mac` to `SourceMac` and that cause too many problems to be worth handling right now.

mvachhar: TL;DR, I'm going to suggest we keep this change as is.

mgmt/src/processor/confbuild.rs
mvachhar: Because the interface set is still a list here, I would still have to do the check dynamically since we do not want to count on validate doing the right thing as the code changes.  We either encode the checks we did in the type system or we check dynamically where there is a possible error.  

I am glad to add a check to the validator as well though so that we fail earlier and move these to internal errors.
I will make that fix.

mgmt/src/processor/confbuild.rs
mvachhar: Fixed.

mgmt/src/models/internal/interfaces/interface.rs
Fredi-raspall: > We should be using the type system to help us avoid errors and we should catch the error as early as possible so that we don't have to constantly recheck. Since we have to do some checks in gRPC -> mgmt anyway we might as well build the best types possible.
> 
> By using the type system in this way, we don't have to reason globally about the code, we can know what has been validated and what hasn't.

Agree. Vlan 4097 is invalid. However, here, you're talking about an Ip address which can be ipv4 or ipv6. 
This is a different case. An ipv6 unicast address is perfectly valid for a tunnel. We simply don't want to accept one at this point because of a limitation of the software we build or use (FRR). What I was trying to avoid was needing to refactor back the code when FRR supports it for the sake of efficiency.

mgmt/src/processor/confbuild.rs
Fredi-raspall: ok!

mvachhar: Moved the real checking code into the validator.  Any checks needed for type conversion/extraction in confbuild now return `InternalFailure`

mgmt/src/models/internal/interfaces/interface.rs
mvachhar: I don't think it is that big of a deal to expand this back to IpAddr later when it is supported.  If that support was coming in the next few weeks before release, I'd agree, just leave it, but given that it is going to be months if not a year+ before it is refactored, it is worth changing IMHO.

PR #587, opened 2025-06-13
fix(mgmt): Remove unused reconcile_demo annotation

mvachhar:
The annotation was causing tests to fail on some machines


daniel-noland (approved):
LGTM

Sorry about random issue.  Not sure what is going on yet

PR #589, opened 2025-06-13
checksum updates

daniel-noland:
Rushed PR to add support for checksum updates.

This is missing the type of testing I would normally write.  Instead only smoke tests are included.

### DIFF
--- a/net/src/checksum.rs
+++ b/net/src/checksum.rs
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Traits for checksum calculation and manipulation
+
+use std::fmt::Debug;
+
+/// A trait for checksum calculation and manipulation.
+///
+/// This trait is used to calculate and manipulate checksums in various headers.
+pub trait Checksum {
+    /// The payload type for the header.
+    ///
+    /// This is used to calculate the checksum.
+    type Payload<'a>: ?Sized
+    where
+        Self: 'a;
+    /// The checksum type.
+    ///
+    /// This is used to represent the checksum value.
+    type Checksum: Eq + Copy + Sized + Debug;
+
+    /// Get the checksum value from the header
+    fn checksum(&self) -> Self::Checksum;
+
+    /// Compute the checksum value from the header and payload
+    fn compute_checksum(&self, payload: &Self::Payload<'_>) -> Self::Checksum;
+
+    /// Set the checksum value in the header.
+    ///
+    /// # Safety
+    ///
+    /// The validity of the checksum is not checked.
+    ///
+    /// The contract of the [`Checksum`] trait _does not_ require that the implementation of this
+    /// function be free of panics.
+    /// "Normal" input should never cause this trait to panic, but truly exceptional conditions
+    /// such as wildly out of the ordinary MTU values (e.g., 2^32) may not be possible to handle
+    /// without a panic.
+    fn set_checksum(&mut self, checksum: Self::Checksum) -> &mut Self;
+
+    /// Validate the checksum value in the header.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`ChecksumError`] if the checksum is invalid.
+    fn validate_checksum(
+        &self,
+        payload: &Self::Payload<'_>,
+    ) -> Result<Self::Checksum, ChecksumError<Self>> {
+        let expected = self.compute_checksum(payload);
+        let actual = self.checksum();
+        if expected == actual {
+            Ok(expected)
+        } else {
+            Err(ChecksumError { expected, actual })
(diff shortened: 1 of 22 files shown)


net/src/udp/mod.rs
```diff
@@ -128,6 +120,22 @@ impl Udp {
         self.0.length = length.get();
         self
     }
+
+    fn compute_checksum_ipv4(&self, net: &Ipv4, payload: impl AsRef<[u8]>) -> UdpChecksum {
+        #[allow(clippy::expect_used)] // payload greater than 2^16 bytes should be excluded by DPDK
+        self.0
+            .calc_checksum_ipv4(&net.0, payload.as_ref())
+            .expect("unreasonable payload")
```
mvachhar, line 128: Can we be more specific than "unreasonable payload".

Also, is there a way not to panic here?  Maybe create an issue to not panic in these cases and just drop the packet?

net/src/packet/mod.rs
```diff
@@ -202,43 +206,48 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
...
-
         #[allow(clippy::cast_possible_truncation)] // checked
         let udp_len = NonZero::new(len as u16).unwrap_or_else(|| unreachable!());
         #[allow(unsafe_code)] // sound usage due to length check
         unsafe {
             udp.set_length(udp_len);
         }
 
+        // the VXLAN spec says that the checksum SHOULD be zero
+        udp.set_checksum(UdpChecksum::ZERO);
```
mvachhar, line 217: Really, I wonder why?  That seems busted, but oh well.

net/src/headers.rs
```diff
@@ -1372,4 +1378,290 @@ mod test {
...
+                .build()
+                .unwrap()
+        }
+    }
+
+    fn test_checksum(mut headers: Headers) {
+        match &headers.transport {
+            None => {}
+            Some(Transport::Udp(transport)) => {
+                let net = headers.net.clone().unwrap();
```
mvachhar, line 1545: Can we have some tests with a non-zero payload perhaps?  Even if you look at wireshark, find the correct check sum and just hardcode the constant in at least some test cases?

mvachhar (changes requested):
We can create an issue for the panic test, but we should have a test with an actual non-empty payload for each packet type and compare to a reference checksum.  That is a good stop gap until we have a full fuzzer and better checksum oracle.

net/src/ipv4/mod.rs
```diff
@@ -201,7 +204,6 @@ impl Ipv4 {
             return Err(TtlAlreadyZero);
         }
         self.0.time_to_live -= 1;
```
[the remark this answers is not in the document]
mvachhar: Checksums are globally recomputed upon packet serialization, this is a non-issue.

net/src/checksum.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    /// Set the checksum value in the header.
+    ///
+    /// # Safety
+    ///
+    /// The validity of the checksum is not checked.
+    ///
+    /// The contract of the [`Checksum`] trait _does not_ require that the implementation of this
+    /// function be free of panics.
+    fn set_checksum(&mut self, checksum: Self::Checksum) -> &mut Self;
```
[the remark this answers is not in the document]
daniel-noland: description updated.

The root issue is that some checksum calculations are payload length limited.  We could make this return a reult type but that is very costly in this case.

net/src/icmp6/checksum.rs
```diff
@@ -0,0 +1,121 @@
...
+    ///
+    /// # Safety
+    ///
+    /// It is undefined behavior of the payload length to be greater than the maximum allowed by
+    /// the [`Icmp6`] header (2^32 bytes - 8 bytes for the header itself).
+    ///
+    /// This is a very unlikely event, and as such is only checked if debug assertions are enabled.
+    #[must_use]
+    pub fn new(src: Ipv6Addr, dst: Ipv6Addr, payload: &'a [u8]) -> Self {
+        debug_assert!(
```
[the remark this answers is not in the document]
daniel-noland: The payload is the exact slice we hand it in this case, and the length _can't_ be incorrect unless DPDK is broken (so far as I know).  Given that DPDK's mbufs are shorter than the payload length which triggers this error condition, that seems like a safe bet.

Basically, if we were going to hit this error, there is a 99.99% chance we would segfault first anyway.

net/src/packet/mod.rs
daniel-noland: https://datatracker.ietf.org/doc/html/rfc7348#section-5

net/src/headers.rs
daniel-noland: Done.

PR #598, opened 2025-06-17
Annotate source vni in packets

Fredi-raspall:


dataplane/src/packet_processor/ipforward.rs
```diff
@@ -124,6 +124,7 @@ impl IpForwarder {
                             return;
                         };
                         debug!("Next fib/vrf is {next_vrf}");
+                        packet.get_meta_mut().src_vni = Some(vni);
```
qmonnet, line 127: Maybe log the VNI for debug as well?

Fredi-raspall: It's logged in line 119, above.

qmonnet: Sorry, I'd missed it. Thanks!

PR #599, opened 2025-06-17
feat(nat): Use actual source-VPC VNI with NAT

qmonnet:
Based on #598

Based on Fredi's recent work to add the VNI from the source VPC to packet metadata, here's an update to the NAT code to read the VNI from metadata (rather than the hardcoded placeholder that we had).

We also update unit tests accordingly.


qmonnet:
> My only concern is that one we have already discussed: using VNI where it would make more sense to me to use VRF id.

That discussion happened in the context of stateful NAT. I understand we probably want to use the same for stateless NAT as well, but the current implementation (for stateless) has been using the VNI from the beginning, and changing it to the VRF id is beyond the scope of the current PR (we'd need to change the stateless NAT tables to use the VRF id as a key rather than the VNI when we build the tables). I'll merge this, but create a ticket to track it.

Thanks for the review!

PR #602, opened 2025-06-17
Miscellaneous CLI fixes for dataplane

mvachhar:
* Address all copilot comments from #601 
* Fix path bug in `test-runner.sh` so that `just cargo run -p dataplane` works as expected
* Fix help messages for cli arguments.

### DIFF
--- a/dataplane/src/args.rs
+++ b/dataplane/src/args.rs
@@ -42,7 +42,7 @@ pub(crate) struct CmdArgs {
         long,
         value_name = "ADDRESS",
         default_value = "[::1]:50051",
-        help = "IP Address and port or UNIX socket path to listen for managment connections"
+        help = "IP Address and port or UNIX socket path to listen for management connections"
     )]
     grpc_address: String,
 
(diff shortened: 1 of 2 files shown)


mvachhar:
> Looks good to me.
> 
> I recommend $() instead of backticks in bash files, but in this case it doesn't really matter

Fixed.

dataplane/src/args.rs
```diff
@@ -52,21 +52,24 @@ pub(crate) struct CmdArgs {
...
+        help = "Unix socket to listen for dataplane cli connections",
         default_value = DEFAULT_DP_UX_PATH_CLI
     )]
     cli_sock_path: String,
 
     #[arg(
         long,
-        value_name = "Unix socket to connect to FRR agent that controls FRR configuration reload",
+        value_name = "FRR Agent Unix socket path",
+        help = "Unix socket to connect to FRR agent that controls FRR configuration reload",
```
Fredi-raspall, line 72: I would be more precise on the descriptions, esp. in the last. The frr-agent socket is the remote address of the frr-agent, unlike the first two, which are addresses of the dataplane process. I'd also squash this commit with the first?

Fredi-raspall: So, I'd say:
   help = "Address of unix socket of frr-agent" (or where frr-agent listens for connections)

Fredi-raspall (changes requested):
Looks good to me @mvachhar I'd just address the minor comment if possible.

PR #603, opened 2025-06-17
Add MTU support into gRPC interface

mvachhar:


daniel-noland (approved):
I have mixed feelings about mtu being a u32 but I see your point.

LGTM

PR #610, opened 2025-06-18
Integration mgmt-routing - part-1

Fredi-raspall:
This PR contains a bunch of clean-ups and a conservative integration of mgmt (config) and the routing crate.
"Conservative" because I've tried to keep both as-is and not yet unify types in order to avoid circular dependencies between the crates. This is part-1 because further work needs to be done around the interfaces outside of this PR and I don't want to defer it longer.

Leaving aside the code clean-ups and refactors, the process is as follows.
* The routing entity (with its dedicated thread) exposes a config API over the control channel (that already existed).
* When mgmt has a new config to apply, it builds a request with the required "router" configuration. This includes VRFs, Interfaces and vtep. The routing thread then applies the config using the existing code and replies. So, for the mgmt processor, the router behaves as the frr-agent or the vpc-manager: an external entity to request some reconfiguration.
* To avoid races with FRR, prior to applying a config, the mgmt processor "locks it" by not allowing it to receive messages from FRR. It then sends the config, which the routing thread 1) validates 2) applies 3) verifies (verify means that we check if the internal state conforms to the config after applying it) and when the process completes, the routing entity is unlocked to process messages from FRR.
* When a config is such that a VRF should be deleted, we don't delete it immediately but mark it as status = deleting, to give FRR a chance of deleting all routes. When no route remains, the vrf is removed.
* Auto-learn of VRF is removed but not the interface ones.

Closes https://github.com/githedgehog/dataplane/issues/394

### DIFF
--- a/mgmt/src/models/external/overlay/vpc.rs
+++ b/mgmt/src/models/external/overlay/vpc.rs
@@ -116,7 +116,7 @@ impl Vpc {
 pub struct VpcTable {
     vpcs: BTreeMap<String, Vpc>,
     vnis: BTreeSet<Vni>,
-    ids: BTreeSet<VpcId>,
+    ids: BTreeMap<VpcId, String>, // name of vpc
 }
 impl VpcTable {
     /// Create new vpc table
@@ -137,14 +137,14 @@ impl VpcTable {
         if self.vnis.contains(&vpc.vni) {
             return Err(ConfigError::DuplicateVpcVni(vpc.vni.as_u32()));
         }
-        if self.ids.contains(&vpc.id) {
+        if self.ids.contains_key(&vpc.id) {
             return Err(ConfigError::DuplicateVpcId(vpc.id));
         }
         if self.vpcs.contains_key(&vpc.name) {
             return Err(ConfigError::DuplicateVpcName(vpc.name.clone()));
         }
         self.vnis.insert(vpc.vni);
-        self.ids.insert(vpc.id.clone());
+        self.ids.insert(vpc.id.clone(), vpc.name.clone());
         self.vpcs.insert(vpc.name.to_owned(), vpc);
         Ok(())
     }
@@ -152,6 +152,14 @@ impl VpcTable {
     pub fn get_vpc(&self, vpc_name: &str) -> Option<&Vpc> {
         self.vpcs.get(vpc_name)
     }
+    /// Get a [`Vpc`] by [`VpcId`]
+    pub fn get_vpc_by_vpcid(&self, vpcid: VpcId) -> Option<&Vpc> {
+        match self.ids.get(&vpcid) {
+            Some(name) => self.vpcs.get(name),
+            None => None,
+        }
+    }
+
     /// Iterate over [`Vpc`]s in a [`VpcTable`]
     pub fn values(&self) -> impl Iterator<Item = &Vpc> {
         self.vpcs.values()
@@ -169,10 +177,6 @@ impl VpcTable {
     /// Clear set of vnis
     pub fn clear_vnis(&mut self) {
         self.vnis.clear();
-    }
-    /// Clear set of ids
-    pub fn clear_ids(&mut self) {
-        self.ids.clear();
     }
     /// Validate the [`VpcTable`]
     pub fn validate(&self) -> ConfigResult {
--- a/net/src/interface/display.rs
+++ b/net/src/interface/display.rs
@@ -132,3 +132,22 @@ impl Display for MultiIndexInterfaceMap {
         Ok(())
     }
 }
+
+pub struct MultiIndexInterfaceMapView<'a, F>
+where
+    F: Fn(&Interface) -> bool,
+{
+    pub map: &'a MultiIndexInterfaceMap,
+    pub filter: &'a F,
+}
+
+impl<F: for<'a> Fn(&'a Interface) -> bool> Display for MultiIndexInterfaceMapView<'_, F> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        fmt_kernel_interface_heading(f)?;
+        let interfaces = self.map.iter().filter(|(_, iface)| (self.filter)(iface));
+        for (_, iface) in interfaces {
+            iface.fmt(f)?;
+        }
+        Ok(())
+    }
+}
--- a/net/src/interface/display.rs
+++ b/net/src/interface/display.rs
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Display trait implementations
+
+use crate::interface::MultiIndexInterfaceMap;
+use crate::interface::{AdminState, OperationalState};
+use crate::interface::{
+    BridgeProperties, Interface, InterfaceProperties, VrfProperties, VtepProperties,
+};
+use std::fmt::Display;
+use std::string::ToString;
+
+macro_rules! KERNEL_INTERFACE_FMT {
+    () => {
+        " {:>8} {:>8} {:>16} {:>8} {:>20} {:>8} {:>8} {:>8} {:}"
+    };
+}
+fn fmt_kernel_interface_heading(f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    writeln!(
+        f,
+        "{}",
+        format_args!(
+            KERNEL_INTERFACE_FMT!(),
+            "index", "contrl", "name", "mtu", "mac", "Adm", "Oper", "type", "properties"
+        )
+    )
+}
+
+impl Display for AdminState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AdminState::Down => write!(f, "down"),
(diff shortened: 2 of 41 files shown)


routing/src/rib/vrf.rs
```diff
@@ -163,6 +163,10 @@ impl RouterVrfConfig {
         self.vni = Some(vni);
         self
     }
+    pub fn set_vni_opt(mut self, vni: Option<Vni>) -> Self {
```
mvachhar, line 167: Why not just call this `set_vni`?  We have plenty of other similarly named methods that take an Option

routing/src/ctl.rs
```diff
@@ -82,6 +82,20 @@ impl RouterCtlSender {
         reply?;
         Ok(())
     }
+    pub async fn configure(&mut self, config: RouterConfig) -> Result<(), RouterError> {
+        debug!("Requesting router to apply configuration...");
```
mvachhar, line 86: What is the "Router" in this case, FRR, a different part of the dataplane?

mgmt/src/models/external/overlay/vpc.rs
```diff
@@ -116,7 +116,7 @@ impl Vpc {
 pub struct VpcTable {
     vpcs: BTreeMap<String, Vpc>,
     vnis: BTreeSet<Vni>,
-    ids: BTreeSet<VpcId>,
+    ids: BTreeMap<VpcId, String>, // name of vpc
```
mvachhar, line 119: Is a duplicate vpc name ok as long as it has a unique Id?

routing/src/rib/vrf.rs
```diff
@@ -272,7 +272,8 @@ impl Vrf {
     /// Set the status of a [`Vrf`]
     /////////////////////////////////////////////////////////////////////////
     pub fn set_status(&mut self, status: VrfStatus) {
-        if self.status != status {
+        // the default vrf (vrfid = 0) can't be deleted and it's always active
+        if self.status != status && self.vrfid != 0 {
```
mvachhar, line 277: Do we want to log attempts to change the status of the default vrf as that would be a bug?

routing/src/rib/vrf.rs
```diff
@@ -279,6 +280,42 @@ impl Vrf {
...
 
+    /////////////////////////////////////////////////////////////////////////
+    /// Check if a [`Vrf`] needs to be deleted and mark it as such. Only
+    /// [`Vrfs`] in state `Deleting` can be deleted and the default VRF never
+    /// gets to that status.
+    /////////////////////////////////////////////////////////////////////////
+    pub fn check_deletion(&mut self) {
+        if self.status == VrfStatus::Deleting {
+            let one = NonZeroUsize::new(1).unwrap_or_else(|| unreachable!());
+            if self.routesv4.len() == one && self.routesv6.len() == one {
```
mvachhar, line 291: Why do we return NonZeroUsize from len() instead of usize?  Can there never be 0 routes in the data structure?

routing/src/config/vrf.rs
```diff
@@ -135,13 +135,22 @@ impl ReconfigVrfPlan {
     }
 
     #[must_use]
-    fn enforce_deletions(&self, vrftable: &mut VrfTable) -> Result<(), RouterError> {
+    fn enforce_deletions(
+        &self,
+        vrftable: &mut VrfTable,
+        iftw: &mut IfTableWriter,
```
mvachhar, line 141: For VRF deletion, what if we get a config that says to delete a VRF and then a new config that recreates that VRF, could we somehow be left over with stale routes, or can that not happen?

routing/src/display.rs
```diff
@@ -210,6 +210,16 @@ impl Display for VrfStatus {
     }
 }
 
+fn fmt_vrf_oneline(vrf: &Vrf, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    let description = vrf.description.clone().unwrap_or_else(|| "--".to_string());
+    writeln!(
+        f,
+        "\n ━━━━━━━━━\n Vrf: '{}' (id: {}) description: {description}",
```
mvachhar, line 217: This looks like formatting for the cli.  It is fine here since it was kind of already there, but instead of hijacking Display for a core type to get the cli to work, we should have wrapper types that implement display for the CLI specifically.  Can be a bit hard if there is deep nesting of course, but that isn't quite the case here.  In the future make a wrapper type instead of hijacking Display for a particular case of how something should be displayed.

```
struct CoreType {
}

impl Display for CoreType {
 // Implementation should be a nice way to log or otherwise display the type in many different scenarios
}

#[repr(transparent)]
struct CoreTypeForCli(CoreType)

impl Display for CoreTypeForCli { 
  // Very CLI specific display output
}
```

should do the trick in the future.

routing/src/config/interface.rs
```diff
@@ -112,7 +112,10 @@ impl ReconfigInterfacePlan {
         iftw: &mut IfTableWriter,
         vrftable: &VrfTable,
     ) -> Result<(), RouterError> {
-        self.enforce_deletions(iftw)?;
+        if false {
```
mvachhar, line 115: comment to explain this in the code instead of just the commit?

net/src/interface/display.rs
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Display trait implementations
+
+use crate::interface::MultiIndexInterfaceMap;
```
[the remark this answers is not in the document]
Fredi-raspall: Yes, I thought of adding the MTU when I saw it merged. I didn't because the interface work will be done in part-2.

routing/src/cpi.rs
```diff
@@ -96,73 +96,102 @@ fn open_unix_sock(path: &String) -> Result<UnixDatagram, RouterError> {
     Ok(sock)
 }
 
-#[allow(clippy::too_many_lines)]
+pub(crate) const CPSOCK: Token = Token(0);
```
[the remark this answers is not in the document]
Fredi-raspall: Happy to discuss about it! I think the solution is reasonably neat and safe in that the guard will ensure that we "unlock" the router at some point.

routing/src/rib/vrf.rs
Fredi-raspall: Yes. There are currently 3 methods to set the vni. I should probably unify them all.

routing/src/cpi.rs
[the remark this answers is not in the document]
Fredi-raspall: Note that this is not a real lock. I can rename it to avoid confusion.

routing/src/config/vrf.rs
Fredi-raspall: We don't get "delete" or "add" VRF. We get the full set of VRFs "that should be there" and determine if we need to add or remove them ourselves. I don't think we could be left over stale routes, if FRR does the right thing. Let me explain:

* If there is some VRF A at some point, we'll have it and FRR will install routes in it.
* If, later, it is no longer there (because vpc manager pulls it off the kernel), we'll mark it as deleting and give frr a chance to clean it (if needed).
* If, later on, some VRF A shows up again, it's going to be a different VRF, with a different ifindex in the kernel. So, even if the old A was still there (because frr did not empty it), the new "A" will happily coexist with it and we'll have two "VRF A", one in deleting state (until frr is done) and one in active state, into which frr will install routes.

routing/src/config/interface.rs
Fredi-raspall: Ok!

net/src/interface/display.rs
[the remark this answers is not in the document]
Fredi-raspall: Added mtu!

routing/src/rib/vrf.rs
mvachhar: Ok, do you want to do that here, or later?

routing/src/config/vrf.rs
mvachhar: Ahh, ok great.  And this answers my other question about vpcs with different ids and the same name.

mgmt/src/models/external/overlay/vpc.rs
mvachhar: Yep, it is fine per other discussion here.

mvachhar (approved):

routing/src/cpi.rs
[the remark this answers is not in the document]
Fredi-raspall: It is there to prevent us from learning stuff from FRR that we're not ready to digest.
Simple example:
* vpc manager creates a vrf in the kernel. FRR sees it and starts adding routes. However, the routing logic is not prepared to handle those (without auto-learning) because it yet does not know anything about that VRF, since FRR has been faster. 
* The routing logic needs to know about that vrf because it is indexed by ifindex, which is the way how frr refers to it. So, we can't inform the routing about it in advance without creating the vrf in the kernel.
* One solution (the one used) is to "lock" the routing when we mess around in the kernel, so that we don't get to process anything that frr  sends.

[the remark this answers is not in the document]
mvachhar: This is a real lock, but it is fine, I'll approve it.

PR #619, opened 2025-06-20
veth support

Fredi-raspall:


Fredi-raspall:
This code is no longer needed. So, I'm closing the PR....

PR #635, opened 2025-06-25
NAT: Split stateful/stateless, move to own crate

qmonnet:
Please see individual commit descriptions for details.


Fredi-raspall (changes requested):
LGTM @qmonnet!!. There seems to be a small issue with the rust docs. Can you please fix it so that we can merge? Thanks!

PR #636, opened 2025-06-25
host gateway part 1

daniel-noland:
This is part one of the changes needed for host vtep to function.

The first few commits are just cleanup to make followon commits easier to manage and develop.


mgmt/Cargo.toml
```diff
@@ -11,7 +11,7 @@ path = "tests/reconcile.rs"
 required-features = ["bolero"]
 
 [features]
-default = ["bolero"]
```
mvachhar, line 14: Does removing bolero here also mean that fuzz testing isn't in the default `cargo test` run?

mvachhar (changes requested):

mgmt/Cargo.toml
daniel-noland: No.  bolero is also listed as a dev dependency.  It is always used by the tests.

This controls the optionality of bolero for produced artifacts.  Basically, do we want downstream librarires to be able to use the fuzzers from this crate.  The answer to that is "maybe, but only if they are doing a test build."

So it shouldn't be on by default

daniel-noland: No.  bolero is also listed as a dev dependency.  It is always used by the tests.

This controls the optionality of bolero for produced artifacts.  Basically, do we want downstream libraries to be able to use the fuzzers from this crate.  The answer to that is "maybe, but only if they are doing a test build."

So it shouldn't be on by default

interface-manager/Cargo.toml
```diff
@@ -17,6 +17,7 @@ rekon = { workspace = true }
 # external
 bolero = { workspace = true, optional = true, default-features = false, features = ["alloc"] }
 derive_builder = { workspace = true, default-features = false, features = ["default"] }
+futures = { workspace = true, features = ["default"] }
```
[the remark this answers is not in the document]
daniel-noland: It provides the `TryStream` trait which is used a lot by rtnetlink.

`futures` is a likely dependency any time you end up using an async library which is trying to be executor agnostic.  `rtnetlink` works with tokio, async-std, and async-global-executor so it is very much in that camp.

net/src/interface/mtu.rs
```diff
@@ -0,0 +1,148 @@
...
+use std::ops;
+
+/// The MTU of a network interface.
+#[derive(Copy, Clone, Debug, Hash, Ord, PartialOrd, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "u32", into = "u32")]
+#[repr(transparent)]
+pub struct Mtu(NonZero<u32>);
+
+impl Mtu {
+    pub(crate) const MIN_U32: u32 = 1280; // 1280 IPv6 minimum MTU
```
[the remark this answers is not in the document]
mvachhar: In the interest of time, let's leave it as is.  It is crate scope only anyway.

mvachhar (approved):
Ok, looks good enough.

PR #637, opened 2025-06-25
misc routing fixes I (so expect a II)

Fredi-raspall:
Several changes here and there, some important, some not so much.
Important changes:
  * remove the auto-learn mode. We learn interfaces from kernel.
  * fixes ip prefix list configuration.
  * TTL handling: do not decrement for locally-destined traffic (This does not currently matter atm, but avoid ugly log).

Notes:
  * we still rely on FRR to tell us about ip addresses. We can learn them from kernel or config too. Part II should cover this. Right now these are not strictly needed in dataplane, but will soon.


mgmt/src/processor/confbuild/router.rs
```diff
@@ -190,6 +190,13 @@ pub(crate) fn generate_router_config(
...
     generate_router_vtep_config(internal, &mut router_config);
-    generate_router_interfaces_config(internal, kernel_vrfs, &mut router_config)?;
+
+    #[cfg(test)]
+    let gen_intf_cfg = false;
+    #[cfg(not(test))]
+    let gen_intf_cfg = true;
+    if gen_intf_cfg {
+        generate_router_interfaces_config(internal, kernel_vrfs, &mut router_config)?;
+    }
```
qmonnet, line 200: Might deserve a comment in the code?

qmonnet (approved):
Looks good to me, nothing blocking from my side

PR #638, opened 2025-06-25
host gateway part 2

daniel-noland:
This is a very rushed implementation of tc functionality we will not need for some time.

Due to sudden priority changes, I'm submitting it without the type of testing I would normally add.  Testing this properly requires hardware in CI in any case.

My goal with this PR is to get this functionality merged and integrated so that we avoid losing all of this work as the code changes.

My request for the review here is basically that you flag things you think will be in our way if we merge this.  I expect the remainder of this functionality to go unused for a while.


mvachhar (approved):
Looks good.

PR #648, opened 2025-06-28
Stateless NAT: Fix reverse path

qmonnet:


qmonnet:
The PR needs commit re-ordering and some clean-up, but _I Think It Works_ :tm:
I'll get back to it in the evening.

mvachhar (approved):
I think this looks right now, and the tests which are way more complete seem to confirm it.  I will refrain from merging until @Fredi-raspall confirms this can work with routing.

Issue #650, opened 2025-06-29
NAT: Support empty `ips` (respectively `as_range`) with non-empty `not` (respectively `not_as`) prefix lists


qmonnet:

qmonnet:
Copied from https://github.com/githedgehog/dataplane/issues/661, and related to `VpcExpose` overlap verification:

> Another issue arises when an expose object has an empty ips or as_range, but a non-empty corresponding not or not_as. In that case, we don't check overlap correctly because the VpcExpose object does not, initially, contain a root prefix (such as 0.0.0.0/0) on which to “apply” these exclusion prefixes, making it hard to check whether they overlap. Again, this would be easier to address if we had already optimised the expose objects prior to validation. Note that this second issue is currently prevented by the fact we reject configurations with exclusion prefixes only (see https://github.com/githedgehog/dataplane/issues/650), but it stands in the way if we wanted to support such configurations.

Note: I don't remember the exact semantic of an expose with no allowed prefix - we need to clarify that, first.

PR #653, opened 2025-06-30
Stateless NAT: Remove `nots` and `not_as` from NAT tables and processing logics

qmonnet:
Implement a function to collapse exclusion prefixes (`nots`, `not_as`) into allowed prefixes (`ips`, `as_range`) to create a list of smaller fragment prefixes representing the same IP addressing ranges, but without the need to keep track of any exclusion prefixes.

Use this function to get rid of exclusion prefixes altogether in the NAT tables. This allows us to greatly simplify the processing when building NAT tables and when computing the IP address mapping for stateless NAT.

One other consequence is that it allows us to revert e296c013fe969ee57ab852979dd748224601782a (we don't do it in this PR) and address #650.


mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -105,93 +93,138 @@ pub fn add_peering(table: &mut PerVniTable, peering: &Peering) -> Result<(), Tri
...
+        let split_address = address_split(&prefix_covering_exclude);
+
+        let Ok(subprefix_low) = Prefix::try_from((prefix_address, prefix_len + 1)) else {
+            // We can't reach this: prefix_len cannot be the max length for the prefix, because it
+            // means the allowed prefix is a single IP address. If this were the case, the exclusion
+            // prefix would be a single address as well. And if this were the case, the two prefixes
+            // would be equal. This is impossible, because we returned early from the loop in the
+            // calling function if we found that exclude.covers(prefix).
+            warn!("Bug in apply_exclude logics, for subprefix_low");
+            return Err(());
```
daniel-noland, line 189: I think this `warn!` should be an `unreachable!` personally. Programmer error is not a recoverable thing in general

mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -251,4 +284,228 @@ mod tests {
...
+        );
+        assert_eq!(
+            address_split(&"1.0.0.0/31".into()),
+            IpAddr::V4(Ipv4Addr::from_str("1.0.0.1").unwrap())
+        );
+    }
+
+    #[test]
+    #[allow(clippy::too_many_lines)]
+    fn test_collapse_prefix_lists() {
```
daniel-noland, line 314: these tests look fine to me but I really do think you would have an easier time with bolero.  Won't block PR on it tho :)

daniel-noland (approved):
Only a couple small things from me (neither of which block).

I appreciate how much complexity this PR cuts.  Nice work.

That said, I am very out of context on this PR so may want another reviewer

mgmt/src/models/internal/natconfig/table_extend.rs
qmonnet: I was more or less expecting this comment :) But I'm not sure how to do it (in particular, how to validate that the output for non-deterministic inputs will be as expected, without running the same algo we're trying to evaluate). We probably need to talk about it at some point.

mvachhar (changes requested):

mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -194,6 +195,140 @@ fn optimize_peering(peering: &Peering) -> Peering {
...
+                result.remove(prefix);
+                break;
+            }
+
+            // If allowed prefix covers the exclusion prefix, then it means the exclusion prefix
+            // excludes a portion of this allowed prefix. We need to remove the allowed prefix, and
+            // add instead the smaller fragments resulting from the application of the exclusion
+            // prefix.
+            let Ok(mut apply_exclude_result) = apply_exclude(prefix, exclude) else {
+                return Err(());
```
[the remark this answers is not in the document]
qmonnet: What do you mean by “just drop the prefix”? I agree with Daniel in the sense that a programming error seems tricky to “recover from”, here. If we don't process a prefix, then what happens? Does this mean that we keep the `nots` and `not_as` for this prefix?  This will likely cause bugs in the mapping processing, given that we assume we don't have these exclusion prefixes anymore. So I'd just fail the config build indeed.

As for the specific error enum, yes, it's cleaner, I'll do it today.

mgmt/src/models/internal/natconfig/table_extend.rs
mvachhar: Looks like we'll reject the config so we should return an error.

mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -194,6 +195,140 @@ fn optimize_peering(peering: &Peering) -> Peering {
...
+// prefixes.
+fn apply_exclude(prefix: &Prefix, exclude: &Prefix) -> Result<BTreeSet<Prefix>, ()> {
+    let mut result = BTreeSet::new();
+    let mut prefix_covering_exclude = *prefix;
+    let len_diff = exclude.length() - prefix.length();
+
+    for _ in 0..len_diff {
+        let prefix_len = prefix_covering_exclude.length();
+        let prefix_address = prefix_covering_exclude.as_address();
+        let split_address = address_split(&prefix_covering_exclude);
```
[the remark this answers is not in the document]
qmonnet: Addressed now.

mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -109,95 +109,6 @@ pub fn add_peering(table: &mut PerVniTable, peering: &Peering) -> Result<(), Tri
     Ok(())
 }
 
-/// Optimizes a list of prefixes and their exclusions:
-///
-/// - Remove mutually-excluding prefixes/exclusion prefixes pairs
-/// - Collapse prefixes and exclusion prefixes when possible
-fn optimize_expose(
```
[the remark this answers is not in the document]
qmonnet: No bug in `covers` as far as I can tell.

The bug you found with fuzzing was somewhere else in my algo, and should have been addressed.

mvachhar (approved):

Issue #661, opened 2025-07-01
NAT: Fix VpcExpose overlap validation


qmonnet:
There are at least two issues.

First: When we check overlap between `expose` objects, we can have one `expose` with empty `as_range`/`not_as`, and in this case, the “publicly-exposed” IP address will be the one from `ips` (no NAT for this expose). This means that, when checking overlap with other exposes, we should compare the `ips` ranges from this expose with the `as_range` of the other expose objects. Something like:

```diff
diff --git i/mgmt/src/models/external/overlay/vpcpeering.rs w/mgmt/src/models/external/overlay/vpcpeering.rs
index d030b1415881..aed8f1a3e908 100644
--- i/mgmt/src/models/external/overlay/vpcpeering.rs
+++ w/mgmt/src/models/external/overlay/vpcpeering.rs
@@ -157,18 +157,40 @@ impl VpcManifest {
         for (index, expose_left) in self.exposes.iter().enumerate() {
             // Loop over the remaining exposes in the list
             for expose_right in self.exposes.iter().skip(index + 1) {
-                validate_overlapping(
-                    &expose_left.ips,
-                    &expose_left.nots,
-                    &expose_right.ips,
-                    &expose_right.nots,
-                )?;
-                validate_overlapping(
-                    &expose_left.as_range,
-                    &expose_left.not_as,
-                    &expose_right.as_range,
-                    &expose_right.not_as,
-                )?;
+                if expose_left.as_range.is_empty()
+                    && expose_left.not_as.is_empty()
+                    && (!expose_right.ips.is_empty() || !expose_right.nots.is_empty())
+                {
+                    validate_overlapping(
+                        &expose_left.ips,
+                        &expose_left.nots,
+                        &expose_right.ips,
+                        &expose_right.nots,
+                    )?;
+                } else if expose_right.as_range.is_empty()
+                    && expose_right.not_as.is_empty()
+                    && (!expose_left.ips.is_empty() || !expose_left.nots.is_empty())
+                {
+                    validate_overlapping(
+                        &expose_right.ips,
+                        &expose_right.nots,
+                        &expose_left.ips,
+                        &expose_left.nots,
+                    )?;
+                } else {
+                    validate_overlapping(
+                        &expose_left.ips,
+                        &expose_left.nots,
+                        &expose_right.ips,
+                        &expose_right.nots,
+                    )?;
+                    validate_overlapping(
+                        &expose_left.as_range,
+                        &expose_left.not_as,
[...]

PR #662, opened 2025-07-01
Support empty `ips`/`as_range` with non-empty `nots`/`not_as`

qmonnet:
In progress - blocked by #661

### DIFF
--- a/mgmt/src/models/internal/natconfig/table_extend.rs
+++ b/mgmt/src/models/internal/natconfig/table_extend.rs
@@ -138,6 +138,14 @@ fn collapse_prefix_lists(
     excludes: &BTreeSet<Prefix>,
 ) -> Result<BTreeSet<Prefix>, NatPeeringError> {
     let mut result = prefixes.clone();
+
+    if result.is_empty() && !excludes.is_empty() {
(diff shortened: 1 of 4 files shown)


mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -138,6 +138,14 @@ fn collapse_prefix_lists(
     excludes: &BTreeSet<Prefix>,
 ) -> Result<BTreeSet<Prefix>, NatPeeringError> {
     let mut result = prefixes.clone();
+
+    if result.is_empty() && !excludes.is_empty() {
```
mvachhar, line 142: Did we decide if these configs are valid or if we want to reject them?

qmonnet: Right now we reject them.

I need to clear with @Frostman the exact semantics of empty `ips`/`as` and non-empty `nots`/`not_as` to be sure to validate the `VpcExpose` correctly. But it's not at the top of my priority list at the moment.

qmonnet:
Tracked in #779, closing the stale PR for now.

PR #664, opened 2025-07-02
fix(nat): Fix validation for one-direction NAT

qmonnet:
This is similar to commit 7344ee06abbd: when we only NAT on one direction, then one of the two VPCs uses its `ips` prefix list as public `ips` (no NAT), and these are the prefixes we should compare public addresses from `as_range` in other expose objects with.

In particular, we need to do this comparison when validating VpcManifest objects: when checking for prefix overlap, we may have to check the `as_range` blocks of an expose object with the `ips` blocks of another, if the latter expose has no `as_range`.

Note that even in that case, we still check for overlap between the `ips` lists of both expose objects, as we can't afford to have overlap for them (`ips` lists of exposes in a same manifest).

### DIFF
--- a/mgmt/src/models/external/overlay/tests.rs
+++ b/mgmt/src/models/external/overlay/tests.rs
@@ -231,19 +231,54 @@ pub mod test {
             .ip("1.0.1.0/24".into()) // overlaps with expose1.ips
             .as_range("15.0.0.0/16".into())
             .as_range("16.0.0.0/24".into());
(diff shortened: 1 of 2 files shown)


mvachhar (changes requested):
Is there any way that we can build a small set of functions that give us the exposed IPs for one of these peering blocks so that we don't have to keep having this "If this is empty then that is empty then check this type of logic"?

mvachhar (approved):

PR #667, opened 2025-07-04
Stateless NAT: Optimise NAT tables (get rid of prefix lists in tables)

qmonnet:
For stateless NAT, replace the multiple lists of prefixes that are stored for computing the mapping to translate addresses. Instead, associate to each IP range from the original prefixes a range of the same size among the target prefixes. We end up with more entries (potentially), but each entry contains only three IP addresses instead of the prefix list. This also saves us from looping on whole prefix lists when computing the address mappings. This also allows us to remove the iplist submodule, unused after the update.

Please refer to commit descriptions for more details.


Fredi-raspall:
Marking this as don't merge until other PRs get in

qmonnet:
Talking with Daniel, the current sentiment is that this PR (and my other one on `PrefixSize` arithmetic) should likely go in before his own work.

I'll remove the `dont-merge` label (after triple-checking that auto-merge was disabled :see_no_evil:).

PR #669, opened 2025-07-05
Stats support

Fredi-raspall:


mgmt/src/processor/proc.rs
```diff
@@ -349,13 +352,31 @@ async fn apply_router_config(
...
+/// refresh mappings for per vpc statistics
+fn update_stats_vpc_mappings(config: &GwConfig, vpcmapw: &mut VpcMapWriter<VpcMapName>) {
+    // create a mapping table frome the vpc table in the config
+    // FIXME(fredi): visibility
+    // FIXME(fredi): generalize the vpcmapName table
+    let vpc_table = &config.external.overlay.vpc_table;
+    let mut vpcmap = VpcMap::<VpcMapName>::new();
+    for vpc in vpc_table.values() {
+        let disc = VpcDiscriminant::VNI(vpc.vni);
+        let map = VpcMapName::new(disc, &vpc.name);
```
mvachhar, line 364: Does this clear the stats on every reconfiguration?  If so, is there any way we can not do this?  I will not have this block this PR, but if it does clear stats for an existing VPC, can we create an issue to fix that?

net/src/packet/meta.rs
```diff
@@ -77,10 +77,15 @@ pub struct PacketMeta {
     pub done: Option<DoneReason>, /* if Some, the reason why a packet was marked as done, including delivery to NF */
     pub src_vni: Option<Vni>, /* the vni value of a received vxlan encap packet, if destined to gateway */
     pub dst_vni: Option<Vni>, /* the vni value of a vxlan packet re-encapsulated by the gateway */
-
-    #[cfg(test)]
-    /* Keep the Packet in spite of calling packet.enforce(). This is for testing */
-    pub keep: bool,
+    pub keep: bool,           /* Keep the Packet even if it should be dropped */
```
mvachhar, line 80: In the future, should we move this to dont_keep or something so that keep is the default behavior?

mvachhar: Can we create an issue to detect kept packets that are inadvertently destroyed?

mvachhar (approved):

Frostman (commented):
A few comments to resolve later, but before first release with GW

mgmt/src/processor/proc.rs
Fredi-raspall: No, this does no clear the stats. This just reconfigures the VPC discriminant-to-VPCname mappings. In fact, the opposite should be addressed. If at some point we're measuring stats for some VPC "A" and, later, that VPC "A" no longer exists because it's removed from the config, the stats for it should be removed and we don't. I will address this one, which I forgot.

net/src/packet/meta.rs
Fredi-raspall: > In the future, should we move this to dont_keep or something so that keep is the default behavior?
We can. Our "initial" stages / NFs called filter_map() when iterating over packets because the assumption was that they would drop packets. If we want to go for an approach where that should not happen, we could replace all filter_maps so that they would never filter. That being said, I think that the current approach where they may filter and we can override the behavior of packet.enforce() by setting the keep flag is the most flexible at the moment.

Fredi-raspall: > Can we create an issue to detect kept packets that are inadvertently destroyed?

dataplane/src/args.rs
```diff
@@ -73,6 +73,19 @@ pub(crate) struct CmdArgs {
...
     frr_agent_path: String,
+
+    /// Prometheus metrics server port
+    #[arg(
+        long,
+        value_name = "PORT",
+        default_value_t = 9090,
+        help = "Port for Prometheus metrics HTTP endpoint"
+    )]
+    metrics_port: u16,
```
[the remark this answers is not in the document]
Fredi-raspall: cc @sergeymatov

dataplane/src/args.rs
```diff
@@ -73,6 +73,19 @@ pub(crate) struct CmdArgs {
         default_value = DEFAULT_FRR_AGENT_PATH
     )]
     frr_agent_path: String,
+
+    /// Prometheus metrics server port
+    #[arg(
+        long,
+        value_name = "PORT",
```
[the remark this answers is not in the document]
Fredi-raspall: cc @sergeymatov

dataplane/src/statistics/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+                .build()
+                .expect("Tokio runtime creation failed for metrics server");
+
+            /* block thread to run metrics HTTP server */
+            rt.block_on(async {
+                let app = Router::new()
+                    .route("/metrics", get(metrics_handler))
+                    .with_state(handler_for_thread.as_ref().clone());
+
+                let addr = std::net::SocketAddr::from(([0, 0, 0, 0], port));
```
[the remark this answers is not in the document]
Fredi-raspall: cc @sergeymatov

PR #670, opened 2025-07-05
Nat support

Fredi-raspall:
This targets https://github.com/githedgehog/dataplane/pull/669, which should be merged first

### DIFF
--- a/dataplane/src/packet_processor/ingress.rs
+++ b/dataplane/src/packet_processor/ingress.rs
@@ -45,7 +45,12 @@ impl Ingress {
         if packet.try_ipv4().is_some() || packet.try_ipv6().is_some() {
             match &interface.attachment {
                 Some(Attachment::VRF(fibr)) => {
-                    let vrfid = fibr.get_id().unwrap().as_u32();
+                    let Some(vrfid) = fibr.get_id().map(|x| x.as_u32()) else {
+                        /* we may ocassionaly not be able to enter a fib on reconfigs */
+                        warn!("Failed to access fib on ingress!");
+                        packet.done(DoneReason::Unroutable);
+                        return;
+                    };
                     debug!("{nfi}: Packet is for VRF {vrfid}");
                     packet.get_meta_mut().vrf = Some(vrfid);
                 }
@@ -90,8 +95,8 @@ impl Ingress {
         interface: &Interface,
         packet: &mut Packet<Buf>,
     ) {
-        let nfi = self.name();
         if let Some(if_mac) = interface.get_mac() {
+            let nfi = self.name();
             trace!(
                 "{nfi}: Got packet over interface '{}' ({}) mac:{if_mac}",
                 interface.name, interface.ifindex
@@ -142,8 +147,8 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for Ingress {
         input: Input,
     ) -> impl Iterator<Item = Packet<Buf>> + 'a {
         trace!("{}", self.name);
-        let nfi = self.name.clone();
         input.filter_map(move |mut packet| {
+            let nfi = self.name();
             if !packet.is_done() {
                 if let Some(iftable) = self.iftr.enter() {
                     let iif = packet.get_meta().iif.get_id();
--- a/mgmt/src/models/internal/natconfig/table_extend.rs
+++ b/mgmt/src/models/internal/natconfig/table_extend.rs
@@ -3,10 +3,11 @@
 
 //! NAT rule tables entries creation
 
-use crate::models::external::overlay::vpc::Peering;
+use crate::models::external::overlay::vpc::{Peering, VpcTable};
 use crate::models::external::overlay::vpcpeering::{VpcExpose, VpcManifest};
 use nat::stateless::config::prefixtrie::{PrefixTrie, TrieError};
 use nat::stateless::config::tables::{NatPrefixRuleTable, PerVniTable, TrieValue};
+use net::vxlan::Vni;
 use routing::prefix::Prefix;
 use std::collections::BTreeSet;
 use std::fmt::Debug;
@@ -24,20 +25,20 @@ pub enum NatPeeringError {
 
 /// Create a [`TrieValue`] from the public side of a [`VpcExpose`], for a given prefix in this
 /// [`VpcExpose`]
-fn get_public_trie_value(expose: &VpcExpose, prefix: &Prefix) -> TrieValue {
+fn get_public_trie_value(vni: Vni, expose: &VpcExpose, prefix: &Prefix) -> TrieValue {
     let orig = expose.ips.clone();
     let target = expose.as_range.clone();
 
-    TrieValue::new(orig, target)
+    TrieValue::new(vni, orig, target)
 }
 
 /// Create a [`TrieValue`] from the private side of a [`VpcExpose`], for a given prefix in this
 /// [`VpcExpose`]
-fn get_private_trie_value(expose: &VpcExpose, prefix: &Prefix) -> TrieValue {
+fn get_private_trie_value(vni: Vni, expose: &VpcExpose, prefix: &Prefix) -> TrieValue {
     let orig = expose.ips.clone();
     let target = expose.as_range.clone();
 
-    TrieValue::new(orig, target)
+    TrieValue::new(vni, orig, target)
 }
 
 // Note: add_peering(table, peering) should be part of PerVniTable, but we prefer to keep it in a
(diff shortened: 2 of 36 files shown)


dataplane/src/packet_processor/ingress.rs
```diff
@@ -45,7 +45,12 @@ impl Ingress {
         if packet.try_ipv4().is_some() || packet.try_ipv6().is_some() {
             match &interface.attachment {
                 Some(Attachment::VRF(fibr)) => {
-                    let vrfid = fibr.get_id().unwrap().as_u32();
+                    let Some(vrfid) = fibr.get_id().map(|x| x.as_u32()) else {
+                        /* we may ocassionaly not be able to enter a fib on reconfigs */
+                        warn!("Failed to access fib on ingress!");
```
mvachhar, line 50: When does this happen and why?

routing/src/fib/fibobjects.rs
```diff
@@ -3,6 +3,8 @@
 
 //! Module that contains definitions and methods for fib objects
 
+use net::vxlan::Vni;
```
mvachhar, line 6: I am really not a fan of the behavior added by this commit, seems really hacky and confusing for users to see prefixes that match but are not used in a VRF.  My understanding is we don't really use this and may remove it later.  Please correct me if I am wrong.

mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -3,10 +3,11 @@
 
 //! NAT rule tables entries creation
 
-use crate::models::external::overlay::vpc::Peering;
+use crate::models::external::overlay::vpc::{Peering, VpcTable};
```
mvachhar, line 6: As discussed we will move the destination vni selection logic to its own pipeline stage at a future date, again please correct me if I am mistaken.

mgmt/src/processor/confbuild/internal.rs
```diff
@@ -225,7 +225,7 @@ fn vpc_vrf_config(vpc: &Vpc) -> Result<VrfConfig, ConfigError> {
 
 fn vpc_bgp_af_ipv4_unicast(vpc_rconf: &VpcRoutingConfigIpv4) -> AfIpv4Ucast {
     let mut af = AfIpv4Ucast::new();
-    af.set_vrf_imports(vpc_rconf.vrf_imports.clone());
+    //af.set_vrf_imports(vpc_rconf.vrf_imports.clone());
```
mvachhar, line 228: Maybe a comment here as to why this is disabled now?  I'm not going to block the PR on this, but if it could be added before merge, that would be great.

nat/src/stateless/mod.rs
```diff
@@ -212,6 +212,7 @@ impl StatelessNat {
                     packet.get_meta_mut().dst_vni = Some(vni);
                 }
                 if modified {
+                    packet.update_checksums();
```
mvachhar, line 215: This is not correct, we should recompute the checksum for the inner packet on encap.  Can we please create an issue to do this, or fix it before merge?

mvachhar (approved):
Please take a look at the comments before merging.  There are some things to address (like creating issues, ideally adding a comment or two) but I don't need to re-review the PR before it merges so I am choosing approve here.

nat/src/stateless/mod.rs
Fredi-raspall: I don't think this is incorrect.  We have to decide what components are responsible for what. I agree that doing that on encap would be more generic and thought about it. However, if we go that path, we'll recompute the checksum for every encaped packet, which will unnecessarily add processing overhead. The NAT stage knows if addresses were changed or not; that's the reason for adding this there. We can, if folks so prefer, have the best of the two worlds by doing that on encap conditioned on a packet annotation that indicates if it needs to be done.

dataplane/src/packet_processor/ingress.rs
Fredi-raspall: It is a sanity to get rid of the unwrap(). It could happen during the process of removing a VRF until we'd detach the interface.

mgmt/src/models/internal/natconfig/table_extend.rs
Fredi-raspall: You're not mistaken. But until we have that stage, we need this to be done by the nat code.

routing/src/fib/fibobjects.rs
Fredi-raspall: Yes, we no longer need this as it is with the changes later in the branch.

nat/src/stateless/mod.rs
```diff
@@ -86,53 +80,84 @@ impl StatelessNat {
...
+    fn translate_src(&self, net: &mut Net, ranges_src_nat: &TrieValue) -> Result<bool, NatError> {
+        let nfi = self.name();
+        let current_src = net.src_addr();
+        let target_src = map_ip_src_nat(ranges_src_nat, &current_src);
+        if target_src == current_src {
+            return Ok(false);
+        }
+        match (net, target_src) {
+            (Net::Ipv4(hdr), IpAddr::V4(src)) => {
+                debug!("{nfi}: Changing ipv4 src: {current_src} -> {src}");
```
qmonnet, line 102: Same, I'm not a fan of starting a log line with a non-deterministic name. I can clean this up later, though.

nat/src/stateless/mod.rs
```diff
@@ -143,19 +168,38 @@ impl StatelessNat {
...
-            warn!("{}: Packet comes without vni annotation!", self.name);
+            warn!("{nfi}: Packet has no vni annotation!. Will drop...");
             packet.done(DoneReason::Unroutable);
             return;
         };
-        let Some(net) = packet.headers_mut().try_ip_mut() else {
+
+        /* get per vni table */
+        let Some(table) = nat_tables.get_table(vni) else {
+            error!("{nfi}: Cant't find nat tables for vni {vni}");
```
qmonnet, line 192: Typo

```suggestion
            error!("{nfi}: Can't find NAT tables for VNI {vni}");
```

nat/src/stateless/mod.rs
```diff
@@ -166,17 +210,16 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for StatelessNat {
         input: Input,
     ) -> impl Iterator<Item = Packet<Buf>> + 'a {
         input.filter_map(|mut packet| {
-            if !packet.is_done() {
+            if !packet.is_done() && packet.get_meta().nat {
```
qmonnet, line 232: Nit: I'd maybe rename `.nat` to `.nat_required` or something like that, it would make the check clearer.

mgmt/src/models/internal/natconfig/test.rs
```diff
@@ -173,12 +173,30 @@ mod tests {
             remote_id: "67890".try_into().expect("Failed to create VPC ID"),
         };
 
+        // This code is extremely convoluted
+        let mut vpctable = VpcTable::new();
```
qmonnet, line 177: What does that mean? Any suggestion for improvement in particular?

qmonnet (approved):
Some minor comments, but not blockers from my side. Some NAT clean-up changes could have been submitted separately, it might have made review (and my own work) easier, although I do understand the timing constraints. Great work, thank you Fredi!

nat/src/stateless/mod.rs
Fredi-raspall: Why? That's the name of the stage.

mgmt/src/models/internal/natconfig/test.rs
Fredi-raspall: I added this as a reminder that the tests need to be cleaned up so that they are easier to maintain.

nat/src/stateless/mod.rs
qmonnet: The person who writes the pipeline stage can use whatever name they want, right? They can pick something that looks like `StatelessNAT01` which is explicit, or just as well something like `demo-next-month-xlate` which may or may not be intuitive when parsing the logs. Prefixing with `StatelessNat` would make it obvious what sort of stage we're talking about, however it is named.

nat/src/stateless/mod.rs
```diff
@@ -155,6 +163,8 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for StatelessNat {
                 } else {
                     packet.done(DoneReason::InternalFailure);
                 }
+            } else {
+                warn!("Packet is done and will not NATed");
```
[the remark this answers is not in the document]
qmonnet: You're probably right, I can't find it on the full diff. Sorry.

nat/src/stateless/mod.rs
```diff
@@ -165,7 +176,7 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for StatelessNat {
                     packet.done(DoneReason::InternalFailure);
                 }
             } else {
-                warn!("Packet is done and will not NATed");
+                warn!("{}: Packet is done and will not NATed", self.name);
```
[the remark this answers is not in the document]
mvachhar: This is a larger conversation, I'm going to not have this addressed in this PR.

nat/src/stateless/mod.rs
mvachhar: After talking to Fredi, we can leave this, `nat_required` is long, we can decide how to name these flags later.

nat/src/stateless/mod.rs
mvachhar: Done.

PR #672, opened 2025-07-07
Pr/daniel noland/replace iptrie

daniel-noland:


mgmt/src/models/internal/natconfig/table_extend.rs
```diff
@@ -705,44 +704,44 @@ mod tests {
...
-                }
-                for addr in addrs {
-                    let addr_net = IpNet::from(*addr);
-                    let oracle_result = prefix_oracle(&addr_net, &prefixes_trie, &excludes_trie);
-                    let collapsed_result = collapsed_prefixes_trie.contains(&addr_net);
-                    assert_eq!(oracle_result, collapsed_result);
-                }
-            });
-    }
+    // fn prefix_oracle(addr: &IpNet, prefixes: &IpRTrieSet, excludes: &IpRTrieSet) -> bool {
```
mvachhar, line 707: Why is this test still commented out?

Fredi-raspall: Good catch!. Forgot to ask too.

qmonnet: @mvachhar What do you mean? You were the one to request and write it (c3ad0cef4711390d8ce6a951386b90db0352207b) 😅

net/src/ipv4/prefix.rs
```diff
@@ -0,0 +1,665 @@
...
+
+impl From<Ipv4PrefixLen> for u8 {
+    fn from(value: Ipv4PrefixLen) -> Self {
+        value.0
+    }
+}
+
+impl From<Ipv4Addr> for Ipv4Prefix {
+    fn from(value: Ipv4Addr) -> Self {
+        Ipv4Prefix::new_assert(value.octets(), Ipv4PrefixLen::MAX.as_u8())
```
Fredi-raspall, line 120: I see this might cause trouble.

net/src/ipv4/prefix.rs
```diff
@@ -0,0 +1,665 @@
...
+    /// Useful in const contexts and testing.
+    ///
+    /// Avoid this method outside const contexts or testing settings.
+    ///
+    /// # Panics
+    ///
+    /// * Panics if the provided prefix is greater than 32
+    /// * Panics if the provided address contains non-network bits.
+    #[must_use]
+    pub const fn new_assert(addr: [u8; 4], prefix: u8) -> Self {
```
Fredi-raspall, line 162: Shouldn't we gate this with cfg(test)? ... and if it is to be used from another crate behind some feature?
In other parts we created feature "testing"

net/src/ipv4/prefix.rs
```diff
@@ -0,0 +1,665 @@
...
+    ///    this does not happen).
+    /// 2. You need to convert an interface address assignment into a route
+    ///
+    /// This method will log a warning if the provided address contains non-network bits.
+    ///
+    /// # Errors
+    ///
+    /// * Returns an error if the provided prefix length is greater than [`Ipv4PrefixLen::MAX_LEN`].
+    #[tracing::instrument(level = "debug")]
+    pub fn new_tolerant<E>(
```
Fredi-raspall, line 219: I see the merit of this vs new_strict(). However, shouldn't we just agree on what's acceptable and have a single fallible constructor?

net/src/ipv4/prefix.rs
daniel-noland: Normally yest but I wouldn't recommend that in this case.

This method is `const fn`, and it is completely reasonable (and encouraged) to invoke a method with potential panic at compile time.  Better to catch the bug before you even submit the PR if possible :).  For example, if you want to define a `const ROOT: Ipv4Prefix = Ipv4Prefix::new_assert([0, 0, 0, 0], 0)` you can do so quite safely.

   In fact, it is better to do it that way; many IDEs can give you real time error information if a `const fn` will panic on compile based on const arguments.  This use case is one of the reasons I wrote the function signature with such plain types.  If we only use it in const / testing contexts there is zero benefit in the more complex types.  They just produce syntax noise.  If rust's const generics were more advanced I would have written it slightly differently, but we aren't quite there yet with the compiler.
   
That said, calling this method outside a const context is usually unwise.

net/src/ipv4/prefix.rs
daniel-noland: I am happy to rename the method to `new` if that reduces syntax noise, but the other constructor is still quite useful in other settings.

It is one thing to accept invalid input and hammer it into "validity" by squinting at it and hoping you know what it means.

It is quite another to ignore invariants _within_ our own data structures.

For example, I think it is sorta kinda ok to accept a route with the host bits set under the theory that you know the prefix length and you are just assuming that the data source forgot to clear the low order bits or something.

It is another thing entirely to fail to honor invariants within your own data structures.  As we have discussed before, there isn't really any way to recover from internal programmer error.

routing/src/prefix.rs
```diff
@@ -46,12 +46,12 @@ impl Prefix {
     /// Build 0.0.0.0/0. "Default" is a very overloaded term. Calling this `root_v4`.
     #[must_use]
     pub fn root_v4() -> Prefix {
-        Prefix::IPV4(Ipv4Prefix::new_assert([0, 0, 0, 0], 0))
+        Prefix::IPV4(Ipv4Prefix::ROOT)
```
daniel-noland, line 49: yup.  This is a good fix.

mgmt/src/models/internal/natconfig/table_extend.rs
daniel-noland: I commented it out because it was the last test I needed to convert over to the new prefix map and I wanted to run this through CI before I did the last step

mvachhar: I can't even remember what is was supposed to do :(

mvachhar: Ahh, right, this tests the prefix calculator, we need this test.

net/src/ipv4/prefix.rs
mvachhar: Agreed, Daniel and I discussed not using this in the code.  I'd be in favor of dropping it, but I am not hell bent on that.  But it certainly should not be used on the datapath.

nat/src/stateless/config/prefixtrie.rs
```diff
@@ -181,7 +160,7 @@ mod tests {
     }
 
     fn prefix_v6(s: &str) -> Ipv6Prefix {
-        Ipv6Prefix::from_str(s).expect("Invalid IPv6 prefix")
+        Ipv6Prefix::from_str(s).unwrap()
```
qmonnet, line 163: Why, and how is this relevant to this PR?

nat/src/stateless/config/prefixtrie.rs
```diff
@@ -202,8 +181,8 @@ mod tests {
             .expect("Failed to insert prefix");
 
         pt.insert_ipv6(
-            prefix_v6("aa:bb:cc:dd::/32"),
-            "prefix_aa:bb:cc:dd::/32".to_string(),
+            prefix_v6("aa:bb:cc:dd::/64"),
+            "prefix_aa:bb:cc:dd::/64".to_string(),
```
qmonnet, line 185: Why does this need to change? It's probably a concern if inserting the `/32` doesn't work anymore?

routing/src/fib/fibtype.rs
```diff
@@ -219,11 +219,17 @@ impl Fib {
     pub fn lpm_with_prefix(&self, target: &IpAddr) -> (Prefix, &FibGroup) {
         match target {
             IpAddr::V4(a) => {
-                let (prefix, group) = self.routesv4.lookup(a);
+                let (prefix, group) = self
+                    .routesv4
+                    .get_lpm(&Ipv4Prefix::from(*a))
+                    .unwrap_or_else(|| unreachable!());
```
qmonnet, line 225: How can you be sure you never reach this? My understanding is that `.get_lpm()` can return `None`, contrarily to the `iptrie` lookup.

routing/src/prefix.rs
```diff
@@ -166,7 +167,7 @@ impl Prefix {
         T: TryInto<Prefix>,
         T::Error: Debug,
     {
-        val.try_into().expect("Invalid prefix")
+        val.try_into().unwrap()
```
qmonnet, line 170: why?

net/src/ipv4/prefix.rs
```diff
@@ -0,0 +1,665 @@
...
+    AddrParseError(AddrParseError),
+    /// invalid ip or prefix length
+    #[error(transparent)]
+    InvalidIpv4Network(InvalidIpv4Network),
+}
+
+impl Ipv4PrefixLen {
+    /// The largest possible prefix length for IPv4 (i.e. /32)
+    pub const MAX_LEN: u8 = 32;
+    /// The largest possible prefix length for IPv4 (i.e. /32)
```
qmonnet, line 67: Nit: s/largest/longest/ ? (I find that “largest” is somewhat confusing in that it lead to believe the prefix with the “largest” size will be the “largest”, when it is in fact the smaller one.)

net/src/ipv4/prefix.rs
```diff
@@ -0,0 +1,665 @@
...
+
+    /// Create an [`Ipv4Prefix`] even if the argument contains non-network bits.
+    ///
+    /// This method is useful in the cases that
+    ///
+    /// 1. You can't trust your routing input and still don't wish to reject the routes (ideally,
+    ///    this does not happen).
+    /// 2. You need to convert an interface address assignment into a route
+    ///
+    /// This method will log a warning if the provided address contains non-network bits.
```
qmonnet, line 213: I'd expect a `warn!()` for a warning, but this is only a `debug!()` message.

net/src/ipv6/prefix.rs
```diff
@@ -0,0 +1,669 @@
...
+pub enum Ipv6PrefixParseError {
+    /// failure to interpret string as an ip and a prefix length
+    #[error(transparent)]
+    AddrParseError(AddrParseError),
+    /// invalid ip or prefix length
+    #[error(transparent)]
+    InvalidIpv6Network(InvalidIpv6Network),
+}
+
+impl Ipv6PrefixLen {
```
qmonnet, line 65: Method suggestion: `pub fn to_prefix_size(&self) -> PrefixSize` (for a follow-up :slightly_smiling_face: or I may end up adding it when rebasing my PR).

net/src/ipv6/prefix.rs
```diff
@@ -0,0 +1,669 @@
...
+
+    /// Create an [`Ipv6Prefix`] even if the argument contains non-network bits.
+    ///
+    /// This method is useful in the cases that
+    ///
+    /// 1. You can't trust your routing input and still don't wish to reject the routes (ideally,
+    ///    this does not happen).
+    /// 2. You need to convert an interface address assignment into a route
+    ///
+    /// This method will log a warning if the provided address contains non-network bits.
```
qmonnet, line 216: Same comment warning/debug

qmonnet (commented):
I've got some very minor observations, ~apart from lookups in `lpm_with_prefix()` or `lpm_v4()`/`lpm_v6()`, where I don't understand what keeps us from hitting the `unreachable!()` section (I'm maybe missing something)~

net/src/ipv4/prefix.rs
Fredi-raspall: agree

routing/src/prefix.rs
daniel-noland: oh, I was debugging the test and the expect was eating the error message

net/src/ipv4/prefix.rs
daniel-noland: fair enough

net/src/ipv6/prefix.rs
daniel-noland: Yup.  Will leave that for another PR tho

routing/src/prefix.rs
qmonnet: Oh, I never realised it would :exploding_head:

nat/src/stateless/config/prefixtrie.rs
qmonnet: Ah you mean it's a requirement now. OK, makes sense to adjust the test then, thank you.

net/src/ip/prefix.rs
```diff
@@ -0,0 +1,419 @@
...
+    }
+}
+
+impl From<Ipv6Prefix> for IpPrefix {
+    fn from(value: Ipv6Prefix) -> Self {
+        IpPrefix::V6(value)
+    }
+}
+
+impl IpPrefix {
```
qmonnet, line 114: This looks redundant with `routing::prefix::Prefix`? Do we need to keep both versions?

qmonnet: Ok we chatted about it and I get it now - the plan is to move the `routing`'s version into the `net` crate instead.

mvachhar:
This will be replaced by a better implementation.

PR #711, opened 2025-07-16
Frr-integration-part-1

Fredi-raspall:
Fixes https://github.com/githedgehog/dataplane/issues/684
Fixes https://github.com/githedgehog/dataplane/issues/701
Fixes https://github.com/githedgehog/dataplane/issues/702
Fixes https://github.com/githedgehog/dataplane/issues/703
Fixes https://github.com/githedgehog/dataplane/issues/704
Partially addresses https://github.com/githedgehog/dataplane/issues/708

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ cli = { path = "./cli", package = "dataplane-cli", features = [] }
 dpdk = { path = "./dpdk", package = "dataplane-dpdk" }
 dpdk-sys = { path = "./dpdk-sys", package = "dataplane-dpdk-sys" }
 dpdk-sysroot-helper = { path = "./dpdk-sysroot-helper", package = "dataplane-dpdk-sysroot-helper" }
-dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", version = "1.1.0"}
+dplane-rpc = { git = "https://github.com/githedgehog/dplane-rpc.git", version = "1.1.1"}
 errno = { path = "./errno", package = "dataplane-errno" }
 gateway_config = { git = "https://github.com/githedgehog/gateway-proto", tag = "v0.11.0", version = "0.11.0" }
 id = { path = "./id", package = "dataplane-id" }


cli/bin/cmdtree_dp.rs
```diff
@@ -9,6 +9,13 @@ use log::Level;
 use std::convert::AsRef;
 use strum::IntoEnumIterator;
 
+fn cmd_show_cpi() -> Node {
```
mvachhar, line 12: Any reason we didn't make these Prometheus stats?  Is there a way to have hidden or second tier stats there?  I don't think we should bother changing it now, I am just curious.

routing/src/frrmi.rs
```diff
@@ -0,0 +1,384 @@
...
+use std::str::from_utf8;
+use std::time::Instant;
+use std::{collections::VecDeque, time::Duration};
+use thiserror::Error;
+
+#[allow(unused)]
+use tracing::{debug, error, info, warn};
+
+#[derive(Error, Debug)]
+pub(crate) enum FrrErr {
```
mvachhar, line 19: Since we are not using tokio, is it possible for this thread to get blocked on Tx in a way that causes issues (e.g., receiving new routes, etc.)?  Generally, if this thread blocks I presume something is wrong with FRR but the agent could be hosed but FRR ok.  What do we do in that case?  Again, does not need to be fixed for this PR but if this is an issue, we should create an issue and put it in the R1 release.  We can triage it out to R2 based on where we get timeline wise.

routing/src/frrmi.rs
```diff
@@ -0,0 +1,384 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use mio::net::UnixStream;
```
mvachhar, line 5: Why is the old code not deleted in this PR?

routing/src/frrmi.rs
```diff
@@ -332,7 +372,8 @@ impl Frrmi {
             self.stats.last_fail_genid = Some(response.genid);
             self.stats.apply_failures += 1;
             let out = response.get_response_data();
-            error!("Failed to apply FRR configuration: {out}");
+            error!("Failed to apply FRR configuration for gen {respgen}: {out}");
+            self.config_retry(request);
```
mvachhar, line 376: The commit says that frrmi does not try to reapply configuration but this looks like it does?  Am I reading this wrong?

Also, is there no case where there can be an intermittent failure of frr-reload to apply the config but a retry would succeed?

routing/src/rio.rs
```diff
@@ -165,7 +165,7 @@ impl Rio {
         let poller = Poll::new().map_err(|_| RouterError::Internal("Poll creation failed"))?;
         poller
             .registry()
-            .register(&mut ev_cpsock, CPSOCK, Interest::READABLE)
+            .register(&mut ev_cpsock, CPSOCK, Interest::PRIORITY)
```
mvachhar, line 168: Does this behavior make it possible to get stale FRR updates that we don't recognize as stale because they are left on the socket buffer after FRR restarts?  E.g., frr sends messages and they get buffered, frr gets restarted, dataplane starts reading messages but starts with the stale ones that were buffered making us out of sync with frr.

mvachhar (commented):
I have a few questions that need answers before I approve, but they do not necessarily require changes to fix.

cli/bin/cmdtree_dp.rs
Fredi-raspall: I'm not sure I'd expose those. These are not metrics that a user would be interested in. They are mainly for troubleshooting.

routing/src/rio.rs
Fredi-raspall: That's a good point. This behavior makes us buffer messages until frr gets its very first configuration. In principle, FRR should not send anything until that happens, unless it learns things from kernel (addresses, routes). Now, if FRR gets restarted afterwards, what you're suggesting is feasible. This mechanism is not mean to address that however, nor causes it. The problem you're mentioning needs to be addressed in a different way. Fundamentally, the question is how to know if messages need to be processed if, later, FRR may restart. We can't anticipate this situation, so any measure to avoid the problem needs to be enacted after the fact. This may include flushing old state on FRR restart, but this goes against trying to preserve connectivity while FRR restarts. We may need to implement some mechanism of deferred flush as in graceful restart, which is outside the scope of this PR.

Fredi-raspall: See https://github.com/githedgehog/dataplane/issues/719

routing/src/cpi.rs
```diff
@@ -1,483 +1,377 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Open Network Fabric Authors
 
-//! Control-plane interface (CPI)
+//! Main processing functions of the CPI
```
qmonnet, line 4: Nit: I'd like to keep the expanded “Control-plane interface” somewhere, so that readers can understand what CPI means. This was a good place to have it. Maybe like this?:

```suggestion
//! Main processing functions of the CPI (Control-plane interface)
```

routing/src/rio.rs
```diff
@@ -0,0 +1,488 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Router IO, which includes the control-plane interface CPI and the FRR management interface (FRRMI)
```
qmonnet, line 4: :heart:

routing/src/frrmi.rs
```diff
@@ -45,6 +46,20 @@ pub(crate) struct Frrmi {
     inservice: Option<FrrmiRequest>,  /* the request currently being serviced */
     timeout: Option<Instant>,         /* timeout for the current request in service */
     requests: VecDeque<FrrmiRequest>, /* queue of other requests to frr-agent */
+    stats: FrrmiStats,                /* stats */
+}
+
+/// Stats for the `Frrmi`
+#[derive(Default)]
+pub(crate) struct FrrmiStats {
```
qmonnet, line 67: Just asking, are these stats exported to Prometheus or are they simply exposed - as I believe - to the CLI?

routing/src/cpi.rs
```diff
@@ -325,7 +325,8 @@ fn handle_request(
 
     // ignore additions if have no config. Connects are allowed, so are deletions to wipe out old state
     if !db.have_config() && op == RpcOp::Add {
-        debug!("Ignoring request: no config is available");
+        error!("Ignoring request: there's no config. This should not happen...");
+        error!("... but won't cause malfunction. Trust the 🦔's");
```
qmonnet, line 329: Would you please consider removing the last part of the comment?
- If it _does_ cause malfunction because of whatever bug, it's not good for trust, and might create frustration with users seeing this.
- I would also recommend against tying (even more) the open-source project to the vendor company. Others could work on this in the future and not be part of Hedgehog.

```suggestion
        error!("Ignoring request: there's no config. This should not happen, but won't cause malfunction if it does.");
```

cli/bin/cmdtree_dp.rs
```diff
@@ -268,11 +268,24 @@ fn cmd_local() -> Node {
     root
 }
 
-#[allow(unused)]
+fn cmd_frrmi_apply_last() -> Node {
+    let mut root = Node::new("apply");
+    root += Node::new("last-config")
+        .desc("Apply the last config in FRR")
+        .action(CliAction::FrrmiApplyLastConfig as u16);
```
qmonnet, line 275: Is there a command to display this config before applying it?

Fredi-raspall: Yes, you can do `show frrmi last-config`.
The apply command is only meant for troubleshooting.

Fredi-raspall: But it will have already been applied...

mvachhar (approved):

mgmt/src/processor/proc.rs
```diff
@@ -393,18 +393,6 @@ async fn apply_gw_config(
...
-       let res: Result<(), ConfigError> = frrmi
-           .probe()
-           .await
-           .map_err(|_| ConfigError::FrrAgentUnreachable);
-
-       if genid != ExternalConfig::BLANK_GENID && res.is_err() {
-           return Err(ConfigError::FrrAgentUnreachable);
-       }
-    */
-
```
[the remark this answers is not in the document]
qmonnet: It does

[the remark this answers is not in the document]
qmonnet: Oh you mean the PR _removes_ it. Silly me :facepalm:. Sorry for the noise.

routing/src/cpi.rs
qmonnet: :cry: I like hedgehogs too! But I believe it's for the best. Thanks!

PR #712, opened 2025-07-17
ci: add optional vlab job

Frostman:


qmonnet:
Nice to have a VLAB job!

> It's in the dev workflow b/c it relies on the image pushed in the "debug sterile" and we should only run vlab and especially hlab if all unit tests and etc. passed.

Not a fan of this part:
- We run the full workflow each time we add or remove a label to the PR. I just added `ci` to mark it as CI-related, and we're in for another run :cry:
- I don't like having the current dev.yml and the new VLAB/HLAB job mixed together in the same YAML file, makes it harder to follow what's going on in CI in my opinion.

Could we move the job to a dedicated workflow file? Given that we push the image from the sterile build, we should be able to [wait for it](https://github.com/cilium/cilium/blob/main/.github/actions/wait-for-images/action.yaml) and then pull it from a separate workflow? Same for the test results, we can probably wait and check through GH API that they succeeded for the PR, before running the workflow?

.github/workflows/dev.yml
```diff
@@ -270,22 +276,72 @@ jobs:
...
+    if: "${{ needs.check_changes.outputs.devfiles == 'true' }}"
+    needs:
+      - check
+
+    name: "${{ matrix.hybrid && 'hlab' || 'vlab' }}-${{ matrix.fabricmode == 'spine-leaf' && 'sl' || 'cc' }}-${{ matrix.gateway && 'gw-' || '' }}${{ matrix.includeonie && 'onie-' || '' }}${{ matrix.buildmode }}-${{ matrix.vpcmode }}"
+
+    uses: githedgehog/fabricator/.github/workflows/run-vlab.yaml@master
+    with:
+      skip: ${{ matrix.hybrid && !contains(github.event.pull_request.labels.*.name, 'ci:+hlab') || !matrix.hybrid && !contains(github.event.pull_request.labels.*.name, 'ci:+vlab') }}
+      fabricatorref: master
```
qmonnet, line 289: Should there be a way to configure this to pass a specific fabricator reference?

.github/workflows/dev.yml
```diff
@@ -270,22 +276,72 @@ jobs:
...
+        if: ${{ needs.check.result != 'success' && needs.check.result != 'skipped' }}
+        run: |
+          echo '::error:: Some check job(s) failed'
+          exit 1
+      - name: "Flag any vlab matrix failures"
+        if: ${{ needs.vlab.result != 'success' && needs.vlab.result != 'skipped' }}
         run: |
-          echo '::error:: Some job(s) failed'
+          echo '::error:: Some vlab job(s) failed'
           exit 1
```
qmonnet, line 347: This makes the VLAB job as required to pass (if not skipped) for merging. Is this intentional? Do we expect it to be stable enough at the moment?

.github/workflows/dev.yml
qmonnet: Yes, that's the use cases I had in mind :slightly_smiling_face:

Frostman:
> * We run the full workflow each time we add or remove a label to the PR. I just added `ci` to mark it as CI-related, and we're in for another run 😢
Yes, I don't like that part either, but unfortunately, there is no easy way to achieve the same configurability on the PRs. Having a bot and controlling it with comments is a bit nicer (more granular but less visibility), but it requires way more work so maybe some day (c).

> * I don't like having the current dev.yml and the new VLAB/HLAB job mixed together in the same YAML file, makes it harder to follow what's going on in CI in my opinion.
IMO it makes it way harder to follow what's happening in CI when splitting similar jobs (testing-related in that case) into separate workflows. It makes sense to me when you split out some security scanning, commitlint, etc into separate workflows.

> Could we move the job to a dedicated workflow file? Given that we push the image from the sterile build, we should be able to [wait for it](https://github.com/cilium/cilium/blob/main/.github/actions/wait-for-images/action.yaml) and then pull it from a separate workflow? Same for the test results, we can probably wait and check through GH API that they succeeded for the PR, before running the workflow?
Sorry, I really don't like that approach for two reasons - one is a previous comment, and the other is that it adds unnecessary complication to the workflow.  It may make sense for a big project like Cilium, but IMO it still makes it very hard to follow. Let's get this one merged. If you think it's important, you can work on it later when you have time.

qmonnet (changes requested):
No. Please let's at least add filters on the labels to run the workflows only when HLAB/VLAB labels are used, and not other labels.

Frostman:
> No. Please let's at least add filters on the labels to run the workflows only when HLAB/VLAB labels are used, and not other labels.

I'm not sure what you want to do. Could you please elaborate? If you're talking about somehow figuring out what labels were changed in the trigger event, it's not going to work b/c when you're editing labels a separate event is added on every addition/removal and concurrency setting will get all but the last one cancelled

qmonnet:
Yes, that's also what I was realising. My objective would be to avoid re-triggering the CI when we change labels. I was thinking of skipping the jobs in that case, but that won't do, because it would still override the status check and mark everything as skipped, without letting us check that the workflow have succeeded before we changed the labels.

I really don't like the new run on labels :cry: It's going to be super frustrating to mark a workflow as ready by removing a `dont-merge` label, only to go through a full run again even if CI was green.

qmonnet:
> IMO it makes it way harder to follow what's happening in CI when splitting similar jobs (testing-related in that case) into separate workflows. It makes sense to me when you split out some security scanning, commitlint, etc into separate workflows.

That's a side discussion, but I disagree here, by the way. Build + unit tests are quite different from hardware-based e2e tests in my mind, and mixing all them together makes it difficult to follow what's running and under what conditions - the case in point is a good example, we're mixing tests that run on each PR with tests triggered by labels, I really find it confusing.

qmonnet:
Another possibility would be to go through a `workflow_dispatch` for the HLAB/VLAB tests? A bit more involved for triggering, though, so it's not ideal, either (and would require moving them to a separate workflow file).

.github/workflows/dev.yml
qmonnet: OK I can work with that, indeed

Frostman:
> Discussed at the meeting - one alternative could be to use labels to decide whether or not to run the `vlab` job, but to _not_ trigger on label changes - instead, leaving it to contributors to trigger a new workflow run, if necessary, by updating or closing/reopening the PR.

Yup, I think it's a good solution for now to at least avoid re-triggers on manipulating the labels

.github/workflows/dev.yml
```diff
@@ -10,7 +10,11 @@
 name: "dev.yml"
 
 on:
-  pull_request: { }
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
```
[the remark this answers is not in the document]
Frostman: good point, those are defaults

qmonnet (approved):
OK Let's go for it

PR #714, opened 2025-07-17
Pull model types out of mgmt crate

Fredi-raspall:
N.B. This rides on  https://github.com/githedgehog/dataplane/pull/711

Goal: avoid circular dependencies with mgmt crate.

* Moves the internal/external model types out of mgmt crate into a new `config` crate (reads nicer when aliased to dataplane-config).
* Moves gRPC converters `config`. This is needed as those implement TryForm trait (orphan rule).
* Addresses orphan rule for frr renderers by defining a new trait.
* Puts all the NAT code and tests back to the nat crate.
* Reorganizes a bit the code and fixes lints.
* A deeper cleanup is possible, but I don't want to defer this PR not to block others.

Several commits should be squashed.
* Those beginning with feat(config) refer to the new crate and should be squashed.
* The  rest could, esp those dealing with NAT since some NAT tests have been unnecessarily moved between config, mgmt and finally nat.

Finally, when that code lived in `mgmt` we defined enum `ConfigError`. That enum has gone to `config` too since many validations are made there. We may opt to rename that enum to ValidationError or the like.

Fixes https://github.com/githedgehog/dataplane/issues/715


mgmt/src/frr/renderer/prefixlist.rs
```diff
@@ -3,40 +3,39 @@
...
-use crate::frr::renderer::builder::{ConfigBuilder, Render};
-use crate::models::internal::routing::prefixlist::*;
-use std::fmt::Display;
+use crate::frr::renderer::builder::{ConfigBuilder, Render, Rendered};
+use config::internal::routing::prefixlist::*;
 
 /* Impl Display */
-impl Display for PrefixListMatchLen {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl Rendered for PrefixListMatchLen {
```
mvachhar, line 10: I'm ok with this for now, but a better way to do this would be to create a transparent wrapper type around `PrefixListMatchLen` and then implement `Display` for that, then the caller just wraps that.  Same for the other types.  But, not worth changing now.

nat/src/stateless/compute/mod.rs
```diff
@@ -5,17 +5,23 @@
 
 #![deny(clippy::all, clippy::pedantic)]
```
mvachhar, line 6: Not a fan of the compute name, I probably would have chosen `natconfig` but not worth holding this up for that either.

mvachhar (approved):

mgmt/src/frr/renderer/prefixlist.rs
Fredi-raspall: Hum. Why? Isn't the current approach simpler (no extra types) and clearer?

nat/src/stateless/compute/mod.rs
Fredi-raspall: I wanted to distinguish this from the actual dataplane processing. Not that I like it very much either and it can be easily changed. This PR is on top of https://github.com/githedgehog/dataplane/pull/711, so until we merged that, merging this is worthless.

Fredi-raspall:
Agree on the don't merge. I think I broke s/t

mgmt/src/tests/nat.rs
```diff
@@ -441,25 +438,23 @@ mod tests {
...
             .nat_table
             .expect("Failed to build NAT tables");
 
         let (mut nat, mut tablesw) = StatelessNat::new("stateless-nat");
-        tablesw.update_nat_tables(nat_tables);
+        tablesw.update_nat_tables(nat_tables.clone());
 
         // Template for other packets
-        let pt = build_test_ipv4_packet(u8::MAX).unwrap();
+        let _pt = build_test_ipv4_packet(u8::MAX).unwrap();
```
qmonnet: I think this line can be removed completely, it seems to be a leftover from when I wrote the tests. Packet build has moved to `check_packet()`.

nat/src/stateless/compute/mod.rs
qmonnet: How about something like “build” or “setup” rather than “compute”?

README.md
```diff
@@ -169,6 +169,21 @@ To do this, add the following to your `.vscode/settings.json` file:
...
+Since Rust is not a good friend of circular dependencies, here come some guidelines to avoid those.
+
+The `dataplane` crate contains the main binary and may include any other as a dependency. The crates developed within this project
+are aliased to `dataplane-NAME` and referred to as internal. There is a set of low-level infrastructure crates with limited internal dependencies
+which many other crates may refer to. This set includes: `net`, `pipeline`, `lpm` or `config`. Note that some of those refer to the others.
+E.g. `net` is a dependency of `pipeline`.
+
+A second tier of crates use the prior set to add extended functionalities. These include `nat` or `routing`. These crates may have `config` as
+dependency, but not vice-versa. Finally, the `mgmt`crate may make use of any the internal crates, but no other crate other than `dataplane` should
+depend on `mgmt`.
```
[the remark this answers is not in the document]
Fredi-raspall: Ok! Will  rewrite later....

mgmt/src/frr/renderer/prefixlist.rs
mvachhar: This interface doesn't use the formatter and constructs a string directly which adds some overhead.  In this case, I think there is nothing wrong with it per se.  However, the standard way to do this in rust is through a formatter, hence the approach I suggested.  But also, it isn't super important, so I don't think it needs to be changed.

The new code would have looked something like:
```
#[repr(transparent)]
struct Rendered<T>(T)

impl Display for Rendered<PrefixListMatchLen> {
  // Same code as before the move
}
```

Then usage would be identical except instead of doing
```
println!("Value: {v}");
```
You would do:
```
println!("Value: {}", Rendered(v))
```

Fredi-raspall: Agreed, we do this in several places. My point in favor of using a very specific trait in that case is that it is for a very particular purpose: making FRR happy. Using such a trait, we minimize the problem of somebody not liking a particular Display and changing it, breaking the integration with FRR.

Fredi-raspall:
I have squashed several commits.

README.md
```diff
@@ -169,6 +169,61 @@ To do this, add the following to your `.vscode/settings.json` file:
...
+ │     │     └──────┬───┘   └──────────┘       │ │
+ │     │            │                          │ │
+┌▼─────▼───┐  ┌─────▼────┐  ┌──────────┐  ┌────▼─▼───┐
+│          │  │          │  │          │  │          │
+│   net    │  │   lpm    │  │ pipeline │  │ config   │  tier-1
+│          │  │          │  │          │  │          │
+└───▲──────┘  └──────────┘  └───┬──────┘  └──────────┘
+    │                           │
+    └───────────────────────────┘
+```
```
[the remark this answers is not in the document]
qmonnet: It is for me :+1: (but then I've got more context than the average visitor)

PR #716, opened 2025-07-17
Quick devel builds for testing

Fredi-raspall:
Augment justfile to allow cached/debug builds with extended debug base and deterministic container tag.

Daniel coded this but we never merged it, and I'm using it every day.

### DIFF
--- a/justfile
+++ b/justfile
@@ -38,8 +38,10 @@ _image_profile := if profile == "debug" { "debug" } else { "release" }
 _dpdk_sys_container_repo := "ghcr.io/githedgehog/dpdk-sys"
 [private]
 _dpdk_sys_container_tag := dpdk_sys_commit
+
 [private]
 _libc_container := _dpdk_sys_container_repo + "/libc-env:" + _dpdk_sys_container_tag + "." + _image_profile
+
 [private]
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -38,8 +38,13 @@ _image_profile := if profile == "debug" { "debug" } else { "release" }
 _dpdk_sys_container_repo := "ghcr.io/githedgehog/dpdk-sys"
 [private]
 _dpdk_sys_container_tag := dpdk_sys_commit
+
 [private]
 _libc_container := _dpdk_sys_container_repo + "/libc-env:" + _dpdk_sys_container_tag + "." + _image_profile
+
+[private]
+_doc_env_container := _dpdk_sys_container_repo + "/doc-env:" + _dpdk_sys_container_tag
+
```
qmonnet, line 47: I think we removed the use of the `doc-env` container, this change is probably accidental?

qmonnet (changes requested):
`build-container-quick` looks good, just one comment below

justfile
Fredi-raspall: Yup, good catch Quentin. Will remove...

Fredi-raspall:
The pipeline is failing for unrelated reasons I believe.
I've opened https://github.com/githedgehog/fabric/issues/995 in case the fuzzing test failures are related to it.
cc @daniel-noland

PR #717, opened 2025-07-17
Tap support phase 1

daniel-noland:
This is the first of a series of PRs which add the ability to inject traffic from our pipeline back into the linux kernel.

Each of the commits messages contain reasoning for why this change is needed.

The first commit, which changes the max lenght of InterfaceName is especially important in order to maintain our contract with linux when we create tap devices via ioctl.

### DIFF
--- a/interface-manager/src/interface/tap.rs
+++ b/interface-manager/src/interface/tap.rs
@@ -1,0 +1,139 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use net::buffer::{PacketBuffer, PacketBufferMut};
+use net::interface::InterfaceName;
+use std::num::NonZero;
+use std::os::fd::AsRawFd;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tracing::{debug, error, info};
+
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct TapDevice {
+    file: tokio::fs::File,
+}
+
+mod helper {
+    /// This is a validated type around a value which is regrettably fragile.
+    ///
+    /// 1. Passed directly to the kernel.
+    /// 2. By a privileged thread.
+    /// 3. In an ioctl.
+    /// 4. By an implicitly null terminated pointer.
+    ///
+    /// As a result, strict checks are in place to ensure memory integrity.
+    ///
+    /// <div class=warning>
+    ///
+    /// It is essential that this type remains transparent.
+    /// Only zero-sized types may be added to this structure as we don't control the ABI.
+    /// We are subject to a contract with the kernel.
+    /// </div>
+    #[repr(transparent)]
+    #[derive(Debug, Copy, Clone)]
+    pub(super) struct InterfaceRequest(libc::ifreq);
+
+    use net::interface::InterfaceName;
+    use nix::libc;
+
+    nix::ioctl_write_ptr_bad!(
+        /// Create a tap device
+        make_tap_device,
+        libc::TUNSETIFF,
+        InterfaceRequest
+    );
+
+    nix::ioctl_write_ptr_bad!(
+        /// Keep the tap device after the program ends
+        persist_tap_device,
+        libc::TUNSETPERSIST,
+        InterfaceRequest
+    );
+
+    impl InterfaceRequest {
+        /// Create a new `InterfaceRequest`.
+        #[cold]
+        #[tracing::instrument(level = "trace")]
+        pub(super) fn new(name: &InterfaceName) -> Self {
+            assert_eq!(
+                libc::IF_NAMESIZE,
+                InterfaceName::MAX_LEN + 1,
+                "unsupported platform"
+            );
+            let mut ifreq = libc::ifreq {
+                ifr_name: [0; libc::IF_NAMESIZE],
+                ifr_ifru: libc::__c_anonymous_ifr_ifru {
+                    ifru_ifindex: libc::IFF_TAP | libc::IFF_NO_PI,
+                },
+            };
+            for (i, byte) in name.as_ref().as_bytes().iter().enumerate() {
+                // already confirmed that we are ASCII in the InterfaceName contract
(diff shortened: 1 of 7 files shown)


interface-manager/src/interface/tap.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// We are subject to a contract with the kernel.
+    /// </div>
+    #[repr(transparent)]
+    #[derive(Debug, Copy, Clone)]
+    pub(super) struct InterfaceRequest(libc::ifreq);
+
+    use net::interface::InterfaceName;
+    use nix::libc;
+
+    nix::ioctl_write_ptr_bad!(
```
mvachhar, line 40: I looked at the docs here for `nix` but still don't understand what a "bad" ioctl is here?  Is it one that `nix` does not understand?

interface-manager/src/interface/tap.rs
```diff
@@ -0,0 +1,139 @@
...
+        /** Create a tap device */
+        make_tap_device,
+        libc::TUNSETIFF,
+        InterfaceRequest
+    );
+
+    nix::ioctl_write_ptr_bad!(
+        /** Keep the tap device after the program ends */
+        persist_tap_device,
+        libc::TUNSETPERSIST,
```
mvachhar, line 50: Is this correct, who cleans up this device on exit?

interface-manager/src/interface/tap.rs
```diff
@@ -0,0 +1,139 @@
...
+        InterfaceRequest
+    );
+
+    impl InterfaceRequest {
+        /// Create a new `InterfaceRequest`.
+        #[cold]
+        #[tracing::instrument(level = "trace")]
+        pub(super) fn new(name: &InterfaceName) -> Self {
+            assert_eq!(
+                libc::IF_NAMESIZE,
```
mvachhar, line 60: Is there any way to make this fail at compile time since IF_NAME_SIZE and MAX_LEN are constants?  Perhaps use `static_assert`?

interface-manager/src/interface/tap.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// </div>
+    #[repr(transparent)]
+    #[derive(Debug, Copy, Clone)]
+    pub(super) struct InterfaceRequest(libc::ifreq);
+
+    use net::interface::InterfaceName;
+    use nix::libc;
+
+    nix::ioctl_write_ptr_bad!(
+        /** Create a tap device */
```
qmonnet, line 41: Oh no, Fredi's C-style comments are contagious! :scream_cat:

qmonnet (approved):
Nothing to say in addition to what others already raised, looks OK as far as I can tell

interface-manager/src/interface/tap.rs
Fredi-raspall: They are valid rust comments! https://doc.rust-lang.org/reference/comments.html

qmonnet: Yes they are!

I just prefer sticking to a single type, for consistency, that's all. And I've a small preference for the double-slash `//`, which doesn't prevent you from surrounding blocks with comments with additional `/* */` markers to temporarily exclude something from compilation. But that's fine! :slightly_smiling_face:

daniel-noland: This is actually more subtle than you might think.  `/**` and `///` are rustdoc, not just comments.  Technically they are a type of attribute.  In this case I am forcing nix to generate a doc string from the ioctl wrapper it cooks

interface-manager/src/interface/tap.rs
daniel-noland: there are ways to do that.  I wasn't sure we wanted to invoke them, but I can :)

interface-manager/src/interface/tap.rs
daniel-noland: The goal is to have the interface manager gain a delete method (via netlink) in a future commit.  This is a good point, but I don't think I can actually remove these via ioctl, and no need for more of the ugly things in any case

interface-manager/src/interface/tap.rs
daniel-noland: I did.  This ioctl is quite old and basically encodes the length of the structure in the ioctl identifier. This is unfortunate, but it can't be fixed :shrug:

qmonnet: (Sorry I was showing the link to Manish :sweat_smile:)

interface-manager/src/interface/tap.rs
mvachhar: I think that would be a good change here.

interface-manager/src/interface/tap.rs
```diff
@@ -0,0 +1,234 @@
...
+        if orig_len < bytes_read.get() {
+            error!("buffer too small: {orig_len} < {bytes_read}");
+            return Err(tokio::io::Error::new(
+                tokio::io::ErrorKind::UnexpectedEof,
+                "unexpected EOF on tap device",
+            ));
+        }
+        #[allow(clippy::expect_used)] // memory integrity requirement already checked
+        buf.trim_from_end(orig_len - bytes_read.get())
+            .expect("failed to trim buffer: illegal memory manipulation");
```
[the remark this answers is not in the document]
daniel-noland: I do not believe this is a correct interpretation of our API

interface-manager/src/interface/tap.rs
```diff
@@ -0,0 +1,234 @@
...
+            Err(err) => {
+                error!("nonsense sized buffer: {}", buf.as_ref().len());
+                return Err(tokio::io::Error::other(err));
+            }
+        };
+        if orig_len < bytes_read.get() {
+            error!("buffer too small: {orig_len} < {bytes_read}");
+            return Err(tokio::io::Error::new(
+                tokio::io::ErrorKind::UnexpectedEof,
+                "unexpected EOF on tap device",
```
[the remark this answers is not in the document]
daniel-noland: reasonable

PR #725, opened 2025-07-21
Moves frr rendering to routing crate & recover fake-frr agent for testing

Fredi-raspall:


routing/src/frr/test.rs
```diff
@@ -0,0 +1,196 @@
...
+                Err(ref e) if e.kind() == ErrorKind::WouldBlock => {
+                    continue;
+                }
+                Err(e) => return Err(e.to_string()),
+            };
+        }
+        Ok(data.into())
+    }
+
+    /// Receive a message made of |length|genid|data. This applies to both requests and responses.
```
qmonnet, line 38:
```suggestion
    /// Receive a message made of |length|genid|data|. This applies to both requests and responses.
```

qmonnet (approved):
Looks good

PR #729, opened 2025-07-22
set physical MTU

daniel-noland:


qmonnet:
This PR needs a rebase

mvachhar (changes requested):
If my comments about the commit data in the early commits being improperly comingled are right, it is probably worth a few minutes to split them out, otherwise this looks good to me.

interface-manager/src/interface/netdevsim.rs
qmonnet, line 1: Your commit title (` feat(interface-manager)!: netdevsim part 1`) mentions a breaking change, what is it exactly? Please expand in commit description.

mgmt/Cargo.toml
```diff
@@ -12,6 +12,7 @@ required-features = ["bolero"]
 
 [features]
 default = []
+fake-pci-as-netdevsim = ["interface-manager/netdevsim"]
```
qmonnet, line 15: This is not used in this PR, is it?

interface-manager/src/interface/netdevsim.rs
qmonnet, line 1: This file seems to not be included in your code. Aren't you missing a `#[cfg(feature = "netdevsim")]\n use netdevsim` in your `interface-manager/src/interface/mod.rs`??

(I pulled your branch, broke syntax in the file, but `just cargo build --features netdevsim -p dataplane-interface-manager` still works.)

I also note this file is very similar to net/src/interface/physical/netdevsim.rs, could this be a leftover committed by mistake after you moved the code from `interface-manager` to `net`?

qmonnet (changes requested):

mgmt/Cargo.toml
daniel-noland: I think I only partially understand the question.

This feature would, if activated, cause the `interface-manager/netdevsim` feature to be activated.  I introduce it in this PR because allowing "fake" pci devices in a future PR is part of the objective.

The intended functionality does not yet exist.  I can delay the introduction of this feature until after netdevsim functionality is more developed if you prefer.

net/src/interface/physical/mod.rs
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
daniel-noland: fixed I think

net/src/interface/physical/switch.rs
```diff
@@ -5,23 +5,21 @@ use arrayvec::ArrayVec;
 use serde::{Deserialize, Serialize};
 use std::fmt::{Debug, Display, Formatter, LowerHex};
 
-const SWITCH_ID_MAX_LEN: usize = 32;
```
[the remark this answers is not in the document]
daniel-noland: fixed

net/src/interface/physical/bus.rs
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Ord, PartialOrd, Deserialize, Serialize)]
+pub enum Bus {
```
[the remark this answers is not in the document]
daniel-noland: Yeah no idea how that happened.  Fixed

interface-manager/src/interface/netdevsim.rs
daniel-noland: Good catch.  The include commit got lost somehow in a refactor and the tests still passed because the feature is (and should be) off by default.

mgmt/tests/reconcile.rs
```diff
@@ -258,16 +259,15 @@ async fn reconcile_demo() {
...
+                | InterfacePropertiesSpec::Tap => {}
                 InterfacePropertiesSpec::Vtep(props) => {
                     req.vteps.try_insert(props.clone()).unwrap();
                 }
                 InterfacePropertiesSpec::Vrf(props) => {
                     req.vrfs.try_insert(props.clone()).unwrap();
                 }
-                InterfacePropertiesSpec::Pci(props) => {
-                    req.pci_netdevs.try_insert(props.clone()).unwrap();
-                }
```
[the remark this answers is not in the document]
daniel-noland: It is correct I think.  There is simply nothing to index in that environment in terms of PCI devices.

net/src/interface/physical/switch.rs
```diff
@@ -5,23 +5,21 @@ use arrayvec::ArrayVec;
...
     Empty,
-    #[error("Maximum length of an ESwitchId is 32 bytes, received {0} bytes")]
+    #[error("Maximum length of an ESwitchId is {MAX} bytes, received {0} bytes", MAX = SwitchId::MAX_LEN)]
     InvalidLength(usize),
 }
 
 impl SwitchId {
     /// The maximum length of an [`SwitchId`] in bytes
-    pub const MAX_LEN: usize = SWITCH_ID_MAX_LEN;
+    pub const MAX_LEN: usize = 128;
```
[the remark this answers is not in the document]
daniel-noland: Split out now

mgmt/src/vpc_manager/mod.rs
```diff
@@ -325,9 +325,53 @@ impl TryFrom<&InternalConfig> for RequiredInformationBase {
         let mut vteps = MultiIndexVtepPropertiesSpecMap::default();
         let mut associations = MultiIndexInterfaceAssociationSpecMap::default();
         for config in internal.vrfs.iter_by_tableid() {
-            if config.default {
-                trace!("skipping default config: {config:?}");
-                continue;
+            for iface in config.interfaces.values() {
```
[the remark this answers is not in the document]
daniel-noland: split out of orig commit and moved into one about taps

net/Cargo.toml
```diff
@@ -24,6 +25,7 @@ ordermap = { workspace = true, features = ["std"] }
 serde = { workspace = true, features = ["derive", "std"] }
 thiserror = { workspace = true }
 tracing = { workspace = true }
+id = { workspace = true }
```
qmonnet, line 28: Should this dependency be marked as optional? It's only pulled if we build with feature `netdevsim`, isn't it?

interface-manager/src/interface/mod.rs
```diff
@@ -11,6 +11,7 @@ mod tap;
 mod vrf;
 mod vtep;
 
+
```
qmonnet: Nit: leftover blank line

qmonnet (approved):
Looks good, thanks! Just some non-blocking small nits/suggestions below.

interface-manager/src/interface/mod.rs
qmonnet: I see you fix it in the last commit. Consider squashing with `feat(interface-manager)!: netdevsim part 1`.

mgmt/Cargo.toml
qmonnet: > This feature would, if activated, cause the `interface-manager/netdevsim` feature to be activated.

I understand that, thanks. I wanted to make sure my understanding was correct and the feature is not used yet.

> The intended functionality does not yet exist. I can delay the introduction of this feature until after netdevsim functionality is more developed if you prefer.

That's OK, no need to defer if you know you'll use it soon.

We don't have docs for the different packages' features yet, but how about adding a one-line comment in the Cargo.toml to at least summarise what this feature does?

net/Cargo.toml
daniel-noland: oh, that does make some sense.  LTO will remove all traces of the dependency if it isn't linked in at all, so technically we already have that met for release builds.  But it is likely a good idea anyway just to document why we need the ID package

Fredi-raspall (approved):
LGTM @daniel-noland !

mvachhar (approved):

PR #771, opened 2025-08-08
Fix VPC manage vtep assumption

Fredi-raspall:
I'm not very fond of this fix, but it should work.
Please refer to the commit for the explanation of the issues.


qmonnet:
I'm not sure what the issue in CI is, but it appears that https://github.com/githedgehog/dataplane/pull/781 fixes it. So CI should pass again if you rebase your PR on top of `main`.

PR #671, opened 2025-07-05
Simplify code to set up vtep config

Fredi-raspall:
This targets https://github.com/githedgehog/dataplane/pull/670, which should be merged first.

PR #772, opened 2025-08-08
Simplify vtep config

Fredi-raspall:
This replaces https://github.com/githedgehog/dataplane/pull/671
THere is one extra simplification that we may want to add.


qmonnet (approved):
Seems OK to me

PR #776, opened 2025-08-11
Add new "concurrency" crate for loom/shuttle tests

qmonnet:
Introduce a new "concurrency" tests that exposes:

- The concurrency items from std::sync, when no feature is selected,
- The items from loom::sync, when the "loom" feature is selected,
- The items from shuttle::sync, when the "shuttle" feature is selected.

The "loom" and "shuttle" features are mutually exclusive.

This is in preparation for running concurrency tests with loom/shuttle.

---

Crate name - I initially thought of naming this `sync`, but:
- `sync` will probably be better suited for when we try to sync states between several gateways
- We could have other submodules from `std`/`loom`/`shuttle` shared in this crate, hence `concurrency` which encompasses the different potential aspects. I remain open to alternative suggestions though.


qmonnet:
Clippy complains, because it tries to compile the crate with `--all-features`, although the features are mutually exclusive. This is unfortunate. I don't really want to remove `--all-features` and name them all in the `justfile` instead.

Another solution could be to use configuration attributes (instead of crate features) to build this, but I don't know how well it integrates with our CI infra. Link: https://docs.rs/loom/latest/loom/#writing-tests

qmonnet:
Or maybe I should also gate it behind `test`, and fall back on the regular implementation otherwise :thinking:

mvachhar (approved):

daniel-noland (approved):
LGTM

PR #787, opened 2025-08-14
fix(concurrency-macros): Fix shuttle/loom macro when all features are on

qmonnet:
When both `shuttle` and `silence_clippy` are turned on in the concurrency crate, then the `with_shuttle` macro is never defined. If another crate attempts to use `#[concurrency_mode(shuttle)]` in that case, the code fails to compile due to the missing macro. The same happens for the `with_loom` macro, under similar conditions.

Fix this by defining the `empty` version of the macros when shuttle/loom are turned off, _or_ when they're on but `silence_clippy` is on as well.


mvachhar (changes requested):
I don't think this is the right fix as I don't want to condition things just on `silence_clippy` but instead that all three features are enabled.  See #802 for an alternate fix.

Also the component name in the commit message is wrong, it is concurrency not concurreny-macros.

PR #802, opened 2025-08-14
fix(concurrency): Define with_shuttle and with_loom macros always

mvachhar:


mvachhar:
Closed in favor of #787

Issue #769, opened 2025-08-08
NAT: Adjust VpcExpose validation to allow IP lists of different sizes


qmonnet:
As part of the work to allow _stateless_ NAT support in the Gateway, we introduced a restriction on the `VpcExpose` objects that we work with, enforced at validation time: we make sure that the list of IP addresses internal to the VPC and used in the `VpcExpose` contains exactly the same number of addresses as the list of publicly-exposed IP addresses for this `VpcExpose` (put differently: we make sure that `ips_sizes - nots_sizes == as_range_sizes - not_as_sizes)`. [This is enforced here](https://github.com/githedgehog/dataplane/blob/52be68f9842f9685efb8015133c55fb87201ee9c/config/src/external/overlay/vpcpeering.rs#L151~L161). This restriction is to ensure that we can operate a 1:1 mapping between the initial and target lists of addresses.

Now that we're adding support for _stateful_ NAT, we can and should support `VpcExpose` that don't have this restriction: for stateful NAT, there's no 1:1 mapping, and it doesn't matter if the sizes of both lists are not the same.

We should move this restriction. The most likely place to have it is the code where we build the configuration for stateless NAT: we should perform additional verifications here, before building the NAT tables.

PR #803, opened 2025-08-15
Pr/qmonnet/mv valid nat

qmonnet:
As part of the validation step for all `VpcExpose` objects in a `VpcTable`, we ensure that, for each expose containing prefixes for NAT, the number of IP addresses defined by the initial list of prefixes is equal to the number of IP addresses defined by the target list of prefixes. This is to ensure that we can apply a 1:1 mapping between IP addresses of each set.

This validation step is essential for stateless NAT, however, we cannot apply it indistinctively when processing the Gateway config, because it would prevent users from having IP sets of different sizes, for example if they're using stateful NAT rather than stateless NAT.

As a first step towards a more flexible approach, move the validation step from the validation functions for the `VpcTable` in the `config` crate to the `nat` crate, and call it from `mgmt` right before building the stateless NAT tables.

A follow-up step will be required to conditionally validate the `VpcTable` for stateless NAT and build the NAT tables in `mgmt`, once we have an API to indicate whether or not stateless NAT is desired.

This PR also contains to other clean-up commits related to the same code section.

Based on top of #697
Relates to, but does not fully addresses, #769


qmonnet:
> A follow-up step will be required to conditionally validate the `VpcTable` for stateless NAT and build the NAT tables in `mgmt`, once we have an API to indicate whether or not stateless NAT is desired.

@mvachhar There's some discussion required, here. For a given configuration passed from the user, how do we know whether the user expects to have stateless or stateful NAT? I guess it also depends on what pipeline stages are also aligned to process the packets, but then how are we supposed to know whether we should validate `VpcExpose` objects, and build the NAT tables, for stateless NAT or not?

mvachhar (approved):

PR #806, opened 2025-08-15
Add concurrency::thread, clean up NAT test organization and use concurrency::thread

mvachhar:


nat/src/stateful/apalloc/test_alloc.rs
```diff
@@ -446,85 +361,167 @@ mod tests_shuttle {
...
+                    assert!(res.src.is_some());
+                    assert!(res.dst.is_some());
+                    assert!(res.return_src.is_some());
+                    assert!(res.return_dst.is_some());
+                }));
+
+                let _results: Vec<()> = handles
+                    .into_iter()
+                    .map(|handle| handle.join().unwrap())
+                    .collect();
```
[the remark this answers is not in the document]
mvachhar: I don't want to change @qmonnet's code too much here so I just added the `_`.

daniel-noland (approved):
LGTM

nat/Cargo.toml
```diff
@@ -5,6 +5,9 @@ edition = "2024"
 publish = false
 license = "Apache-2.0"
 
+[features]
+shuttle = ["concurrency/shuttle", "dep:shuttle"]
```
qmonnet, line 9: @mvachhar The `shuttle` feature is unused at the moment, right? Why introduce it in a a clean-up commit, without mentioning it (or its future use case) in the commit/PR descriptions?

qmonnet: @mvachhar Thinking of this again, I still have doubts. What's the point of this feature? Sure, if you build with `cargo build -F shuttle`, this will pull `concurrency/shuttle` and make the “concurrency mode” be `shuttle`.

_But_ if you use `cargo build -F shuttle`, you _already_ build the `concurrency` crate with feature `shuttle` _anyway_, so what does the local `shuttle` feature bring? Why not keep just one feature in `concurrency`? It's not as if we picked features individually for each crate on the command line, is it?

PR #824, opened 2025-09-03
style: Replace tabs with spaces

qmonnet:
We use speaces nearly everywhere for indentation in the project, meaning most of us have our editors configured to use spaces, accordingly. As a result, each time I add a new crate in the toplevel Cargo.toml, I need to rework my change to convert the spaces to tabs for consistency...

Let's just clean up these few tabs that are here for no apparent reason, and use spaces instead. That's just three files to modify, two of which had inconsistent indent, with that.

You can hide whitespace changes for review.


mvachhar (approved):
LGTM.  Is there a way to auto format this in time and non-rs files?

qmonnet:
> Is there a way to auto format this in time and non-rs files?

For all file types at once, I'm not sure. I did a simple `rg '\t'` though, something like that would be easy to add. I didn't want to spend more time on that at the moment though.

PR #844, opened 2025-09-12
Pipeline replication

Fredi-raspall:
* This is incomplete since the Packetdumper NF has to be fixed.
* Note that this does not  implement multiple packet workers but is a first step towards that goal.


qmonnet:

PR #849, opened 2025-09-15
ci: Use GitHub App token for bump workflow

qmonnet:
We have a GitHub App set up for the dataplane repository, so that updating the dpdk-sys repository can automatically create a Pull Request for the dataplane repository to update the changes to the compile-env image. We should be able to reuse this App when creating Pull Requests to bump dependencies in the dataplane repository: the expected advantage is that the generated Pull Requests should now trigger CI on creation, without the need for reviewers to close and re-open the Pull Request to manually trigger it.

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -30,6 +30,13 @@ jobs:
   cargo-upgrades:
     runs-on: "lab"
     steps:
+      # Use a GitHub App token so that the generated PR can trigger CI
+      - name: "Generate GitHub App token"
+        id: "app-token"
+        uses: "actions/create-github-app-token@v2"
+        with:
+          app-id: "${{ secrets.DP_APP_ID }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/bump.yml
```diff
@@ -30,6 +30,13 @@ jobs:
   cargo-upgrades:
     runs-on: "lab"
     steps:
+      # Use a GitHub App token so that the generated PR can trigger CI
+      - name: "Generate GitHub App token"
+        id: "app-token"
+        uses: "actions/create-github-app-token@v2"
```
mvachhar, line 36: Does creating this token increase permissions beyond triggering CI?  What roles does this token give the CI job?

mvachhar: Looks like permission are restricted to what the github app can do so this is fine.

mvachhar (approved):

PR #850, opened 2025-09-15
feat: multithreading kernel driver

sergeymatov:
Workers are spinned up by driver startup, using pipeline
fabric to create separate workflow for each thread. TX and RX
done by master thread.

### DIFF
--- a/dataplane/src/args.rs
+++ b/dataplane/src/args.rs
@@ -41,7 +41,7 @@ pub(crate) struct CmdArgs {
     #[arg(
         long,
         value_name = "N",
-        help = "Number of worker threads for the kernel driver (defaults to be one worker)"
+        help = "Number of worker threads for the kernel driver (defaults to one worker)"
     )]
     num_workers: Option<usize>,
 
@@ -102,15 +102,13 @@ impl CmdArgs {
 
     /// Number of worker threads to use for the kernel driver.
     /// Falls back to 1 if not specified or invalid.
-    pub fn num_workers(&self) -> usize {
+    pub fn kernel_num_workers(&self) -> usize {
         match self.num_workers {
             Some(n) if n > 0 => n,
             _ => 1,
         }
     }
-
-    #[allow(clippy::unused_self)]
-    pub fn kernel_params(&self) -> Vec<String> {
+    pub fn kernel_interfaces(&self) -> Vec<String> {
         self.interface.clone()
     }
 
--- a/dataplane/src/args.rs
+++ b/dataplane/src/args.rs
@@ -41,9 +41,10 @@ pub(crate) struct CmdArgs {
     #[arg(
         long,
         value_name = "N",
-        help = "Number of worker threads for the kernel driver (defaults to one worker)"
+        default_value_t = 1,
+        help = "Number of worker threads for the kernel driver"
     )]
-    num_workers: Option<usize>,
+    num_workers: usize,
 
     /// gRPC server address (IP:PORT for TCP or path for UNIX socket)
     #[arg(
@@ -103,10 +104,7 @@ impl CmdArgs {
     /// Number of worker threads to use for the kernel driver.
     /// Falls back to 1 if not specified or invalid.
     pub fn kernel_num_workers(&self) -> usize {
-        match self.num_workers {
-            Some(n) if n > 0 => n,
-            _ => 1,
-        }
+        self.num_workers
     }
     pub fn kernel_interfaces(&self) -> Vec<String> {
         self.interface.clone()
--- a/dataplane/src/args.rs
+++ b/dataplane/src/args.rs
@@ -37,6 +37,14 @@ pub(crate) struct CmdArgs {
     #[arg(long, value_name = "name of kernel interface")]
     interface: Vec<String>,
 
+    /// Number of worker threads for the kernel driver.
+    #[arg(
+        long,
+        value_name = "N",
+        help = "Number of worker threads for the kernel driver (defaults to be one worker)"
+    )]
+    num_workers: Option<usize>,
+
     /// gRPC server address (IP:PORT for TCP or path for UNIX socket)
     #[arg(
         long,
@@ -89,6 +97,15 @@ impl CmdArgs {
         match &self.driver {
             None => "dpdk",
             Some(name) => name,
+        }
+    }
+
+    /// Number of worker threads to use for the kernel driver.
+    /// Falls back to 1 if not specified or invalid.
+    pub fn num_workers(&self) -> usize {
+        match self.num_workers {
+            Some(n) if n > 0 => n,
+            _ => 1,
         }
     }
 
--- a/dataplane/src/drivers/kernel.rs
+++ b/dataplane/src/drivers/kernel.rs
@@ -281,6 +281,14 @@ impl DriverKernel {
             }
         };
 
+        let num_worker_chans = to_workers.len();
+        assert!(num_worker_chans != 0, "No worker channels available!");
+        if num_worker_chans != num_workers {
+            warn!(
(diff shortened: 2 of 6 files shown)


dataplane/src/drivers/kernel.rs
```diff
@@ -42,20 +48,21 @@ pub struct Kif {
...
-        if let Err(e) = sock.bind(name) {
-            error!("Failed to bind to interface '{name}'");
-            return None;
-        }
+    fn new(ifindex: u32, name: &str, token: Token) -> io::Result<Self> {
+        let mut sock = RawPacketStream::new().map_err(|e| {
+            error!("Failed to open raw sock for interface {name}: {e}");
+            e
+        })?;
+        let _ = sock.set_non_blocking();
```
mvachhar, line 63: The let _ here is unecessary unless there is some lifetime issue I am not aware of.

dataplane/src/drivers/kernel.rs
```diff
@@ -42,20 +48,21 @@ pub struct Kif {
...
+        sock.bind(name).map_err(|e| {
+            error!("Failed to bind to interface '{name}': {e}");
```
mvachhar, line 65: This is an abuse of map_err since there is actually no mapping going on.  You could use:
```rust
sock.bind(name).inspect_err(|e| error!("Failed to open raw sock for interface {name}: {e}"))?
```

dataplane/src/drivers/kernel.rs
```diff
@@ -75,51 +82,51 @@ pub struct KifTable {
...
-            self.by_token.insert(token, interface);
-            self.next_token += 1;
-            debug!("Successfully registered interface '{name}' with token {token:?}");
-        }
+        let interface = Kif::new(ifindex, name, token)?;
+        let mut source = SourceFd(&interface.raw_fd);
+        self.poll
+            .registry()
+            .register(&mut source, token, Interest::READABLE)
+            .map_err(|e| {
```
mvachhar, line 108: Use
```rust
.inspect_err(|e| {
  error!("Failed to register interface '{name}': {e}");
})?;

File: dataplane/src/drivers/kernel.rs [hunk 4]
```diff
@@ -130,100 +137,217 @@ fn get_interface_ifindex(interfaces: &[Interface], name: &str) -> Option<u32> {
...
+        for wid in 0..w {
+            let (tx_to_worker, rx_to_worker) = chan::bounded::<Packet<TestBuffer>>(8192);
+            to_workers.push(tx_to_worker);
+
+            let tx_out = tx_from_workers.clone();
+            let setup = setup_pipeline.clone();
+
+            let builder = std::thread::Builder::new().name(format!("dp-worker-{wid}"));
+            let handle_res = builder.spawn(move || {
+                let mut pipeline = (setup)();
```
mvachhar, line 233: Are the parens around setup needed here?

dataplane/src/drivers/kernel.rs
```diff
@@ -130,100 +137,217 @@ fn get_interface_ifindex(interfaces: &[Interface], name: &str) -> Option<u32> {
...
+                loop {
+                    match rx_to_worker.recv() {
+                        Ok(pkt) => {
```
mvachhar: Can you create an issue to batch packets for the kernel driver?  No need to fix it in this PR, but going packet by packet kind of sucks.  We should be able to send a collection of packets over instead, though this has its own performance considerations.

dataplane/src/drivers/kernel.rs
```diff
@@ -130,100 +137,217 @@ fn get_interface_ifindex(interfaces: &[Interface], name: &str) -> Option<u32> {
...
+                                }
+                            }
+                            Err(e) => error!("Serialize failed: {e:?}"),
+                        }
+                    } else {
+                        warn!("TX drop: unknown oif {}", oif_id);
+                    }
+                } else {
+                    // No oif set; nothing to send (likely dropped upstream).
+                    debug!("No oif in packet meta; dropping");
```
mvachhar, line 320: Is this an error?  Doesn't the packet have a drop flag we can check to know if it is supposed to be dropped?

mvachhar (changes requested):

mvachhar:
I added the `ci+vlab` label so that we test the end-to-end correctness of these changes.  Any time we change the datapath or relevant config we should add the `ci+vlab` label so that we run the vlab tests.  Sadly, when adding the tag you either have to push to the PR or close and reopen the PR to trigger those tests.

dataplane/src/drivers/kernel.rs
sergeymatov: I thought it was must use thingy, but I may be wrong. Let me check and fix

dataplane/src/drivers/kernel.rs
sergeymatov: I will create a separate issue

dataplane/src/drivers/kernel.rs
```diff
@@ -130,100 +136,223 @@ fn get_interface_ifindex(interfaces: &[Interface], name: &str) -> Option<u32> {
...
+    /// Compute a **symmetric** worker index for a parsed `Packet` using a bidirectional flow key.
+    #[must_use]
+    pub fn compute_worker_idx(pkt: &Packet<TestBuffer>, workers: usize) -> usize {
+        let n = workers.max(1);
+
+        // Prefer symmetric flow-key hash (A<->B go to the same bucket)
+        if let Ok(flow_key) = FlowKey::try_from(Bidi(pkt)) {
+            let mut h = DefaultHasher::new();
+            flow_key.hash(&mut h);
+            let hv = usize::try_from(h.finish()).unwrap_or(0);
```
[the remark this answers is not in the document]
mvachhar: I would do:
```
h.finish() as usize // The default behavior here should be truncate
```
And suppress the clippy warning.

dataplane/src/drivers/kernel.rs
```diff
@@ -130,100 +136,223 @@ fn get_interface_ifindex(interfaces: &[Interface], name: &str) -> Option<u32> {
...
+            if n.is_power_of_two() {
+                hv & (n - 1)
```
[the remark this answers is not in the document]
mvachhar: Looks like this is done.

Fredi-raspall:
@sergeymatov I've pushed a commit. Feel free to: drop / squash.

dataplane/src/drivers/kernel.rs
```diff
@@ -274,14 +273,14 @@ impl DriverKernel {
...
         // Spawn workers
-        let (to_workers, from_workers) = match Self::spawn_workers(workers, setup_pipeline) {
+        let (to_workers, from_workers) = match Self::spawn_workers(num_workers, setup_pipeline) {
             Ok(chans) => chans,
             Err(e) => {
                 error!("Failed to start workers: {e}");
                 return;
             }
         };
-        let buckets = to_workers.len().max(1);
```
mvachhar: I disagree with this change as if `num_workers` is different than `to_workers.len()` we'll panic on the index of `to_workers` below.  I'll push a fix since I don't quite understand why the `max(1)` is needed either.

dataplane/src/args.rs
```diff
@@ -41,7 +41,7 @@ pub(crate) struct CmdArgs {
     #[arg(
         long,
         value_name = "N",
-        help = "Number of worker threads for the kernel driver (defaults to be one worker)"
```
mvachhar: Why are we not using `default_value_t` here instead of a manual string?

dataplane/src/drivers/kernel.rs
mvachhar: Fixed.

dataplane/src/args.rs
mvachhar: Moved to use default_value_t.

PR #858, opened 2025-09-19
chore(README.md): Add some settings.json stuff for Zed

mvachhar:


README.md
```diff
@@ -170,6 +170,34 @@ To do this, add the following to your `.vscode/settings.json` file:
 },
 ```
 
+## Zed Setup
+
+Because this repository uses a custom sysroot with custom libraries and binaries, you need to set up your environment accordingly.
```
qmonnet, line 175: What about having an “IDE configuration” section where we put this sentence, then subsections for VSCode and Zed?

README.md
```diff
@@ -170,6 +170,34 @@ To do this, add the following to your `.vscode/settings.json` file:
...
+Here is a suggested configuration for Zed that you can put in ".zed/settings.json":
+```json
+{
```
qmonnet, line 178: Not strictly related to LSP config, but could we please also fix the preferred line length (to align with rustfmt)?

```suggestion
{
   "preferred_line_length": 100,
```

qmonnet (commented):
If we add a dedicated section in the README.md, why not add the config file to the repo under `.zed` directly, as we do for `.vscode`?

README.md
mvachhar: I am going to change this to format on save using rust so this shouldn't be needed.  I'm not a fan of the behavior when typing longer lines.  Better to format on save.

README.md
mvachhar: Done.

README.md
mvachhar: Added the language server formatting on save.

qmonnet: > I am going to change this to format on save using rust so this shouldn't be needed. I'm not a fan of the behavior when typing longer lines.

What behaviour? This doesn't wrap lines when you type them, except if you turn `soft_wrap` on. I don't expect people to have it on, and if they do, then we want to wrap at 100 char and not 80. My use case is when I use the Vim binding to rewrap a comment, I want it to wrap on 100 char.

> Better to format on save.

Comments aren't reformatted by autoformat on save.

Also, I'm pretty sure formatting on save is already on by default?

README.md
qmonnet: I'd still move this sentence before the VSCode subsection, to avoid repeating the same thing twice. (Not blocking, though)

PR #865, opened 2025-09-24
fix(dataplane): Validate num workers to be in range(1..=64)

mvachhar:


mvachhar:
> Hum.. I just approved this PR and have the impression that by doing that the vlab test was cancelled?

It looks like the vlab tests ran and passed.

PR #867, opened 2025-09-24
ci: add cargo audit workflow

pau-hedgehog:


mvachhar (changes requested):

pau-hedgehog:
> Thanks @pau-hedgehog, I think we're already covered for this repo.

Ok, then I'm closing this. Sorry for wasting your time

> We've got `cargo deny` set up in CI, and my understanding is that it checks advisories like `cargo audit` does (plus some other things). The two advisories that you found as warnings are known and explicitly ignored in `deny.toml`. Is there anything more that you're trying to get via `cargo audit`?

After updating the govulncheck in our Go repos I was doing my first incursion in the Rust world to see if there was something equivalent. Next time I'll spend some more time researching and/or asking. Thanks

qmonnet:
No problem at all, thanks for looking into this!

PR #875, opened 2025-09-29
ci: Pull testn/n-vm image for bump.yml workflow

qmonnet:
The image is now necessary to have tests pass. We pull it for the dev.yml workflow, we also need to get it for the bump.yml workflow.


qmonnet:
See also changes proposed in https://github.com/githedgehog/dataplane/pull/868/files

PR #876, opened 2025-09-29
refactor(net): Refactor parser calling for L3 -> L4 parsing

qmonnet:
The code for calling parsers for IP, IP extensions, IP authentication is pretty much the same, repeated many times accross different submodules. Let's refactor it, to make future changes easier. We introduce a new ParseTarget trait that we implement (with default methods) for the different Ip\*Next types, and call its methods everywhere necessary.

We don't really get a shorter line count at the end of the day, because we need to implement the From trait for all possible variants for the Ip\*Next enums each time, but it remains clearer than the chunk we avoid duplicating.


qmonnet:
Tests are expected to fail until #868 gets in

net/src/parse.rs
```diff
@@ -59,6 +61,36 @@ pub(crate) trait ParsePayloadWith {
     fn parse_payload_with(&self, param: &Self::Param, cursor: &mut Reader) -> Option<Self::Next>;
 }
 
+pub(crate) trait ParseTarget {
+    fn call_parser<T>(cursor: &mut Reader) -> Option<Self>
```
daniel-noland, line 65: I need to think about this API a bit.

I think that it makes more sense to `impl<R> ParseTarget for R where R: Reader`

daniel-noland: and I would rename `call_parser` to `try_parse`

net/src/parse.rs
```diff
@@ -59,6 +61,30 @@ pub(crate) trait ParsePayloadWith {
...
+pub trait ParseHeader {
+    fn parse_header<T: Parse, O: From<T>>(&mut self) -> Option<O>;
+    fn parse_header_with<T: ParseWith, O: From<T>>(&mut self, param: T::Param) -> Option<O>;
+}
+
+impl ParseHeader for Reader<'_> {
+    fn parse_header<T: Parse, O: From<T>>(&mut self) -> Option<O> {
+        self.parse::<T>()
+            .map_err(|e| {
+                debug!("failed to parse {}: {e:?}", core::any::type_name::<T>());
```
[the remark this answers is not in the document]
qmonnet: Addressing in https://github.com/githedgehog/dataplane/pull/886

PR #880, opened 2025-09-29
tracing tune up

daniel-noland:
~~On top of #879~~

^^ rebased

### DIFF
--- a/dataplane/Cargo.toml
+++ b/dataplane/Cargo.toml
@@ -35,9 +35,9 @@ routing = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 stats = { workspace = true }
 tokio = { workspace = true }
-tracing = { workspace = true }
-tracing-subscriber = { workspace = true, features = ["default"] }
 tracectl = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true, default-features = true }
 vpcmap = { workspace = true }
 
 [dev-dependencies]


dataplane/src/packet_processor/ipforward.rs
```diff
@@ -66,7 +66,7 @@ impl IpForwarder {
 
         /* get destination ip address */
         let Some(dst) = packet.ip_destination() else {
-            error!("{nfi}: Failed to get destination ip address for packet");
+            debug!("{nfi}: Failed to get destination ip address for packet");
```
Fredi-raspall, line 68: I see the intent here (reducing verbosity), but to me the changes in this commit will harm more than anything.
If something is an ERROR or a warning, we should log it as such, because when displayed the severity/loglevel tells you what it is. We now have the option to completely disable traces in the pipeline. We may want to adjust the levels in the source code if some errors get very frequently logged; and we still have the option to rate-limit them for instance. However, in this particular example, I'd say you'd want to see the log as an ERROR in neon-lights since failing to get the destination ip address should be the symptom of something going really wrong.

dataplane/src/packet_processor/ingress.rs
```diff
@@ -19,7 +19,7 @@ use routing::interfaces::interface::{Attachment, IfState, IfType, Interface};
 use tracectl::trace_target;
 trace_target!("ingress", LevelFilter::WARN, &["pipeline"]);
 
-#[allow(unused)]
+#[derive(Debug)]
```
Fredi-raspall, line 22: Why do we want to derive Debug?

dataplane/src/packet_processor/ingress.rs
```diff
@@ -69,30 +69,39 @@ impl Ingress {
...
     }
 
+    #[tracing::instrument(level = "trace")]
     fn interface_ingress_eth_bcast<Buf: PacketBufferMut>(
         &self,
-        _interface: &Interface,
+        interface: &Interface,
         packet: &mut Packet<Buf>,
     ) {
         let nfi = self.name();
```
Fredi-raspall, line 95: If the purpose of this change is not taking references that may not be used if the tracing is disabled, I believe you missed this one?

Fredi-raspall (changes requested):
I like some of the changes in this PR (even though I'm unsure about the urgency of them compared to other stuff). The only objection I have has to do with changing the log-levels in the traces. Happy to discuss in any case.

dataplane/src/packet_processor/ipforward.rs
mvachhar: If this can be triggered by a carefully crafted user packet, then this should *not* be error level as then malicious senders can cause logs to fill up and DOS the gateway.

daniel-noland: if the user can drive us into any situation where ip address is missing then this message is a dos vector

dataplane/src/packet_processor/ipforward.rs
```diff
@@ -80,19 +80,19 @@ impl IpForwarder {
 
         /* Read-only access to the fib table */
         let Some(fibtr) = self.fibtr.enter() else {
-            error!("{nfi}: Unable to lookup fib for vrf {vrfid}");
```
daniel-noland, line 83: this can likely remain an `error`.  If the user is able to give us messages for vrfs which don't exist then we have a more fundamental problem than DoS

daniel-noland: I think warn is fair

dataplane/src/packet_processor/ipforward.rs
Fredi-raspall: Well, those packets (a packet with ethertype=IP and no IP header) should not even make it here, right?

dataplane/src/packet_processor/ipforward.rs
Fredi-raspall: Fine with me!

Fredi-raspall (changes requested):
Daniel, thanks for addressing my comments. There's one change, though, that I believe is a regression. See https://github.com/githedgehog/dataplane/pull/880#discussion_r2394378997

dataplane/src/packet_processor/ipforward.rs
daniel-noland: sure, but `error!` is intended for serious application level errors.  This is just what is happening if we happen to get invalid data.

Or are you concerned that we might manipulate the data into an invalid shape by accident (which, I suppose, would promote this to `error!` in truth)

dataplane/src/packet_processor/ingress.rs
mvachhar: I'm not sure if we use it now, but I don't think it does any harm if we have it and use it in the future right?

Fredi-raspall: It does not harm I guess (other than increasing build times?) :shrug: . I was just curious if there was a specific reason.

dataplane/src/packet_processor/ipforward.rs
Fredi-raspall: As discussed, I believe that to address your concerns (DoS) we should clearly delimit where it is safe to log "issues" and where not. In my mind, if I have an Ip packet and can't retrieve the ip address:
  * either there is a bug in the getter (unlikely), and that should be error
  * or the packet is malformed, in which case it should not be a valid packet and make it here (this is the IP-forward stage).
  
I agree that the parser may not know up to which "layer" it should validate a packet. But we should decide at which point packets can be considered malformed/well-formed so that we can "reasonably" tag logs with the right severity and at the same time safely be protected against DoS.

daniel-noland: fair enough.  It is left at error for now

daniel-noland:
> Daniel, thanks for addressing my comments. There's one change, though, that I believe is a regression. See [#880 (comment)](https://github.com/githedgehog/dataplane/pull/880#discussion_r2394378997)

I think this is all sorted now

PR #916, opened 2025-10-16
Fixes for ICMP Error messages processing

qmonnet:
- **chore(net): Use more .unwrap() in ICMP Error msg's Bolero generator**
- **fix(net): Fix DestinationUnreachable ICMP message Bolero generator**
- **fix(net): Do not return empty inner network headers for ICMP generators**
- **fix(net): Pass the right buffer slice when parsing Icmp4/Icmp6**
- **fix(net): Fix ICMP Error messages checksum payload computation**
- **fix(net): Fix deparsing for ICMP Error messages' inner headers**
- **fix(net): Fix checksums update order for ICMP Error msg's inner headers**
- **fix(net): Return empty extensions, not empty packet in ICMP's generator**
- **fix(net): Fix total packet size computation in ICMP's generator**
- **fix(net): Fix theoretical paylod size in ICMP Error msg generator**

These are fixes related to the processing (or Bolero generation) of ICMP Error messages' inner packets.
Refer to individual commits for details.

### DIFF
--- a/net/src/packet/mod.rs
+++ b/net/src/packet/mod.rs
@@ -579,12 +579,14 @@ pub mod contract {
                 .map_or(0, EmbeddedHeaders::transport_headers_len)
                 as usize;
             // Theoretical payload size for inner TCP/UDP (inner packet may be truncated)
-            let theoretical_inner_payload_size = if driver.produce::<bool>()? {
+            let theoretical_inner_payload_size = if payload_size == 0 {
+                0
+            } else if driver.produce::<bool>()? {
                 // Payload is full
                 payload_size
             } else {
-                // Payload is truncated
-                payload_size - (driver.produce::<usize>()? % payload_size)
+                // Payload is truncated: return a bigger theoretical payload size value
+                1000 - (driver.produce::<usize>()? % 1000) + payload_size
             };
             // Theoretical payload size for inner IP header (inner packet may be truncated)
             let theoretical_inner_net_payload_size =


net/src/icmp4/mod.rs
```diff
@@ -261,7 +261,7 @@ mod contract {
             let icmp_header = Icmpv4Header {
                 icmp_type: Icmpv4Type::DestinationUnreachable(
                     DestUnreachableHeader::from_values(
-                        driver.produce()?,
+                        driver.produce::<u8>()? % 16,
```
mvachhar: Does the 16 have to be a magic number?  Do we have a constant that explains what it is?

net/src/headers/mod.rs
```diff
@@ -166,7 +166,12 @@ impl Net {
...
 
 impl Transport {
-    pub(crate) fn update_checksum(&mut self, net: &Net, payload: impl AsRef<[u8]>) {
+    pub(crate) fn update_checksum(
+        &mut self,
+        net: &Net,
+        embedded_headers: Option<&EmbeddedHeaders>,
+        payload: impl AsRef<[u8]>,
+    ) {
         match (net, self) {
```
mvachhar: I'm not really a fan of this approach but it gets the job done.  We probably need a better way to just handle structured payloads and serialize them at the end.  Not worth doing now though.

mvachhar (changes requested):

net/src/headers/mod.rs
qmonnet: Not a fan, either. I don't see a cleaner, non-trivial solution. I can't find a good model that allows to efficiently consider inner headers both as some headers layer, and as payload for the outer headers :disappointed:.

net/src/icmp4/mod.rs
mvachhar: Not worth it, we'll keep the magic number for now.  I wish etherparse had a constant, but oh well.

PR #918, opened 2025-10-17
Fix concurrency for a Fib

Fredi-raspall:
This rides on https://github.com/githedgehog/dataplane/pull/917

Fixes https://github.com/githedgehog/dataplane/issues/894

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1263,6 +1263,7 @@ dependencies = [
  "bytes",
  "chrono",
  "dataplane-cli",
+ "dataplane-concurrency",
  "dataplane-config",
  "dataplane-left-right-tlcache",
  "dataplane-lpm",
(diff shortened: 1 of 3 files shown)


routing/src/fib/fibgroupstore.rs
```diff
@@ -452,10 +455,11 @@ pub mod tests {
         .unwrap();
 
         assert_eq!(fibroute.0.len(), 4);
-        assert_eq!(*fibroute.0[0].borrow(), *store.get(&key1).unwrap());
-        assert_eq!(*fibroute.0[1].borrow(), *store.get(&key2).unwrap());
-        assert_eq!(*fibroute.0[2].borrow(), *store.get(&key3).unwrap());
-        assert_eq!(*fibroute.0[3].borrow(), *store.get(&key4).unwrap());
+        assert_eq!(unsafe { &*fibroute.0[0].get() }, store.get(&key1).unwrap());
```
mvachhar: Ugh, I definitely hate this.  Can we have some shuttle tests here to make sure we don't get any memory corruption?  This probably needs some sort of rethink, but there isn't time for that right now, so really good multi-threaded testing is in order here.

mvachhar (changes requested):

PR #927, opened 2025-10-21
fix(metrics): Add periodic run_upkeep call

sergeymatov:
We have to make sure consistency of metrics for installed recorder is verified. According to docs, run_upkeep should be called periodically
https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/#upkeep-and-maintenance

### DIFF
--- a/dataplane/src/statistics/mod.rs
+++ b/dataplane/src/statistics/mod.rs
@@ -5,6 +5,7 @@ use axum::{Router, response::Response, routing::get};
 use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
 use stats::StatsCollector;
 use std::thread::JoinHandle;
+use std::time::Duration;
 use tracing::{error, info};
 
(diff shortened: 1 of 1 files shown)


dataplane/src/statistics/mod.rs
```diff
@@ -77,6 +78,16 @@ impl MetricsServer {
     #[tracing::instrument(level = "info", skip(stats))]
     async fn run(addr: std::net::SocketAddr, stats: StatsCollector) {
         let PrometheusHandler { handle } = PrometheusHandler::new();
+
+        let upkeep_handle = handle.clone();
+        tokio::spawn(async move {
+            let mut ticker = tokio::time::interval(Duration::from_secs(30));
```
qmonnet: It might be worth mentioning in the commit log how you picked this duration.

qmonnet: thanks!

PR #961, opened 2025-11-02
Extend cmd line args for interface & PCI address

Fredi-raspall:
Allowed syntax is:
  --interface ifname[=portdesc], where portdesc may be a PCI address

Examples:
  --interface eth0 -- interface eth1
  --interface eth0,eth1 is also allowed
  --interface eth0=0000:02:03.1
  --interface eth= is equivalent to --interface eth0

 * This PR should not break current behavior, but the portdesc is ignored.
 * The portdesc should deprecate --allowed of DPDK driver.

### DIFF
--- a/args/src/lib.rs
+++ b/args/src/lib.rs
@@ -2,13 +2,94 @@
 // Copyright Open Network Fabric Authors
 
 pub use clap::Parser;
+use hardware::pci::address::PciAddress;
 use mgmt::processor::launch::GrpcAddress;
+use net::interface::InterfaceName;
 use routing::rio::DEFAULT_DP_UX_PATH;
 use routing::rio::DEFAULT_DP_UX_PATH_CLI;
 use routing::rio::DEFAULT_FRR_AGENT_PATH;
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use std::str::FromStr;
 use tracing::debug;
+
+#[derive(Debug, Clone)]
+#[allow(unused)]
+pub struct InterfaceArg {
+    interface: InterfaceName,
+    pciaddr: Option<PciAddress>,
+}
+impl FromStr for InterfaceArg {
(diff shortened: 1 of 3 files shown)


sergeymatov (commented):
this may be needed for gRPC (mgmt) as well. Maybe args is not the best store for this giving we need to share this functionality?

args/src/lib.rs
```diff
@@ -144,8 +233,17 @@ impl CmdArgs {
     pub fn kernel_num_workers(&self) -> usize {
         self.num_workers.into()
     }
+    // backwards-compatible, to deprecate
```
qmonnet: Do we need to be backward-compatible at this point? Knowing we still haven't cut a release, it's probably a good time to remove it?

qmonnet (approved):
Looks good, just double-checking with regards to your commit description:

> Examples:
>   --interface eth0 -- interface eth1

Should this be `--interface eth0 --interface eth1` (no space before `interface`)?

>   --interface eth0,eth1 is also allowed
>   --interface eth0=0000:02:03.1
>   --interface eth= is equivalent to --interface eth0

For the last one, you mean `--interface eth0=` (with the `0`) equivalent to `--interface eth0`, right?

args/src/lib.rs
Fredi-raspall: It would be trivial to modify the code to use the new method. However, we have changes in the kernel driver in other branches (which are going to severely conflict) and making that change would just add more noise.

Fredi-raspall:
> this may be needed for gRPC (mgmt) as well. Maybe args is not the best store for this giving we need to share this functionality?

Oh, now I think I understand. You mean that args is not the best place to keep the PCI addresses as mgmt may need them? Yes and no. Drivers need this info early, before mgmt. Mgmt will learn them not via args.

PR #978, opened 2025-11-04
[WIP][DON'T MERGE]Test pyroscope

sergeymatov:
Add pyroscope agent starting and pushing into giving Alloy address+port. Default is localhost:4040

PR #982, opened 2025-11-05
feat(dataplane): pyroscope support

daniel-noland:
Add support for [pyroscope](https://pyroscope.io) continuous profiling.

Opened in favor of #978 and #980 to keep git history clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "adler32"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
+
+[[package]]
 name = "afpacket"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -470,9 +476,9 @@ dependencies = [
 
 [[package]]
 name = "bollard"
-version = "0.19.3"
+version = "0.19.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec7646ee90964aa59e9f832a67182791396a19a5b1d76eb17599a8310a7e2e09"
+checksum = "87a52479c9237eb04047ddb94788c41ca0d26eaff8b697ecfbb4c32f7fdc3b1b"
 dependencies = [
  "base64 0.22.1",
  "bollard-stubs",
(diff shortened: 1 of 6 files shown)


args/src/lib.rs
```diff
@@ -1034,6 +1034,14 @@ Note: multiple interfaces can be specified separated by commas and no spaces"
     )]
     metrics_address: SocketAddr,
 
+    /// Pyroscope server address for profiling uploads
+    #[arg(
+        long,
+        value_name = "ALLOY_ADDR:PORT",
```
daniel-noland: The `value_name` parameter says "ALLOY_ADDR:PORT" but this should actually be a URL (with scheme like `http://`). Consider changing to something like "PYROSCOPE_URL" to better reflect that it expects a full URL, not just an address:port.

```suggestion
        value_name = "PYROSCOPE_URL",
```

Cargo.toml
```diff
@@ -129,6 +129,8 @@ prefix-trie = { version = "0.8.1", default-features = false, features = [] }
 pretty_assertions = { version = "1.4.1", default-features = false, features = ["std"] }
 priority-queue = { version = "2.7.0", default-features = false, features = [] }
 procfs = { version = "0.18.0", default-features = false, features = [] }
+pyroscope = { version = "0.5.4", default-features = false, features = [] }
+pyroscope_pprofrs = { version = "0.2.10", default-features = false, features = [] }
```
daniel-noland: Version mismatch: The workspace dependency specifies `version = "0.5.4"` but Cargo.lock shows `version = "0.5.8"` was actually resolved. This suggests the version constraint is working as intended (allowing patch updates), but it's worth noting.

Also, there's a minor spacing inconsistency - this line is missing a space after the opening brace. For consistency with the rest of the file:

```suggestion
pyroscope_pprofrs = { version = "0.2.10", default-features = false, features = [] }
```

dataplane/src/main.rs
```diff
@@ -65,8 +68,29 @@ fn process_tracing_cmds(args: &CmdArgs) {
...
+                    error!("Pyroscope start failed: {e}");
+                    None
+                }
+            },
+            Err(e) => {
+                error!("Pyroscope build failed: {e}");
+                None
+            }
+        }
+    });
```
daniel-noland: Good implementation overall. The agent initialization happens before `process_tracing_cmds()` which can exit via `std::process::exit(0)` in three places (lines 50, 57, 64). 

However, I agree with the author's response to Copilot's comment - this is acceptable because:
1. The agent is only created if it successfully starts (returns `Some(running)`)
2. The process exits cleanly via `std::process::exit(0)` for informational commands
3. Pyroscope agents must be resilient to ungraceful termination anyway (SIGKILL, OOM, etc.)
4. The shutdown logic at the end is best-effort cleanup, not required for correctness

The 100Hz sampling rate seems reasonable for production profiling. Consider documenting this choice or making it configurable if users need different trade-offs between overhead and resolution.

Cargo.lock
```diff
@@ -3478,6 +3631,35 @@ dependencies = [
  "syn 2.0.108",
```
daniel-noland: The dependency additions look reasonable. Key new dependencies:
- `pyroscope` (0.5.8) and `pyroscope_pprofrs` (0.2.10) - core profiling libraries
- `reqwest` (0.12.24) - HTTP client for uploading profiles
- `pprof2` (0.13.1) - profiling backend
- Various supporting libraries for compression, demangling, etc.

Note: This adds `reqwest` which pulls in a substantial dependency tree including async runtime support. This is expected for a profiling client but worth noting for binary size impact.

Also brings in `tower-http` which adds ~24KB of dependencies for HTTP middleware support.

PR #992, opened 2025-11-07
DO NOT MERGE: Run NAT tests to validate latest fixes

pau-hedgehog:


.github/workflows/dev.yml
```diff
@@ -38,7 +38,7 @@ on:
         description: "Run hybrid HLAB tests"
         required: false
         default: false
-      enable_release_tests:
+      run_release_tests:
```
qmonnet: I'm not sure this change will take effect until this is in `main`. If it doesn't, try renaming the other variable as `enable_release_tests` in the condition evaluation?

PR #1001, opened 2025-11-07
Test - Disable session creation

qmonnet:
I just want CI to build an image from this branch.


qmonnet:
doesn't work with iperf3 for some reason

PR #1020, opened 2025-11-14
fix(dataplane): use dataplane cli as the shell

daniel-noland:
It is annoying to launch the dataplane container shell without this.


mvachhar (approved):

qmonnet:
@daniel-noland How does this work? I don't understand how this helps users to use the CLI; I asked Claude how this is supposed to work and Claude doesn't seem convinced, either:

image

We may have missed the point, though?

PR #1021, opened 2025-11-14
fix(nat): Fix timeout reset for table flows in stateful NAT

qmonnet:
When updating the timeout for a flow after a successful lookup in the flow table, we add the desired timeout duration to the existing timeout instant, instead of the current time, meaning that we linearly increment the duration (beyond the desired period) each time we receive a packet for that flow. Let's fix by resetting to the current time, plus the desired duration, instead.

We do this by introducing a new method `reset_expiry()`, doing

```rust
    self.expires_at.store(
        Instant::now() + duration,
        std::sync::atomic::Ordering::Relaxed,
    );
```

To compare with `extend_expiry()`, which does:

```rust
    self.expires_at
        .fetch_add(duration, std::sync::atomic::Ordering::Relaxed);
```

Fixes: #1019

### DIFF
--- a/flow-info/src/flow_info.rs
+++ b/flow-info/src/flow_info.rs
@@ -162,6 +162,33 @@ impl FlowInfo {
             .fetch_add(duration, std::sync::atomic::Ordering::Relaxed);
     }
 
+    /// Reset the expiry of the flow if it is not expired.
+    ///
+    /// # Errors
+    ///
+    /// Returns `FlowInfoError::FlowExpired` if the flow is expired with the expiry `Instant`
+    ///
+    pub fn reset_expiry(&self, duration: Duration) -> Result<(), FlowInfoError> {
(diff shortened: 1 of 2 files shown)


flow-info/src/flow_info.rs
```diff
@@ -162,6 +162,33 @@ impl FlowInfo {
...
+
+    /// Reset the expiry of the flow without checking if it is already expired.
+    ///
+    /// # Thread Safety
+    ///
+    /// This method is thread-safe.
+    ///
+    pub fn reset_expiry_unchecked(&self, duration: Duration) {
+        self.expires_at.store(
+            Instant::now() + duration,
```
mvachhar: This sucks, this makes a system call.  When it shows up at the top of the profile we can fix it.  I think we're quite a way away from that though.

mvachhar (approved):

PR #1026, opened 2025-11-15
Add VPC Counters gRPC message to track total traffic and drops

sergeymatov:
Depending on DoneReason of packet is now classified as TX or Drop. Additional prometheus counters and in-memory VPC stats storage entities are created to handle dropped traffic accordingly. To expose dedicated drops per giving VPC new VpcCounters message used from gateway-proto v0.17.0.

Fixes https://github.com/githedgehog/dataplane/issues/1025

### DIFF
--- a/mgmt/src/processor/proc.rs
+++ b/mgmt/src/processor/proc.rs
@@ -351,13 +351,14 @@ impl ConfigProcessor {
                     dst_vpc: dst_name,
                     packets: fs.ctr.packets,
                     bytes: fs.ctr.bytes,
-                    drops: 0, // TODO: Add this in Release 2
+                    drops: 0, // TODO: Add this in a later release
                     pps: fs.rate.pps,
+                    bps: fs.rate.bps,
                 },
             );
         }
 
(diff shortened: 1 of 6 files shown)


mgmt/src/processor/proc.rs
```diff
@@ -268,7 +270,12 @@ impl ConfigProcessor {
         // Helper to check if a flow stats has any traffic
         #[inline]
         fn has_traffic(fs: &stats::FlowStats) -> bool {
-            fs.ctr.packets > 0 || fs.ctr.bytes > 0 || fs.rate.pps > 0.0 || fs.rate.bps > 0.0
+            fs.ctr.packets > 0
```
Frostman: should drops be included as well?

mvachhar:
Fixes #1025

sergeymatov:
@mvachhar we don't have any kind of test for precise traffic measurement and comparative analysis with counted stats yet

Fredi-raspall (commented):
LGTM aside from a couple of nits.

Issue #952, opened 2025-10-30
Do not log at level Error on every failure to send a packet


mvachhar:
Currently we log in the kernel driver if any packet fails do send, this is a DoS vector.

qmonnet:
We log errors under certain circumstances when processing packets in the NAT stages, too. We should probably rework that? What's an acceptable usage of error generation in the context of NAT, for example? Internal errors: yes, but malformed packets: no?

PR #1031, opened 2025-11-17
chore(nat): Quiet down NAT logs, no errors/warnings on malformed packets

qmonnet:
Make sure we don't log errors or warnings when packets are malformed, but only on internal errors, for both stateless and stateful NAT.

Relates to: #952

### DIFF
--- a/nat/src/stateful/mod.rs
+++ b/nat/src/stateful/mod.rs
@@ -31,7 +31,7 @@ use std::num::NonZero;
 use std::time::{Duration, Instant};
 
 #[allow(unused)]
-use tracing::{debug, error, warn};
+use tracing::{debug, error, info};
 
 use tracectl::trace_target;
 trace_target!("stateful-nat", LevelFilter::INFO, &["nat", "pipeline"]);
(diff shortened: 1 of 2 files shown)


nat/src/stateful/mod.rs
```diff
@@ -644,15 +644,15 @@ impl StatefulNat {
     fn process_packet<Buf: PacketBufferMut>(&mut self, packet: &mut Packet<Buf>) {
         // TODO: What if no VNI
         let Some(src_vpc_id) = Self::get_src_vpc_id(packet) else {
-            warn!(
+            debug!(
```
Fredi-raspall: Ideally, the same here. If packets have no vpc discriminants, we should probably not even try to nat them ?
Also, shouldn't `Self::get_src_vpc_id(packet) `be replaced by just `packet.get_meta().dst_vpcd`? and 
`Self::get_src_vpc_id(packet)`  by  `packet.get_meta().src_vpcd`?
What is the benefit of those wrappers?

mvachhar (approved):

Fredi-raspall (changes requested):
I think we should clarify the purpose of these type of changes, not in this PR but in general.

PR #1032, opened 2025-11-17
feat(dataplane): simplify kernel driver tx_packet()

Fredi-raspall:
On top of https://github.com/githedgehog/dataplane/pull/1027

Do not use packet.enforce() since that is probably only useful in pipeline code. Reorganize tx_packet() and logs accordigngly, acquire interface lock only and after packet serialize succeeeds and remove redundant trace.


mvachhar (approved):
I think this looks good, but we should decide if we want to merge as part of R1 or wait until after the release.  Just depends on test schedule.  I've marked don't merge pending discussion.

Issue #957, opened 2025-10-31
Shared IP addresses behind stateful NAT will not route return traffic


mvachhar:
Looking through the code to add sessions to the session table, the stateful NAT code appears to never add the destination vpcd information to the flow table for reverse flows.  The NAT stage is the only stage that knows this information if there is a shared exposed IP address since it creates the flow entry based on the source vpcd.

Here [on this line ](https://github.com/githedgehog/dataplane/blob/dbd13ce276de96e2f8221433b126d2282a0898f4/nat/src/stateful/mod.rs#L560)
when the session table entry for the reverse session is created, we should see if the existing flow table entry has a dst vpcd, and if it does not, we should add one with the correct vpcd, which I believe is `src_vpc_id` in this case.

We should also add a test to exercise this case.

Note that if the IP address is not shared, then the destination vpcd lookup stage will resolve the destination vpcd, so we need to explicitly test when the IP is actually shared, in which case we expect this vpcd to be `None`.  Moreover, the test needs 2 peerings to 2 different VPCs.  

Finally, we need to check the validation logic to make sure that we do not forbid this shared IP case.

PR #1038, opened 2025-11-17
Stateful NAT: Support shared IPs

qmonnet:
Add support for shared IPs (IPs from overlapping CIDR ranges exposed by multiple VPCs to a given VPC) with stateful NAT.

To that end:

- Make the destination VPC lookup stage return no result, in case of ambiguity.
- Store the destination VPC discriminant as part of flow table entries.
- Add a 3rd session table entry with an empty destination VPC discriminant in the key, to allow for lookups when the discriminant is unknown.
- Only drop packets after flow table lookup, and not the initial destination VPC lookup, in the case when the latter could not return a unique destination VPC for a given packet.
- Add associated unit tests.

See individual commits for more details.

Fixes: #957

### DIFF
--- a/nat/src/stateful/test.rs
+++ b/nat/src/stateful/test.rs
@@ -4,7 +4,8 @@
 #[cfg(test)]
 mod tests {
     use crate::StatefulNat;
-    use config::GwConfig;
+    use crate::stateful::NatAllocatorWriter;
+    use concurrency::sync::Arc;
     use config::external::ExternalConfigBuilder;
     use config::external::overlay::Overlay;
     use config::external::overlay::vpc::{Vpc, VpcTable};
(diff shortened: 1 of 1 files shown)


mvachhar:
Adding don't merge since this is for R2

pkt-meta/src/dst_vpcd_lookup/mod.rs
```diff
@@ -168,11 +168,18 @@ impl DstVpcdLookup {
         let dst_ip = net.dst_addr();
         if let Some(vpcd_table) = tablesr.tables_by_discriminant.get(&src_vpcd) {
             let dst_vpcd = vpcd_table.dst_vpcds.lookup(dst_ip);
-            if let Some((prefix, dst_vpcd)) = dst_vpcd {
-                debug!(
-                    "{nfi}: Set packet dst_vpcd to {dst_vpcd} from src_vpcd:{src_vpcd}, prefix:{prefix}"
-                );
-                packet.meta.dst_vpcd = Some(*dst_vpcd);
+            if let Some((prefix, dst_vpcd_opt)) = dst_vpcd {
+                if let Some(dst_vpcd) = dst_vpcd_opt {
```
mvachhar: Should this be better done with a `match` on `Some` and `None`?

nat/src/stateful/test.rs
```diff
@@ -3,6 +3,7 @@
 
 #[cfg(test)]
 mod tests {
+    use crate::StatefulNat;
```
mvachhar: Generally the "paragraphs" I use are, `std::` deps, outside crate deps, workspace deps, and the `crate::` deps or sub-modules.  Doesn't really matter, just FYI.  Not sure how this ended up the way it did.

mvachhar (approved):
This looks good, though if you could make the minor changes, that would be appreciated.

mvachhar (changes requested):
We need to discuss the case of overlapping IPs for the same dst vpc, sorry I missed this in my prior approval.

pkt-meta/src/dst_vpcd_lookup/mod.rs
mvachhar: Oh, ok, then leave as is.

PR #1047, opened 2025-11-18
Fix default recipe in justfile

Fredi-raspall:
This does not need to go for R1, but we should merge for our ergonomics.


Frostman (approved):
It looks ok, though we can use `[default]` to make it more explicit, but I'm good with just reordering 

https://just.systems/man/en/the-default-recipe.html

PR #1049, opened 2025-11-18
bump(dpdk-sys): manual bump of dpdk-sys

qmonnet:
From https://github.com/githedgehog/dpdk-sys/pull/204

Let's see if this addresses the issues we observe with CI in #1024.

Do not merge


qmonnet:
CI remains stuck at

```
16:38:35 INF Downloading name=dataplane version=x86_64-unknown-linux-gnu.release.cebe4825a7b4d4500e10efd8a0e5e9d2d93e69e1 type=oci
```

so I guess it's the busybox change that causes the issue :disappointed:

PR #1074, opened 2025-11-24
Pass params to mgmt

Fredi-raspall:
Define struct MgmtParams to pass all of the parameters that the mgmt thread requires. This makes it much easier to pass new stuff, without the need to change several function signatures every time.


qmonnet (approved):
(Some of the minor changes in this commit (`pub` -> `pub (crate)`, `nattablew` -> `nattablesw`) look not strictly related and could be in their own commits, but OK.)

Looks good, and I agree it will make it easier to update the parameters, thank you!

qmonnet (approved):

PR #1091, opened 2025-11-26
fix: add dependency features in crates

Fredi-raspall:
Restore dependency features that were removed from the workspace Cargo.toml

feat(pkty-meta): fix dependency features


qmonnet:

PR #1092, opened 2025-11-26
fix(ci): fix upgrade skip logic

edipascale:


qmonnet:
image

Could we also skip it from the merge queue somehow?

Issue #1083, opened 2025-11-25
Stateful NAT: Session collision for shared IPs


qmonnet:
### The issue

Consider the following setup:

```
VPC-1          <->     VPC-2          <->   VPC-3
- ips: 1.0.0.1         - ips: 5.0.0.1       - ips: 3.0.0.1
  as:  2.0.0.1         web server port 80     as:  2.0.0.1
```

- VPC-1 has a peering with VPC-2
- VPC-3 has a peering with VPC-2
- Both VPC-1 and VPC-3 expose the same NAT-ed ranges
- VPC-2 expose the same range to both VPC-1 and VPC-3

VPC-1 and VPC-3 have no problems reaching out VPC-2, but replies from VPC-2 cannot use the simple destination VPC lookup stage based on source VPC + destination IP, there's a conflict. We recently added support to address this via a flow table lookup, to use existing flows to determine the destination VPC.

Now:

1. VPC-1 initiates a connection to VPC-2
2. We create a flow table entry for packets going from VPC-1 to VPC-2, a reverse entry for replies
3. We also create a third entry for replies, with an empty destination VPC discriminant, to use with the destination VPC lookup stage. The key for this third flow table entry is:
   - source VPC discriminant: VPC-2 id (reverse direction)
   - destination VPC discriminant: unset (we're looking for it)
   - source IP: 5.0.0.1
   - destination IP: 2.0.0.1
   - source port: 80
   - destination port: dst_port_1 (allocated from source NAT for the request)
4. Now, VPC-3 initiates a connection to VPC-2
5. We create the flow table entries for packets from VPC-3 to VPC-2, and for the reverse direction
6. We also create a third entry, again for the destination VPC lookup stage. The key is:
   - source VPC discriminant: VPC-2 id (reverse direction)
   - destination VPC discriminant: unset (we're looking for it)
   - source IP: 5.0.0.1
   - destination IP: 2.0.0.1
   - source port: 80
   - destination port: dst_port_2 (allocated from source NAT for the request)

Given that we don't set the destination VPC discriminant for these entries, they can be identical. In our example above, we use the same IPs and source port from VPC-2: we could have `dst_port_1` and `dst_port_2` being equal, in which case we get a collision, and we're unable to determine the destination VPC for the packets.

### What concretely happens in case of collision?

[...]

PR #1098, opened 2025-11-28
Fix port allocation with shared ips.

Fredi-raspall:
When using stateful NAT with addresses shared by distinct VPCs, the source port (translated) must be such that the return traffic allows to unambiguously recover the original flow. The existing implementation considers the source vpc id as part of the key to locate a port allocator. This has the problem that two flows from distinct vpcs natting to the same src IP may choose the same port since distinct port allocators are used. Fix this by not considering the source vpc id as part of the key to select a port allocator. This fixes the problem in general, without breaking the behavior when nat addresses are not reused across vpcs.

When NAT addresses are re-used among VPCs, it may happen that a VPC uses many more ports than the rest, potentially starving the latter. This can and should be addressed later by tracking the number of sessions/ports-allocated on a per VPC level wrt the fair share of ports, which should be the number of ports available over the number of vpcs. This is left for future work, and classical resource-sharing policies may be applied like having a strict quota or a relaxed one with possible eviction/pre-emption.

However, before addressing the above, the number of available sessions per nat IP address could be significantly increased by changing the criteria to allocate ports: ideally, with stateful source NAT, a distinct port allocator should be used per (nat_src_ip, dst_ip, protocol, dst_port and dst_vpc_id). Currently the dst_ip and dst_port are not taken into account.

Fixes: https://github.com/githedgehog/dataplane/issues/1083

### DIFF
--- a/nat/src/stateful/apalloc/mod.rs
+++ b/nat/src/stateful/apalloc/mod.rs
@@ -91,23 +91,15 @@ mod test_alloc;
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 struct PoolTableKey<I: NatIp> {
     protocol: NextHeader,
-    src_id: VpcDiscriminant,
     dst_id: VpcDiscriminant,
     addr: I,
     addr_range_end: I,
 }
 
 impl<I: NatIp> PoolTableKey<I> {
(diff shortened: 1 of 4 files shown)


mvachhar (approved):
Approved per discussion

PR #1102, opened 2025-12-02
bump(dpdk-sys): Manual bump to investigate update failures

qmonnet:
This is a manual bump to try a branch from dpdk-sys that hopefully does not contain the commits causing failures to update from main branch.

For testing only.


qmonnet:
CI's all red - reverting the DPDK-related commits didn't fix the issue. See https://github.com/githedgehog/dataplane/pull/1104 instead.

PR #1103, opened 2025-12-02
Add cmdline args to CLI

Fredi-raspall:
* Allows starting the CLI with the path of dataplane and the local binding address.
* This may simplify starting the CLI container cc @Frostman

### DIFF
--- a/cli/bin/main.rs
+++ b/cli/bin/main.rs
@@ -188,6 +188,9 @@ fn main() {
     while terminal.runs() {
         let mut bad_syntax = false;
         let mut input = terminal.prompt();
+        if !terminal.is_connected() {
+            terminal.connect(&cmdline.bind_address, &cmdline.path);
+        }
         if let Some(node) = cmds.find_best(input.get_tokens()) {
             if let Some(action) = &node.action {
                 if let Ok(args) = process_args(&input) {


cli/bin/cmdline.rs
```diff
@@ -2,3 +2,28 @@
...
 
 //! Cmd line to start cli
+
+use clap::Parser;
+
+pub(crate) const DEFAULT_DATAPLANE_PATH: &str = "/var/run/dataplane/cli.sock";
+pub(crate) const DEFAULT_CLI_BIND: &str = "/var/run/dataplane/cliclient.sock";
+
+#[derive(Parser)]
+#[command(about = "Hedgehog Fabric Gateway CLI", long_about = None)]
```
qmonnet: Still not a fan of naming Hedgehog in the project :confused:.

And you're adding yet a new variant here. Grepping in the repo, I find:
- Hedgehog Dataplane
- Hedgehog Open Network Fabric
- Hedgehog's Open Network Fabric
- Hedgehog Open Fabric Network   :upside_down_face:
- Hedgehog EVPN router           :thinking:
- Hedgehog Gateway               <-- Should we at least remove “Fabric” and use this instead?

qmonnet (approved):

cli/bin/cmdline.rs
qmonnet: I'd be happy to revive this debate. The internal doc I did about OSS is nearly one-year old, and not much has changed since then (this doc contains a number of names proposals if you're interested :see_no_evil:). I suppose that with our current release target, it's still not our priority :disappointed:

PR #1105, opened 2025-12-02
fix(README.md): Fix project name

qmonnet:
It should be "Open Network Fabric", and not "Open Fabric Network".


qmonnet:
> It would be useful to pack fixes like this into a single PR.

Right, but what fixes are you referring to? I don't plan to change other occurrences at the moment, I just wanted to fix this one which was manifestly wrong (and exposed in the README.md, with that).

PR #1106, opened 2025-12-02
fix(net): Fix Bolero generator for UnicastIpv4Addr

qmonnet:
Since commit efe3008734d7 ipv4 addr"), we forbid broadcast IPv4 addresses (namely, `255.255.255.255`) when building a `UnicastIpv4Addr`:

```rust
    impl UnicastIpv4Addr {
        pub fn new(ip: Ipv4Addr) -> Result<UnicastIpv4Addr, Ipv4Addr> {
            if ip.is_multicast() || ip.is_broadcast() {
                Err(ip)
            } else {
                Ok(UnicastIpv4Addr(ip))
            }
        }

        [...]
    }
```

We need to adjust the Bolero generator for `UnicastIpv4Addr` accordingly, or test `unicast_ipv4_is_unicast_ip` may end up trying to `unwrap()` on an `Err` value:

```rust
    fn unicast_ipv4_is_unicast_ip() {
        bolero::check!()
            .with_type()
            .cloned()
            .for_each(|ipv4: UnicastIpv4Addr| {
                let ip: UnicastIpAddr = ipv4.inner().try_into().unwrap();
                [...]
            });
    }
```

### DIFF
--- a/net/src/ipv4/addr.rs
+++ b/net/src/ipv4/addr.rs
@@ -108,6 +108,11 @@ mod contract {
                     raw[2],
                     raw[3],
                 )))
+            } else if ip.is_broadcast() {
+                // Note: is_broadcast() means "is 255.255.255.255?", it doesn't account for the
+                // current network in use. So when we reach this case we can just replace by any
+                // valid unicast address.
+                Some(UnicastIpv4Addr(Ipv4Addr::new(1, 1, 1, 1)))
             } else {
                 Some(UnicastIpv4Addr(ip))
             }


qmonnet:
> Is 255.255.255.255 always a broadcast address? What if it is in a /32 or a /31, I think it is not in that case. We may have to reconsider the name of this type, or how we treat the 255.255.255.255 address.

In this PR, I'm just trying to fix the CI flake we've observed.

I understand your point, do we want to change the code to account for broadcast (it will be painful to handle netmasks everywhere though, there's no notion of masks in the current NAT code for now but I still need to handle unicast addresses when translating source addresses), or simply to mention in the docs for the type  that it doesn't account for broadcast addresses (but just the multicast range)? @Fredi-raspall Why did we need to exclude 255.255.255.255 in the first place?

I'd like to get this fix (or a variant of it) merged to fix the flake, and we can open a follow-up issue to fix this multicast/broadcast confusion.

PR #1114, opened 2025-12-04
Add sample issue/PR templates

Fredi-raspall:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,30 @@
+---
+name: "🐛 Bug Report"
+about: Report a bug in the project
+title: "[BUG] "
+labels: bug
+---
+
+## 🐛 Description
+Describe the bug or the symptom
```
qmonnet: You probably want descriptions all around the template to be HTML comments, so that users don't have to select and delete them every time:

```suggestion
## 🐛 Description
<!-- Describe the bug or the symptom -->
```

mvachhar:
> What's the objective, and who is this targetting?

Same question, why are we doing this, why now, and for who?

Fredi-raspall:
> > What's the objective, and who is this targetting?
> 
> Same question, why are we doing this, why now, and for who?

There was some discussion in dataplane channel and wanted to get feedback on whether we'd need / want s/t like this.

qmonnet:

PR #1121, opened 2025-12-08
make sysroot optional

daniel-noland:
Draft commit exploring the removal of the sysroot as a mandatory build component


mvachhar (changes requested):
I really don't like the scope of breakage here in this commit.  Is there any way to automatically use the compile-env if it is there and then fallback if it isn't?

.github/workflows/dev.yml
```diff
@@ -339,17 +365,26 @@ jobs:
...
-            ${{matrix.profile.sterile}} cargo doc --no-deps
+          if [ "${{matrix.profile.sterile}}" != "sterile" ]; then
+            cargo doc --no-deps
+          else
+            RUSTDOCFLAGS="-D warnings" just \
+                debug_justfile="${{matrix.debug_justfile}}" \
+                profile=${{matrix.profile.name}} \
+                target=x86_64-unknown-linux-gnu \
+                ${{matrix.profile.sterile}} cargo doc --no-deps
+          fi
```
qmonnet: If you rebase, you should see an additional step just below this one, for running the doctests - You probably want to update that one, too.

.github/workflows/dev.yml
```diff
@@ -149,14 +155,17 @@ jobs:
...
-            refresh-compile-env
-          just --yes debug_justfile="${{matrix.debug_justfile}}" fake-nix
+          if [ "${{matrix.profile.sterile}}" = "sterile" ]; then
+            just --yes \
+                debug_justfile="${{matrix.debug_justfile}}" \
+                profile=${{matrix.profile.name}} \
+                registry="${{env.CACHE_REGISTRY}}" \
+                refresh-compile-env
+            just --yes debug_justfile="${{matrix.debug_justfile}}" fake-nix
+          fi
```
qmonnet: Split and run new step only for `sterile`?

PR #1127, opened 2025-12-10
fix(k8s-intf): Do not rebuild gateway_agent_crd.rs if not stale

mvachhar:
Previously build.rs was rewriting the gateway_agent_crd.rs file every time.  Cargo only uses timestamps to check for staleness so it was forcing a rebuild of k8s-intf and then all its deps.  Now, gateway-agent_crd.rs is only written if the new version would be different avoiding unneeded rebuilds.

### DIFF
--- a/k8s-intf/build.rs
+++ b/k8s-intf/build.rs
@@ -85,7 +85,29 @@ fn generate_rust_for_crd(crd_content: &str) -> String {
     }
 
     let raw = String::from_utf8(output.stdout).expect("Failed to convert kopium output to string");
-    fixup_types(raw)
+
+    LICENSE_PREAMBLE.to_string() + &fixup_types(raw)
+}
+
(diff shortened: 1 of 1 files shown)


qmonnet:
```console
$ just cargo nextest run -p dataplane-nat                                                                                                                                                       
   [...]
$ just cargo nextest run -p dataplane-nat                                                                                                                                                       
   Compiling dataplane-k8s-intf v0.1.0 (.../dataplane/k8s-intf)                                                                                                                   
   Compiling dataplane-config v0.1.0 (.../dataplane/config)                                                                                                                       
   Compiling dataplane-pkt-meta v0.1.0 (.../dataplane/pkt-meta)                                                                                                                   
   Compiling dataplane-nat v0.1.0 (.../dataplane/nat)                                                                                                                             
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1.72s
```

It doesn't seem to have addressed the issue :confused:

PR #1131, opened 2025-12-11
Add k8s client to pull config from k8s

### LINKED ISSUE #1111: Fetch gateway-agent object from k8s directly

mvachhar:
This PR does change the behavior of the default cli arguments.  If no GRPC command line arguments are specified (e.g., `--grpc-address`), then instead of assuming a default GRPC address of `0.0.0.0:50051`, it will use the k8s client to fetch configuration directly from kubernetes.  

VLAB can be made to use dataplane in agentless mode as well with the `agentless: true` option on the gateway controller configuration (set via the fabricator object).  However, this PR does not change CI to do that as we still do not report dataplane status through the k8s direct interface.  That will be the subject of the next PR in this series.

Because VLAB does not fully work with agentless mode because of the aforementioned lack of status updates, this PR has been tested manually and CI will still use GRPC for the vlab tests.

Fixes #1111

### DIFF
--- a/dataplane/src/main.rs
+++ b/dataplane/src/main.rs
@@ -65,6 +65,7 @@ fn process_tracing_cmds(args: &CmdArgs) {
     }
 }
 
+#[allow(clippy::too_many_lines)]
 fn main() {
     init_logging();
     let args = CmdArgs::parse();
@@ -130,8 +131,15 @@ fn main() {
     let pipeline_factory = setup.pipeline;
 
     /* start management */
+    let hostname = nix::unistd::gethostname()
+        .expect("Failed to get hostname")
+        .to_str()
+        .expect("hostname is not valid unicode")
+        .to_string();
+
     start_mgmt(MgmtParams {
         grpc_addr,
+        hostname,
         processor_params: ConfigProcessorParams {
             router_ctl: setup.router.get_ctl_tx(),
             vpcmapw: setup.vpcmapw,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1416,6 +1416,7 @@ dependencies = [
  "dataplane-config",
  "dataplane-id",
  "dataplane-interface-manager",
+ "dataplane-k8s-intf",
  "dataplane-lpm",
  "dataplane-nat",
  "dataplane-net",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,7 +110,6 @@ etherparse = { version = "0.19.0", default-features = false, features = [] }
 fixin = { git = "https://github.com/githedgehog/fixin", branch = "main", features = [] }
 futures = { version = "0.3.31", default-features = false, features = [] }
 hashbrown = { version = "0.16.1", default-features = false, features = [] }
-hostname = { version = "0.4.2", default-features = false, features = [] }
 hwlocality = { version = "1.0.0-alpha.11", default-features = false, features = [] }
 hyper = { version = "1.8.1", default-features = false, features = [] }
 hyper-util = { version = "0.1.19", default-features = false, features = [] }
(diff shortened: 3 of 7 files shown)


k8s-intf/src/bolero/gateway.rs
```diff
@@ -40,7 +40,11 @@ impl TypeGenerator for LegalValue<GatewayAgentGateway> {
             interfaces: Some(interfaces).filter(|i| !i.is_empty()),
             neighbors: Some(neighbors).filter(|n| !n.is_empty()),
             profiling: None, // FIXME(manishv) Add a proper implementation
-            protocol_ip: Some(d.produce::<UnicastIpv4Addr>()?.to_string()),
+            protocol_ip: Some(format!(
```
daniel-noland: This is a very good change / fix.

That said, this algorithm can generate _any_ `UnicastIipv4Addr`/`mask`, including `0.0.0.0/0`, `128.0.0.0/1` and `192.0.0.0/2`, all of which intersect with the multicast space (`224.0.0.0/4`).

While `0.0.0.0/0` is a potentially interesting case, `128.0.0.0/1` and `192.0.0.0/2` are only likely to show up if somebody is up to no good (or very confused).

I'm not 100% sure how best to handle this intersection.

daniel-noland: Of some note, failure to address this concern means that elements in the `UnicastIpv4Addr/mask` space are not strictly in the set of `UnicastIpv4Addr`

k8s-intf/Cargo.toml
```diff
@@ -10,18 +10,25 @@ bolero = ["dep:bolero", "dep:hardware", "dep:net", "dep:lpm", "net/test_buffer",
...
 bolero = { workspace = true, optional = true }
-hardware = { workspace = true, optional = true, features = [] }
-kube = { workspace = true, features = ["derive"] }
+futures = { workspace = true }
+kube = { workspace = true, features = ["client", "derive", "runtime", "rustls-tls"] }
 k8s-openapi = { workspace = true, features = ["latest", "schemars", "std"] }
-lpm = { workspace = true, optional = true }
-net = { workspace = true, optional = true, features = [] }
+linkme = { workspace = true }
+rustls = { workspace = true, features = ["ring"] }
```
daniel-noland: I think ring is great, but I also think that the world is generally trending toward [awc-lc-rs](https://github.com/aws/aws-lc-rs) due to extensive formal verification.

You might consider a follow up PR to swap this to aws-lc-rs (it should likely be transparent)

k8s-intf/src/client.rs
```diff
@@ -0,0 +1,65 @@
...
+        ..Default::default()
+    };
+    let mut stream = watcher(gws, watch_config)
+        .default_backoff()
+        .applied_objects()
+        .boxed();
+
+    loop {
+        match stream.try_next().await {
+            Ok(Some(ga)) => callback(&ga).await,
```
daniel-noland: nit: i generally use some type of async channel here, but the callback is fine.

I just find the channel to be easier to test / fuzz

k8s-intf/Cargo.toml
mvachhar: Glad to make this change.  I guess that means another test cycle so I can fix up the tokio::spawn stuff.  But I'd like to do it on #1138 instead of this PR.  Is that ok?

dataplane/src/main.rs
```diff
@@ -130,8 +131,15 @@ fn main() {
...
     /* start management */
+    let hostname = nix::unistd::gethostname()
+        .expect("Failed to get hostname")
+        .to_str()
+        .expect("hostname is not valid unicode")
+        .to_string();
+
     start_mgmt(MgmtParams {
         grpc_addr,
+        hostname,
```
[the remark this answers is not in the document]
mvachhar: K8s will guarantee that this is a valid hostname, though we could have a hostname type I suppose.  I believe the way the host name is passed in is about to change anyway so I am going to leave this as is for now.

PR #1136, opened 2025-12-13
ci: Run doctests in CI

qmonnet:
Make sure we catch regressions through doctests.


qmonnet:
```
ERROR: /tmp/rustdoctestnCz3Zt/rust_out is not executable
```

Looks like we need to fix this.

PR #1137, opened 2025-12-13
Port Address Translation (stateless)

qmonnet:
Stateless Port and Address Translation (PAT) for the dataplane.

Best reviewed per-commit, but hold on tight, this PR is dense.

Follow-ups include:
- Obviously, passing the port ranges from the user config.
- Adding port ranges support in stateful NAT (we already support NAT-ing ports in stateful NAT, the update should include allocating only from allowed target ranges, dropping packets for packets not coming from allowed ranges).
- ICMP. How do we do ICMP with PAT?? Still to be determined.
- Probably some more optimisations or code clarifications here and there.

### DIFF
--- a/lpm/src/prefix/with_ports.rs
+++ b/lpm/src/prefix/with_ports.rs
@@ -114,79 +114,122 @@ impl IpRangeWithPorts for PrefixWithPorts {
 
 /// A structure containing a prefix and an optional port range.
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
-pub struct PrefixWithOptionalPorts {
-    prefix: Prefix,
-    ports: Option<PortRange>,
+pub enum PrefixWithOptionalPorts {
+    Prefix(Prefix),
+    PrefixPorts(PrefixWithPorts),
 }
 
 impl PrefixWithOptionalPorts {
     /// Creates a new `PrefixWithOptionalPorts` from a prefix and an optional port range.
     #[must_use]
     pub fn new(prefix: Prefix, ports: Option<PortRange>) -> Self {
-        Self { prefix, ports }
+        match ports {
+            Some(ports) => {
+                PrefixWithOptionalPorts::PrefixPorts(PrefixWithPorts::new(prefix, ports))
+            }
+            None => PrefixWithOptionalPorts::Prefix(prefix),
+        }
     }
 
     /// Returns the prefix.
     #[must_use]
     pub fn prefix(&self) -> Prefix {
-        self.prefix
+        match self {
+            PrefixWithOptionalPorts::Prefix(prefix) => *prefix,
+            PrefixWithOptionalPorts::PrefixPorts(prefix_with_ports) => prefix_with_ports.prefix(),
+        }
     }
 
     /// Returns the optional port range.
     #[must_use]
     pub fn ports(&self) -> Option<PortRange> {
-        self.ports
+        match self {
+            PrefixWithOptionalPorts::Prefix(_) => None,
+            PrefixWithOptionalPorts::PrefixPorts(prefix_with_ports) => {
+                Some(prefix_with_ports.ports())
+            }
+        }
     }
 }
 
 impl IpRangeWithPorts for PrefixWithOptionalPorts {
     fn addr_range_len(&self) -> PrefixSize {
-        self.prefix.size()
+        match self {
+            PrefixWithOptionalPorts::Prefix(prefix) => prefix.size(),
+            PrefixWithOptionalPorts::PrefixPorts(prefix_with_ports) => {
+                prefix_with_ports.prefix().size()
+            }
+        }
     }
 
     fn port_range_len(&self) -> usize {
-        self.ports
-            .as_ref()
-            .map_or(PortRange::MAX_LENGTH, PortRange::len)
+        match self {
+            PrefixWithOptionalPorts::Prefix(_) => PortRange::MAX_LENGTH,
+            PrefixWithOptionalPorts::PrefixPorts(prefix_with_ports) => {
+                prefix_with_ports.ports().len()
+            }
+        }
     }
 
     fn covers(&self, other: &Self) -> bool {
-        match (self.ports, other.ports) {
-            (Some(self_ports), Some(other_ports)) => {
-                self.prefix.covers(&other.prefix) && self_ports.covers(other_ports)
-            }
-            (Some(self_ports), None) if self_ports.len() == PortRange::MAX_LENGTH => {
+        match (self, other) {
+            (
+                PrefixWithOptionalPorts::PrefixPorts(self_prefix_with_ports),
+                PrefixWithOptionalPorts::PrefixPorts(other_prefix_with_ports),
+            ) => self_prefix_with_ports.covers(other_prefix_with_ports),
+            (
+                PrefixWithOptionalPorts::PrefixPorts(self_prefix_with_ports),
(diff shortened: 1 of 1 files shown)


qmonnet:
I think it's ready for a first pass

lpm/src/prefix/with_ports.rs
```diff
@@ -0,0 +1,299 @@
...
+
+/// A type for the size of IP/port combinations for IP prefixes with associated port ranges.
+/// We need something larger than u128 to cover all possible combinations: the maximum value is
+/// `(u128::MAX + 1) * (u16::MAX + 1)`.
+pub type PrefixWithPortsSize = BUint<3>;
+
+/// Trait for IP ranges (CIDR prefix or simple ranges) with associated port ranges.
+pub trait IpRangeWithPorts {
+    /// Returns the address range length.
+    fn addr_range_length(&self) -> PrefixSize;
```
mvachhar: Rename to len to be consistent with container interfaces in Rust (i.e., `len()`)?

lpm/src/prefix/with_ports.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// Returns true if the range overlaps with the given other range.
+    fn overlaps(&self, other: &Self) -> bool;
+    /// Returns the total number of (IP, port) combinations covered by the IP and port ranges.
+    ///
+    /// # Warning
+    ///
+    /// If the port range is empty, this method should return the number of addresses in the address
+    /// range: so this will produce a different value than an object with the same IP range, but
+    /// with a port range covering all existing ports.
+    fn size(&self) -> PrefixWithPortsSize {
```
mvachhar: Why make this semantic distinction?  is it just to allow compatibility between an ip range and an ip range with ports?

In other words, if this trait is called IpRangeWithPorts, then should it always count (Ip, port) pairs available, so no port ranges means the size is the same as all port ranges?

If we what this to mean something like IpRangeOrIpRangeWithPorts (terrible name, we should choose something shorter), then the better way to say this is that it returns the available size of items.  For an IpRange that would be the number of IPs for an IPPortRange, that would be a different thing entirely.

Perhaps a discussion on the correct design here is in order.

lpm/src/prefix/with_ports.rs
```diff
@@ -0,0 +1,299 @@
+// SPDX-License-Identifier: Apache-2.0
```
mvachhar: Where are the tests for this new functionality?  I prefer fuzz tests if they are easy, but there should at least be some tests right?

pkt-meta/src/dst_vpcd_lookup/mod.rs
```diff
@@ -113,9 +117,15 @@ impl VpcDiscTablesWriter {
     }
 }
 
+#[derive(Debug, Clone)]
+enum VpcDiscriminantTableValue {
+    PortRangeTable(DisjointRangesBTreeMap<PortRange, Option<VpcDiscriminant>>),
+    NoPorts(Option<VpcDiscriminant>),
```
mvachhar: I think I like this enumeration approach better for the IpRangeWithPorts trait as well instead of using an empty port range to mean IPs only.  It might clean up my earlier comment about the size method as well.

lpm/src/prefix/with_ports.rs
```diff
@@ -450,3 +450,434 @@ impl Display for PrefixWithOptionalPorts {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
mvachhar: Ok, so now we have some tests, but still missing the basic tests for overlap.

lpm/src/trie/prefix_map_impl.rs
```diff
@@ -132,6 +132,17 @@ where
             .get_lpm(&IpPrefixW(addr.into()))
             .map(|(p, v)| (&p.0, v))
     }
+
+    fn matching_entries<A>(&self, addr: A) -> impl Iterator<Item = (&P, &V)>
+    where
+        A: Into<Self::Prefix>,
+    {
+        self.0
```
mvachhar: Is this a linear search over the entire lpm?  If so this is not acceptable, it will have terrible performance.

If this is a more clever walk of only the populated subtrees it could be ok for now, though we probably want a better data structure long term anyway.

mvachhar:
Great initial cut, but I think it still needs some work.

config/src/utils/collapse.rs
```diff
@@ -45,238 +43,97 @@ pub fn collapse_prefixes_peering(peering: &Peering) -> Result<Peering, ConfigUti
...
-                result.append(&mut apply_exclude_result);
+    for exclude in excludes {
+        for prefix in result.clone() {
+            // If the allowed prefix overlaps with the exclusion prefix, then it means the exclusion
+            // prefix excludes a portion of this allowed prefix. We need to remove the allowed
+            // prefix, and add instead the smaller fragments resulting from the application of the
+            // exclusion prefix.
+            if prefix.overlaps(exclude) {
+                result.remove(&prefix);
+                for p in prefix.subtract(exclude) {
```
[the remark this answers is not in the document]
mvachhar: Oh, prefix is a PrefixWIthOptionalPorts so it supports this.  Very nice abstraction.

lpm/src/prefix/with_ports.rs
qmonnet: The issue is that it's not clear what `.len()` refers to for the struct: IP range length, port range length, total number of IPs x ports?

I can rename to `addr_range_len` if you prefer, though.

lpm/src/prefix/with_ports.rs
qmonnet: I'm not sure I see what functionality you refer to, in particular. There are some unit tests in follow-up commits after the introduction of these structs and methods.

Fuzzing is tricky, most of the interesting methods to fuzz require re-implementing the logic in the fuzzer to validate correctness. I don't have any fuzzing tests in this PR, we could look into it but it would probably be for a follow-up.

config/src/utils/collapse.rs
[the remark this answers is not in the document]
qmonnet: Discussed on a call: Exclusion prefixes (with or without port ranges) should be supported with this PR.

lpm/src/trie/prefix_map_impl.rs
qmonnet: As discussed, it's using this method from the underlying structure:
https://docs.rs/prefix-trie/latest/prefix_trie/map/struct.PrefixMap.html#method.cover

It walks the subtrees so should be OK, at least until we find a better data structure.

lpm/src/prefix/with_ports.rs
qmonnet: The comment is in fact outdated. I had a previous version of the code that made a distinction, but it's been changed a while ago. I'll remove the comment.

qmonnet: Comment removed

lpm/src/prefix/with_ports.rs
qmonnet: I added commit `test(config): Add tests for VpcExpose validation with port ranges`, I think it should address your concern (except for fuzzing).

qmonnet:
@mvachhar Your feedback should be mostly addressed, please take another look when you can.

lpm/src/prefix/with_ports.rs
mvachhar: I think the rest of the tests test this functionality enough to omit tests here for now.

PR #1138, opened 2025-12-13
Add status k8s converters and k8s status update code

mvachhar:
<del>*Do not merge until #1131 is merged*</del>

This PR adds status updates to the gateway agent CRD objects.  It also fixes a small bug related to repeated application of the same config generation.

Testing is once again manual as this mode is not active by default in vlab tests.  We should switch vlab to agentless mode after this PR merges. <del>or if desired for this PR itself.</del>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -228,6 +228,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.15.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b5ce75405893cd713f9ab8e297d8e438f624dde7d706108285f7e17a25a180f"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.34.0"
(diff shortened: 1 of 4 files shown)


k8s-intf/src/client.rs
```diff
@@ -63,3 +74,43 @@ pub async fn watch_gateway_agent_crd(
...
+        status_obj.status = Some(status.clone());
+        let status_json = serde_json::to_vec(&status_obj)?;
+
+        match api
+            .replace_status(gateway_object_name, &PostParams::default(), status_json)
+            .await
+        {
+            Ok(_) => break,
+            Err(kube::Error::Api(api_error)) => {
+                if api_error.code == 409 {
```
daniel-noland: this seems like a good place for a backoff or a sleep

mgmt/src/processor/k8s_client.rs
```diff
@@ -1,56 +1,229 @@
...
+#[derive(Debug, thiserror::Error)]
+enum MgmtStatusError {
+    #[error("Failed to fetch dataplane status: {0}")]
+    FetchStatusError(String),
+    #[error("No config is currently applied")]
+    NoConfigApplied,
+}
+
+pub struct K8sClient {
+    hostname: String,
```
daniel-noland: part of me want to make a qualified type around `String` which describes valid hostnames.  Will leave that for another PR tho

mvachhar: Yep, that is needed along with UnicastIpv[46]Net that only generates value unicast addresses.  Add it to the list :)

PR #1158, opened 2025-12-19
Add flow-filter stage to assert packets belong to a valid peering connection

qmonnet:
The new flow-filter crate introduced in this commit contains a new pipeline stage for validating that a packet matches an existing peering connection, as defined in the configuration provided by the user. All packets that do not have a source IP, port and destination IP, port corresponding to existing, valid connections between the prefixes in exposed lists of peerings, get dropped.

This allows us to enforce that traffic matches the peering rule:

- Generically, whether or not the appropriate rules are set
- For stateless NAT, which would just let the packet go un-NATed if there was no NAT rule to be found, so far
- In the future, for stateful NAT, allowing us to remove some checks in the stateful NAT code.

Because the new stage gets the whole context about the authorized connections (as per the peering configuration), it can also replace the destination VPC lookup, to avoid redundant structures and lookups.

In the future, we could also easily adjust and use the stage to mark whether packets should be processed for stateless or stateful NAT, which would allow us to drop the "NAT exempt list" hack used for stateful NAT.

~PR is not ready to be merged because some points need discussion, but it's ready for review and comments.~


mvachhar (approved):
This looks really nice.  When can we remove the dst_vpcd_lookup stage?

flow-filter/src/setup.rs
```diff
@@ -0,0 +1,308 @@
...
+    vcpd_right: &VpcDiscriminant,
+    get_ips: fn(&VpcExpose) -> &BTreeSet<PrefixWithOptionalPorts>,
+) -> BTreeMap<PrefixWithOptionalPorts, HashSet<VpcDiscriminant>> {
+    let mut overlap = BTreeMap::new();
+    for prefix_left in manifest_left
+        .exposes
+        .iter()
+        .flat_map(|expose| get_ips(expose).iter())
+    {
+        for prefix_right in manifest_right
```
mvachhar: This is fine for now, but will cause problems with very large numbers of exposes and vpcs (which may never happen given switch HW limitations and practical uses).  But if it does, the whole algorithm here can probably be made more efficient by precomputing all the overlaps by stuffing everything into one big trie in advance.  No need to do it now, something to keep in mind if this ever becomes and issue.

PR #1159, opened 2025-12-19
ci: fix workflow_dispatch trigger of vlab jobs

pau-hedgehog:
Addresses issue in dev.yml Workflow Dispatch where VLAB jobs were not created properly: https://github.com/githedgehog/dataplane/actions/runs/20379018360
image

With this PR:
https://github.com/githedgehog/dataplane/actions/runs/20380250628
image

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -111,7 +111,7 @@ jobs:
     needs:
       - check_changes
       - version
-    if: "${{ needs.check_changes.outputs.devfiles == 'true' || startsWith(github.event.ref, 'refs/tags/v') && github.event_name == 'push' }}"
+    if: "${{ needs.check_changes.outputs.devfiles == 'true' || (startsWith(github.event.ref, 'refs/tags/v') || startsWith(github.ref, 'refs/tags/v')) && (github.event_name == 'push' || github.event_name == 'workflow_dispatch') }}"
     permissions:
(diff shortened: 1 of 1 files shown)


qmonnet:
Hmm the filter works for PRs, but the issue is that it's probably different in the context of dispatched workflows. How did you trigger that workflow (manual dispatch I suppose, but on what branch?) From [this doc](https://github.com/dorny/paths-filter/?tab=readme-ov-file#supported-workflows) I'd expect the filter to apply to the latest commit on the branch, and if this was https://github.com/githedgehog/dataplane/commit/6298bc044248eb61707e201a132cfea75207a949 it should have triggered the VLAB :thinking:

pau-hedgehog:
> Hmm the filter works for PRs, but the issue is that it's probably different in the context of dispatched workflows. How did you trigger that workflow (manual dispatch I suppose, but on what branch?) From [this doc](https://github.com/dorny/paths-filter/?tab=readme-ov-file#supported-workflows) I'd expect the filter to apply to the latest commit on the branch, and if this was [6298bc0](https://github.com/githedgehog/dataplane/commit/6298bc044248eb61707e201a132cfea75207a949) it should have triggered the VLAB 🤔

This is triggered via workflow dispatch on `master`: https://github.com/githedgehog/dataplane/actions/runs/20713855006

PR #1165, opened 2025-12-30
Remove gRPC

Fredi-raspall:
On top of #1164

Remove dont-merge lable once #1164 merges and this PR points to main.


mvachhar (approved):
This PR makes me very happy.  We should archive the gateway proto after it merges.  @Frostman are we good to go on the fabricator/fabric side to drop gRPC?

Frostman:
@mvachhar yeah, we've switched to forced agentless on Dec 18

PR #1182, opened 2026-01-12
remove gwname

Fredi-raspall:
* remove the static gateway name
* this is needed in order to perform the offline validation of gateway configs, since we want to use exactly the same code that dataplane will use to validate, and relying on a static would be problematic.


qmonnet (approved):
Looks OK. From what I understand the gateway name is no longer shared as a static variable, but it is still set via the command line after this PR, right?

PR #1187, opened 2026-01-13
Minimal validator

Fredi-raspall:
Addresses https://github.com/githedgehog/dataplane/issues/1174

* Adds a simple validator, as a rust binary. 
* The intent is to compile this validator as wasm/wasi and use it from fabricator.
* The validator expects a GatewayAgent CRD from stdin and produces a JSON-encoded response in stdout.
* The output JSON describes if the CRD could be validated or otherwise the problem encountered.
* The "errors" include issues that the user may not be responsible for, but which prevent the validation from happening.
* The errors are organized mimicking the sequence of steps done by dataplane when validating a config. Roughly:
     * io (specific to this tool)
     * deserialize from json (we should not see these in prod, but if CRD is manually created)
     * metadata issues (these should be integration bugs)
     * conversion to config (e.g. bad values vni = 0 or > 2^24 )
     * config errors (these are the ones the user is responsible for).
         *  these type of errors contain as a list/sequence
         * however, currently we only return one element because dataplane validation early exits when the first issue is encountered.
         
I will open a separate PR to:
    * refine the feedback and the information provided by some of the errors.
    * move some checks from conversion to validation, as they pertain there.
    * see if an exhaustive (non-early-exited) validation can be added.

**Sample deserialization errors**
In these errors, a list of lines (context) near the issue is returned. The line number is not too exact (depending on the error type it is offset by +/-1 or 2). We should seldom (if ever) see deserialization errors when the input JSON is machine generated. So this is mostly for testing manually crafted configs.
```
{
  "DeserializeError": {
    "hint": "trailing comma at line 8 column 3",
    "line": 8,
    "column": 3,
    "category": "Syntax",
    "context": {
      "3": "  \"metadata\": {",
      "4": "    \"generation\": 1,",
      "5": "    \"name\": \"gateway-1\",",
      "6": "    \"namespace\": \"fab\",",
      "7": "  },",
      "8": "  \"spec\": {",
      "9": "  \"agentVersion\": \"v0.32.0\",",
      "10": "  \"communities\": {",
      "11": "    \"0\": \"65000:800\",",
      "12": "    \"1\": \"65000:801\","
    }
  }
}
```

```
{
  "DeserializeError": {
    "hint": "invalid number at line 150 column 15",
    "line": 150,
    "column": 15,
    "category": "Syntax",
    "context": {
      "145": "    },",
      "146": "    \"VPC-2\": {",
      "147": "      \"internalID\": \"BBBBB\",",
      "148": "      \"subnets\": null,",
      "149": "      \"vni\": 000",
      "150": "    },",
      "151": "    \"VPC-3\": {",
      "152": "      \"internalID\": \"CCCCC\",",
      "153": "      \"subnets\": null,",
      "154": "      \"vni\": 2000"
    }
  }
}
```

```
{
  "DeserializeError": {
    "hint": "invalid type: string \"65000-BAD\", expected u32 at line 20 column 22",
    "line": 20,
    "column": 22,
    "category": "Data",
    "context": {
      "15": "    \"4\": \"65000:804\",",
      "16": "    \"5\": \"65000:805\"",
      "17": "  },",
      "18": "  \"gateway\": {",
      "19": "    \"asn\": \"65000-BAD\",",
      "20": "    \"groups\": [",
      "21": "      {",
      "22": "        \"name\": \"gw-group-1\",",
      "23": "        \"priority\": 10",
      "24": "      },"
    }
  }
}

```
```
{
  "DeserializeError": {
    "hint": "invalid value: integer `-1500`, expected u32 at line 79 column 20",
    "line": 79,
    "column": 20,
    "category": "Data",
    "context": {
      "74": "    },",
      "75": "    \"protocolIP\": \"7.0.0.100/32\",",
      "76": "    \"vtepIP\": \"7.0.0.100/32\",",
      "77": "    \"vtepMAC\": \"02:aa:bb:cc:dd:ee\",",
      "78": "    \"vtepMTU\": -1500,",
      "79": "    \"workers\": 1",
      "80": "  },",
      "81": "  \"groups\": {",
      "82": "    \"gw-group-1\": {",
      "83": "      \"members\": ["
    }
  }
}
```

**Sample Metadata errors**
```
{
  "MetadataError": "Missing gateway name"
}
```

```
{
  "MetadataError": "Missing generation Id"
}
```

**Sample Conversion errors**
```
{
  "ConversionError": "Invalid Gateway Agent object: Could not create VPC: Bad VPC Id"
}
```
```
{
  "ConversionError": "Could not parse value: Invalid CIDR format: 192.168.90.0/34: Invalid Prefix: 192.168.90.0/34"
}
``` 
```
{
  "ConversionError": "Invalid Gateway Agent object: Could not create VPC: '0' is not a valid VNI"
}
``` 

**Sample Configuration errors**
```
{
  "Configuration": {
    "errors": [
      "A VPC peering object refers to non-existent VPC 'VPC-1'"
    ]
  }
}
```
**Success case**
```
"Success"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1689,6 +1689,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "dataplane-validator"
+version = "0.8.0"
+dependencies = [
+ "dataplane-config",
+ "dataplane-k8s-intf",
+ "serde",
+ "serde_yaml_ng",
+]
+
+[[package]]
 name = "dataplane-vpcmap"
 version = "0.8.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ members = [
     "sysfs",
     "test-utils",
     "tracectl",
+    "validator",
     "vpcmap",
 ]
 
@@ -75,6 +76,7 @@ stats = { path = "./stats", package = "dataplane-stats", features = [] }
 sysfs = { path = "./sysfs", package = "dataplane-sysfs", features = [] }
 test-utils = { path = "./test-utils", package = "dataplane-test-utils", features = [] }
 tracectl = { path = "./tracectl", package = "dataplane-tracectl", features = [] }
+validator = { path = "./validator", package = "dataplane-validator", features = [] }
 vpcmap = { path = "./vpcmap", package = "dataplane-vpcmap", features = [] }
 
 # External
--- a/validator/Cargo.toml
+++ b/validator/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "dataplane-validator"
+edition.workspace = true
+license.workspace = true
+publish.workspace = true
+version.workspace = true
+
+[[bin]]
+name = "validator"
+path = "src/main.rs"
+
+[dependencies]
+config = { workspace = true }
+k8s-intf = { workspace = true }
+
+serde = { workspace = true }
+serde_yaml_ng = { workspace = true }
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -1,0 +1,204 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! A configuration validator. This validator may perform the same validation that
+//! the dataplane process. The intent is to compile this validator as WASM / WASI.
+//! The validator expects a `GatewayAgent` CRD in JSON or YAML from stdin and produces
+//! a result as a YAML string in stdout.
+
+#![deny(clippy::all)]
+#![allow(clippy::result_large_err)]
+#![allow(clippy::field_reassign_with_default)]
+
+use config::{ExternalConfig, GwConfig};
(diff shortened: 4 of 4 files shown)


Fredi-raspall:
> @Fredi-raspall, we need to agree on the output format; the current one would be very difficult to consume to produce a user-friendly message and difficult to parse on top of it. I'd suggest to at least flatten it a bit into list of errors and move the error type into field `type` instead of it being a key and always have an object as a root, e.g.:

Ok, I'll see what I can do. Note that some of the errors are exclusive. I.e., if we fail to deserialize, we'll not be able to validate the config. So you'll never see "DeserializeError" along with "ConversionError".

> I additionally suggest renaming `hint` to `message` and making it mandatory for all errors to have a `message`.

Ok!

Frostman (approved):

Frostman:
Tried the latest binary @Fredi-raspall shared in Slack, and it looks ok. I think we can merge in that state and main work would be to make all validation messages as user friendly as possible.

PR #1198, opened 2026-01-15
Support for default exposes

Fredi-raspall:
* This is not yet finished. Some changes are needed wrt to validation
* Addresses https://github.com/githedgehog/internal/issues/287
* Need to see how to adapt this to https://github.com/githedgehog/dataplane/pull/1158
* Pipeline may fail until https://github.com/githedgehog/fabricator/issues/1313 gets merged (?)

### DIFF
--- a/config/src/external/overlay/vpc.rs
+++ b/config/src/external/overlay/vpc.rs
@@ -156,20 +156,49 @@ impl Vpc {
         Ok(())
     }
 
-    /// Check that prefixes exposed to a given VPC do not overlap (except with default expose)
-    fn check_overlap(&self) -> ConfigResult {
+    /// Check that prefixes exposed to a given VPC do not overlap (except with default expose).
+    /// Also check that at most one default expose is exposed to the VPC.
+    fn check_overlap_and_default(&self) -> ConfigResult {
+        let mut found_default = false;
+
         // FIXME: Find a less expensive approach to find overlapping prefixes
         for (i, current_peering) in self.peerings.iter().enumerate() {
+            // Check we don't have multiple default expose blocks in the peering
+            for expose in &current_peering.remote.exposes {
+                if expose.default {
+                    if found_default {
+                        error!(
+                            "Multiple 'default' expose blocks for a same peering in VPC {}",
+                            self.name
+                        );
+                        return Err(ConfigError::Forbidden(
+                            "Multiple 'default' expose blocks for a same peering",
+                        ));
+                    }
+                    found_default = true;
+                }
+            }
+
+            // Check we don't have non-default, overlapping prefixes exposed to the VPC
             for other_peering in &self.peerings[i + 1..] {
                 for current_expose in &current_peering.remote.exposes {
-                    if current_expose.default {
-                        // Overlap is allowed with default expose
-                        continue;
-                    }
                     for other_expose in &other_peering.remote.exposes {
-                        if other_expose.default {
-                            // Overlap is allowed with default expose
-                            continue;
+                        match (current_expose.default, other_expose.default) {
+                            (true, true) => {
+                                // We support at most one default destination exposed to any VPC
+                                error!(
+                                    "Multiple 'default' destinations exposed to VPC {}",
+                                    self.name
+                                );
+                                return Err(ConfigError::Forbidden(
+                                    "Multiple 'default' destinations exposed to VPC",
+                                ));
+                            }
+                            (true, false) | (false, true) => {
+                                // Overlap is allowed between a prefix and a default expose
+                                continue;
+                            }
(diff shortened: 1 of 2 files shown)


config/src/external/overlay/vpc.rs
```diff
@@ -114,6 +114,28 @@ impl Vpc {
         }
     }
 
+    /// Check that a [`Vpc`] does not peer more than once with another.
```
qmonnet: I can't remember, do we have a consensus that this is would be an invalid configuration? Asking because I forgot this test was here (I note it was present before your PR though).

config/src/external/overlay/vpc.rs
```diff
@@ -33,6 +34,21 @@ pub struct Peering {
...
+    fn validate(&self) -> ConfigResult {
+        debug!(
+            "Validating manifest of VPC {} in peering {}",
+            self.local.name, self.name
+        );
+        self.local.validate()?;
+        if false {
+            // not needed will be validated when validating the remote vpc
+            self.remote.validate()?;
+        }
```
qmonnet: Looks dangerous if we ever make `self.remote` slightly different from the peer's `self.remote` for whatever reason in the future? Maybe add a comment stating we relying on the manifests being equal for validation, at the location where we assign them to the peerings?

config/src/external/overlay/mod.rs
```diff
@@ -28,48 +28,51 @@ impl Overlay {
...
-            "Overlay configuration is VALID and looks as:\n{}\n{}",
-            self.vpc_table, self.peering_table
-        );
-
-        /* empty peering table: we no longer need it since we have collected
-        all of the peerings and added them to the corresponding VPCs */
-        self.peering_table.clear();
-
-        /* empty collections used for validation */
-        self.vpc_table.clear_vnis();
```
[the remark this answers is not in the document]
Fredi-raspall: We cleared that in the past because we were keeping a bunch of configs in memory.
Nowadays, we only keep one. So that's not a concern. The other concern was that we should not using the peering table once we have collected the peerings. However, I'd like to add a proper Display to log exactly what we got and how we processed it. So, that's the reason why I just removed those noisy methods.
I'm planning to rework part of that stuff for the validation feature, but I wanted to keep this PR focused and small.
We could do what you suggest (I can add it on the validation PR), but not in this one.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -279,6 +279,40 @@ impl VpcExpose {
...
+                }
+                if nat.not_as.iter().any(|p| p.prefix().is_root()) {
+                    return Err(ConfigError::Forbidden(
+                        "Expose: root prefix as NAT 'as-not' is forbidden",
+                    ));
+                }
+            }
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
Fredi-raspall: I think the original, while longer, is way simpler to follow.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -307,6 +307,40 @@ impl VpcExpose {
         self.nat.as_ref().is_some_and(VpcExposeNat::is_stateless)
     }
 
+    fn validate_default_expose(&self) -> ConfigResult {
+        if self.default {
+            if !self.ips.is_empty() || !self.nots.is_empty() || self.nat.is_some() {
```
qmonnet: > `if !self.ips.is_empty() || !self.nots.is_empty()`

These two checks are redundant with what you add to `config/src/converters/k8s/config/expose.rs` in commit 2.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -307,6 +307,40 @@ impl VpcExpose {
...
+            if self.ips.iter().any(|p| p.prefix().is_root()) {
+                return Err(ConfigError::Forbidden(
+                    "Expose: root prefix as 'ip' forbidden",
+                ));
+            }
+            if self.nots.iter().any(|p| p.prefix().is_root()) {
+                return Err(ConfigError::Forbidden(
+                    "Expose: root prefix as 'not' is forbidden",
+                ));
+            }
```
qmonnet: This is not really necessary (same for `not_as` below), given that we check later that the exclusion prefixes do not cover the whole set of allowed prefixes, in step 4 in `VpcExpose.validate()`.

config/src/external/overlay/vpcpeering.rs
Fredi-raspall: Yes, they are. But let's keep them. As part of the validation work, I'll move some of the checks around.
I'm planning to fix those before the release.

PR #1205, opened 2026-01-16
Add support for "default" destination to flow-filter stage

qmonnet:
Work in progress.

~~This needs to be rebased on top of https://github.com/githedgehog/dataplane/pull/1198, once it is merged.~~
**NOTE:** this is now rebased on https://github.com/githedgehog/dataplane/pull/1198 and could be merged to that PR so that we keep a consistent history of PRs vs features in github.

Currently, this PR also misses:

- New tests to validate the behaviour of the changes in flow-filter
- Equivalent changes in NAT context tables to handle the "default" destinations properly


qmonnet:
There's a bug in the logic, I need to address it.

Fredi-raspall:
> @Fredi-raspall I kept but changed your commit to keep just one line of log per lookup, let me know if you really prefer your version.

I preferred mine (even if more verbose) because I dislike debug prints

PR #1209, opened 2026-01-19
fix underlay forwarding

Fredi-raspall:
~~This goes on top of https://github.com/githedgehog/dataplane/pull/1198 although it is not too related.~~
~~However, I needed to add it there to be able to test the latest changes around defaults.~~
Fixes https://github.com/githedgehog/dataplane/issues/1210

### DIFF
--- a/config/src/internal/routing/bgp.rs
+++ b/config/src/internal/routing/bgp.rs
@@ -171,7 +171,7 @@ impl Default for BgpOptions {
         Self {
             network_import_check: false,
             ebgp_requires_policy: false,
-            bestpath_aspath_relax: false,
+            bestpath_aspath_relax: true,
             always_compare_med: false,
             bgp_default_unicast: false,
             supress_duplicates: true,


qmonnet (approved):
Nice, I like the last commit in particular.

Cargo.toml
```diff
@@ -71,7 +71,7 @@ mgmt = { path = "./mgmt", package = "dataplane-mgmt", features = [] }
 nat = { path = "./nat", package = "dataplane-nat", features = [] }
 net = { path = "./net", package = "dataplane-net", features = [] }
 pipeline = { path = "./pipeline", package = "dataplane-pipeline", features = [] }
-pkt-meta = { path = "./pkt-meta", package = "dataplane-pkt-meta", features = [] }
+flow-entry = { path = "./flow-entry", package = "dataplane-flow-entry", features = [] }
```
qmonnet: Nit: Move up in the list, to keep them sorted

qmonnet (commented):
Looks good. I've got mixed feelings about the crate renaming, `flow-entry` makes sense to me but I fear there will be confusion between `flow-entry` and `flow-info`. But I suppose we can always rename again in the future if it's too confusing. It will be nice indeed to have the two closer to each other in the list.

qmonnet:
CI complains about a few remaining includes referring to `pkt-meta`, it seems

Cargo.toml
Fredi-raspall: fixed

PR #1225, opened 2026-01-21
In flow-filter, mark NAT requirements for packets

qmonnet:
Based on the context we have for the flow-filter stage, it's easy to add NAT requirements and to attach them to packets, to simplify processing in the NAT stages (and avoid traversing the NAT stages in the first place for the packets that don't require it).

### DIFF
--- a/nat/src/stateful/apalloc/mod.rs
+++ b/nat/src/stateful/apalloc/mod.rs
@@ -77,6 +77,7 @@ use net::packet::VpcDiscriminant;
 use std::collections::BTreeMap;
 use std::fmt::Display;
 use std::net::{Ipv4Addr, Ipv6Addr};
+use tracing::error;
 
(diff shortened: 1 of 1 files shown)


qmonnet:
Needs a rebase, but shouldn't change much

flow-filter/src/tables.rs
```diff
@@ -454,14 +454,31 @@ impl ValueWithAssociatedRanges for SrcConnectionData {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) enum NatRequirement {
+    StatelessNatRequired,
```
Fredi-raspall: I find the `NatRequirement::StatelessNatRequired` a bit redundant / verbose?
Maybe `NatRequirement::Stateless` ?

PR #1226, opened 2026-01-21
fix time conversion

Fredi-raspall:
Fixes https://github.com/githedgehog/internal/issues/295


k8s-intf/Cargo.toml
qmonnet: After you've switched to `kube-core`, I think (but didn't test) that you can remove dependency crate `serde-duration-ext`, here and in the toplevel Cargo.toml. Would you mind checking and removing it if it's no longer used, please?

qmonnet (changes requested):
I didn't check what format the new type accepts, but I trust you did. The change looks good, can I only ask that we remove the previous dependency we used for this? See below

qmonnet (approved):

PR #1227, opened 2026-01-21
Fix sending from "default" expose in flow-filter

qmonnet:
We recently added support for "default" expose blocks used as a fallback destination for packets from the VPC when the destination IP matches no other exposed prefixes. From the "default" block we deduce information about the destination VPC, and about whether the packet has a valid destination. But we forgot to use it for the reverse check: when a packet is emitted from a VPC, from behind a "default" block, its source address should always be considered as valid. This is not the case! So if we have VPC A and B peered together, and VPC B exposing a "default" block, VPC A can send packets to VPC B through this block, but all replies from VPC B are dropped.

This PR fixes the issue by adding a "default source" connection information object to the context for the flow-filter stage, in the per-source-VPC-discriminant table. This information is used to validate the destination and retrieve the destination VPC when a packet is emitted from behind a "default" expose block.

### DIFF
--- a/lpm/src/prefix/with_ports.rs
+++ b/lpm/src/prefix/with_ports.rs
@@ -594,21 +594,16 @@ impl Display for PortRange {
 
 impl Display for PrefixWithPorts {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{} <{}>", self.prefix, self.ports)
+        write!(f, "{}:[{}]", self.prefix, self.ports)
     }
 }
 
 impl Display for PrefixWithOptionalPorts {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            PrefixWithOptionalPorts::Prefix(prefix) => write!(f, "({prefix} </>)"),
+            PrefixWithOptionalPorts::Prefix(prefix) => write!(f, "{prefix}:[all]"),
             PrefixWithOptionalPorts::PrefixPorts(prefix_with_ports) => {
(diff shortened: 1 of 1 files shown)


flow-filter/src/tables.rs
```diff
@@ -171,18 +175,56 @@ impl FlowFilterTable {
...
+        }
+        Ok(())
+    }
 }
 
 #[derive(Debug, Clone)]
-struct VpcConnectionsTable(IpPortPrefixTrie<SrcConnectionData>);
+struct VpcConnectionsTable {
+    trie: IpPortPrefixTrie<SrcConnectionData>,
+    default_source_opt: Option<SrcConnectionData>,
```
Fredi-raspall: I am not sure to understand the purpose and need of this default_source_opt.

Fredi-raspall: Oh, ok. Now I see what you're trying to do.

Fredi-raspall: Ok. I've had the chance to look into the fix more carefully. I am not 100% sure that this solution is correct. Maybe I am missing something. The concern that I have has to do with the `SrcConnectionData`. In the use case motivating a default expose, you'd probably not want to filter based on source ports, would you? If you would:
* I believe that a better approach would be to use firewall rules / ACLs
* All of the traffic that did not have ports (notably ICMP) or many others would be filtered out.
So, I believe that in 99.9% of the cases you would not want to filter based on source ports?
If that is the case, I think that the logic can be significantly simplified?
Apologies in advance if I am missing something.

Fredi-raspall (changes requested):

PR #1230, opened 2026-01-22
build: Prune some unused dependencies from Cargo.toml files

qmonnet:
Found with cargo-udeps, plus some manual grepping for external dependencies names from toplevel Cargo.toml.


Fredi-raspall:
@qmonnet I've reviewed the changes and they look good to me.
I'd probably not remove the tracectl & linkme in some places because those may be likely needed in the future to control logs.

qmonnet:
I rebased the commit to address a conflict resulting from the merge of the BMP work.

Assuming we can indeed re-add required dependencies when we do need them, and unless @daniel-noland has objections to raise, I think this is good to go?

PR #1245, opened 2026-01-27
Pr/fredi/cleanup and validation

Fredi-raspall:
This work is not finished but can be safely merged, since I'll be working on other higher prio matters and don't want it to rod.

### DIFF
--- a/config/src/external/gwgroup.rs
+++ b/config/src/external/gwgroup.rs
@@ -32,18 +32,13 @@ impl GwGroupMember {
 impl Ord for GwGroupMember {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         let ord_prio = self.priority.cmp(&other.priority);
-        let ord_ip = self.ipaddress.cmp(&other.ipaddress);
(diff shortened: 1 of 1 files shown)


config/src/external/gwgroup.rs
```diff
@@ -33,14 +33,10 @@ impl Ord for GwGroupMember {
...
         let ord_ip = self.ipaddress.cmp(&other.ipaddress);
-        let ord_name = self.name.cmp(&other.name);
 
         if ord_prio == Ordering::Equal {
-            if ord_ip == Ordering::Equal {
-                ord_name
-            } else {
-                ord_ip
-            }
+            assert_ne!(ord_ip, Ordering::Equal);
```
qmonnet: This looks dangerous. Should it be a debug assert instead?

Fredi-raspall: There would be a bug if that happened.

Fredi-raspall: But fine changing it.

PR #1255, opened 2026-02-02
chore: we should use default ns for gw agents in k8s

Frostman:


Frostman:
Upgrade isn't supposed to work anymore

PR #1257, opened 2026-02-03
portfw + masquerading preparation  + flow-table-based dst_vpcd lookup with overlapping prefixes

Fredi-raspall:

PR #1264, opened 2026-02-04
flow-filter update for dst_vpcd lookup fallback with flow table

qmonnet:
Completing https://github.com/githedgehog/dataplane/pull/1257

- The first commit is a fix for the last commit in #1257, and could probably be merged into it.

- Then we wrap the lookup result in an enum that supports a `MultipleMatch` case, and we update the packet processing accordingly.

- After that we update the flow-filter stage to build this `MultipleMatch` case. This includes re-introducing splitting for prefixes with partial overlap, which I think will help us for the follow-up work on port forwarding.

- We adjust the peerings validation step to allow overlapping prefixes, provided they use stateful NAT.

- We add the unit tests back. We like tests.

### DIFF
--- a/flow-filter/src/lib.rs
+++ b/flow-filter/src/lib.rs
@@ -448,6 +448,220 @@ mod tests {
     }
 
     #[test]
+    fn test_flow_filter_multiple_matches_no_dst_vpcd() {
+        // Setup table with overlapping destination prefixes from different VPCs
+        let mut table = FlowFilterTable::new();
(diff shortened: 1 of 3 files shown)


qmonnet:
I think the changes are ready to review. Note that this is a big PR, although most commits should remain pretty simple. Parts of it are re-adding some code that we removed, parts of it are new.

The first commit may be squashed with the commit on the top of the merge branch (it's on purpose that lint-commitlint.yml doesn't pass at the moment).

PR #1283, opened 2026-02-15
fix flow table removals

Fredi-raspall:
Fixes https://github.com/githedgehog/dataplane/issues/1282

### DIFF
--- a/flow-entry/src/flow_table/nf_lookup.rs
+++ b/flow-entry/src/flow_table/nf_lookup.rs
@@ -177,4 +177,95 @@ mod test {
         let num_entries = flow_table.len().unwrap();
         assert_eq!(num_entries, 0);
     }
+
+    //#[traced_test]
+    #[test]
+    fn test_lookups_with_related_flows() {
+        let flow_table = Arc::new(FlowTable::default());
+        let lookup_nf = LookupNF::new("lookup_nf", flow_table.clone());
+        let expirations_nf = ExpirationsNF::new(flow_table.clone());
+        let mut pipeline: DynPipeline<TestBuffer> = DynPipeline::new()
+            .add_stage(lookup_nf)
+            .add_stage(expirations_nf);
+
+        {
+            let mut packet_1 = build_test_udp_ipv4_packet("10.0.0.1", "20.0.0.1", 80, 500);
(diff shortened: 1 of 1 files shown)


flow-entry/src/flow_table/nf_lookup.rs
```diff
@@ -104,4 +112,69 @@ mod test {
...
+    struct FlowInfoCreator(Arc<FlowTable>);
+    impl<Buf: PacketBufferMut> NetworkFunction<Buf> for FlowInfoCreator {
+        fn process<'a, Input: Iterator<Item = Packet<Buf>> + 'a>(
+            &'a mut self,
+            input: Input,
+        ) -> impl Iterator<Item = Packet<Buf>> + 'a {
+            input.filter_map(move |packet| {
+                let flow_key =
+                    FlowKey::try_from(crate::flow_table::flow_key::Uni(&packet)).unwrap();
+                let flow_info = FlowInfo::new(Instant::now() + Duration::from_secs(2));
```
qmonnet: Nit: I'd probably go for only 1 sec here, and 2-second wait below, assuming the tests still pass reliably. I'm not concerned about CI duration but just that it's never pleasing to wait when running the unit tests from CLI 😅. Not blocking, though

flow-entry/src/flow_table/table.rs
```diff
@@ -254,22 +274,31 @@ impl FlowTable {
...
+        let reaped_keys = self.priority_queue.reap_all_expired_with_time(
+            time,
+            Self::decide_expiry,
+            Self::do_reap,
+        );
+        self.remove_flow_entries(&reaped_keys)
     }
 
     #[cfg(test)]
+    #[allow(clippy::len_without_is_empty)]
```
qmonnet: Nit: This should be part of the first commit rather than the second.

qmonnet (approved):
Thanks for that! I think it looks good. I've got two small nits below.

flow-entry/src/flow_table/nf_lookup.rs
```diff
@@ -104,4 +112,58 @@ mod test {
         let output = output_iter.next().unwrap();
         assert!(output.meta().flow_info.is_some());
     }
+
+    // A dummy NF that creates a flow entry for each packet, with a lifetime of 2 seconds
```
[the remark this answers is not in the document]
qmonnet: Doesn't matter much anyway

PR #1289, opened 2026-02-17
Bump to gateway 0.42.0 (new NAT API)

qmonnet:
This is to pull the new API for NAT (introducing port forwarding, mandatory NAT block specification, etc.).

Here are the different steps that led to this commit:

1. We start by updating scripts/k8s-crd.env.
2. Then we regenerate file k8s-intf/src/generated/gateway_agent_crd.rs to pull the new definitions in.
3. After that, we adjust the K8s converter to work with the new NAT configuration.
4. We also update the related Bolero generators.

Note that we do not support port forwarding yet: the dataplane will return an error if the agent passes a configuration setting up port forwarding for an expose block. Support for port forwarding is currently in progress and will come in a separate change set. We prepare the ground anyway: we keep the port-parsing code (marking it as dead code for this commit), and update the Bolero generator in prevision of a future NAT mode.


qmonnet:
Alright let's go for it. Thanks for the reviews!

Issue #1085, opened 2025-11-25
"Overwriting" an entry in flow table drops the value


qmonnet:
TL;DR: `push()`-ing in the flow table's priority queue does not replace with the new value when a key already exists in the queue, resulting in an invalid weak reference in the flow table.

### Expected behaviour

Looking at our flow table implementation, we've got an `insert()` method, [defined here](
https://github.com/githedgehog/dataplane/blob/d25be6ca666eb05fcd9ff4a6a4834361d93d7b9e/pkt-meta/src/flow_table/table.rs) that says:

```rust
    /// Add a flow to the table.
    ///
    /// # Returns
    ///
    /// Returns the old `Arc<FlowInfo>` associated with the flow key, if any.
    ///
    /// [...]
    pub fn insert(&self, flow_key: FlowKey, flow_info: FlowInfo) -> Option<Arc<FlowInfo>> {
       ...
    }
```

Looking at this documentation, the expectation is that, when trying to insert an entry with a key that is already present in the table, the new value will replace the existing one, and the old value will be returned by the method.

### Actual behaviour

- The old value is returned indeed
- The new value is not inserted correctly. Instead, we only keep a "weak" reference in the table.

### Explanations

Here's what happens for the first insertion:

1. We want to create a new entry, we call `self.insert(flow_key, flow_info)`
2. `insert()` takes an `Arc` pointer - a strong reference - to `flow_info`, then calls `self.insert_common(flow_key, &val)` ([defined here](https://github.com/githedgehog/dataplane/blob/d25be6ca666eb05fcd9ff4a6a4834361d93d7b9e/pkt-meta/src/flow_table/table.rs#L126))
3. `insert_common()` downgrades the pointer and stores a **weak pointer** in the `self.table` (the "table" itself)
4. `insert_common()` also pushes a clone of the `Arc` **into the priority queue**, to be able to reap expired entries later
5. `insert_common()` exits, returning `None` (no prior value)
6. `insert()` exits, the `Arc` gets out of scope and is dropped, but it's OK because we still hold a clone in the priority queue so the weak reference in `self.table` remains valid
7. ... Later, at lookup time, we find the weak reference in `self.table` and successfully upgrade it to obtain an `Arc` to our initial `flow_info` object

Now, here's what happens when trying to insert another value (same or different), with an identical key:

1. We call `self.insert(flow_key, flow_info)`
[...]

PR #1290, opened 2026-02-17
fix(flow-entry): Fix entries overwriting for flow table

qmonnet:
We've had a long-standing issue where push()-ing an entry into the flow table's priority queue does not actually insert the new value when an entry with an identical key already exists in the queue, resulting in an invalid weak reference in the flow table; here's a fix.

See commit description or linked issue for details.

Fixes: https://github.com/githedgehog/dataplane/issues/1085

### DIFF
--- a/flow-entry/src/flow_table/table.rs
+++ b/flow-entry/src/flow_table/table.rs
@@ -461,22 +461,16 @@ mod tests {
                 let entry = table
                     .get(&flow_key)
                     .expect("entry should exist after second insert");
-                // FIXME: We currently have a bug here! Overwriting the entry led to dropping the
-                // strong reference from the priority queue.
-                //
-                //let resolved = entry
-                //    .upgrade()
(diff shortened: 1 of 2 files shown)


Fredi-raspall:
> I think we should accept this. The correct fix is to drop the priority queue entirely in favor of tokio's timers.

ok. I'd just add a test, since we don't know when the timers support will be ready?

qmonnet:
> * If you remove the entry from the PQ (which holds the strong references), even if you add it again, isn't that going to drop the object and any weak reference to it become invalidated?

So I checked: I remove any existing entry with a similar key, and that one is dropped, as expected (given that it's also no longer present from the flow table itself, it's been overwritten). But I don't remove or drop the _new_ value, which is different, before inserting it in the priority queue, so we keep the strong reference to that one.

I've added a test to validate the behaviour.

Fredi-raspall (approved):
LGTM, thanks @qmonnet !

PR #1314, opened 2026-03-03
Preserve flows on reconfigs with port-forwarding

Fredi-raspall:
Fixes https://github.com/orgs/githedgehog/projects/2/views/23?sliceBy%5Bvalue%5D=Fredi-raspall&pane=issue&itemId=161526276&issue=githedgehog%7Cdataplane%7C1313

In the original implementation of port-forwarding, rules referred to single ips and ports. When a flow for a port-forwarded packet was created, the corresponding state referred to the matched rule so that the data-path could refer to it (via a Weak reference) to learn things like the timeouts, needed to extend flows' lifetimes.

With rules that can contain prefixes and ports, this is more diffi- cult since, on reconfigurations, it is hard to preserve the origi- nal rule since the object (mostly immutable) may need to be replaced by a newer one in case it would need to be updated. If that happens, the Weak reference to the previous rule is dropped and it is no longer possible to recover, from a flow, the rule that lead to its creation. The option of letting the flow-filter deal with this is insufficient, because, even if after a reconfig the flow-filter would allow the flow, the port-forwarding logic requires knowing the timeouts (present only in the rule) to determine how much to extend the flows.

This patch fixes the logic so that, if a packet in either direction hits a flow with port forwarding state whose corresponding rule has been dropped or modified, the PortForwarder NF re-evaluates the packet to newly determine if it should continue to be port-forwarded according to the new config. In that case, a new Weak reference to the newer rule is created. The approach of using Weak references is hence still a valid one to expedite (omit reevaluations) most of the time in the "fast path". The slower path only applies:
  - for the first packet, when there's no state
  - for the first packet hitting a flow that refers to a stale port-forwarding rule, which can only happen after a reconfigu- ration.

### DIFF
--- a/nat/src/portfw/flow_state.rs
+++ b/nat/src/portfw/flow_state.rs
@@ -7,7 +7,7 @@
 
 use net::buffer::PacketBufferMut;
 use net::ip::UnicastIpAddr;
-use net::packet::{DoneReason, Packet};
+use net::packet::Packet;
 use std::fmt::Display;
 use std::num::NonZero;
 use std::sync::{Arc, Weak};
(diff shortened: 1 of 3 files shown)


sergeymatov (approved):
@Fredi-raspall thank you

mvachhar (approved):

Issue #1312, opened 2026-03-02
Bypass flow-filter if packets have flows


Fredi-raspall:
If packets refer to flows, let them bypass the flow-filter if we know the flow is legitimate with the current config.
We easily know if a flow is legitimate by labeling it with the configuration id.
This:
    * avoids complex processing of packets, helping in performance
    * simplifies the flow-filter code.

PR #1342, opened 2026-03-14
Safely bypass the flow-filter

Fredi-raspall:
* Augments pipeline with some PipelineData so that all NFs that are part of the pipeline can have access to it on their creation.
* Currently the pipeline data only includes the generation Id. We use this so that NFs can easily know the generation id of the currently applied config, with minimal changes.
* Augment flows to have a generation Id. The generation id of a flow is the config generation when the flow was created (or subsequently validated).
* Augment NAT NFs so that they create flows with generation ids
* Implement bypass of flow-filter so that:
     * packets that refer to flows can bypass the filter if the configuration has not changed.
     * on config changes, if a packet refers to a flow set  up with a previous config, the packet does not bypass the filter.
       If the packet survives, the flow is upgraded and subsequent packets bypass, because the flow conforms to the new config.
       If the packet is not allowed, that means that the flow should not be allowed and it gets invalidated.

These changes not only expedite packet processing but also simplify the flow-filter implementation.
E.g. with these, the extra checks in PR  https://github.com/githedgehog/dataplane/pull/1341 should no longer be needed.

This fixes https://github.com/githedgehog/dataplane/issues/1312

### DIFF
--- a/nat/src/stateful/mod.rs
+++ b/nat/src/stateful/mod.rs
@@ -22,7 +22,7 @@ use net::flow_key::{IcmpProtoKey, Uni};
 use net::flows::{ExtractRef, FlowInfo};
 use net::headers::{Net, Transport, TryIp, TryIpMut, TryTransportMut};
 use net::packet::{DoneReason, Packet, VpcDiscriminant};
-use net::{FlowKey, FlowKeyData, IpProtoKey};
+use net::{FlowKey, IpProtoKey};
 use pipeline::{NetworkFunction, PipelineData};
 use std::fmt::{Debug, Display};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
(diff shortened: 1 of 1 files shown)


pipeline/src/static_nf.rs
```diff
@@ -23,6 +26,9 @@ pub trait NetworkFunction<Buf: PacketBufferMut> {
         &'a mut self,
         input: Input,
     ) -> impl Iterator<Item = Packet<Buf>> + 'a;
+
+    /// Let NFs access some `PipelineData` if they wish on their creation
```
qmonnet: Although the intent is to use it at creation time, there's no mechanism preventing NFs from calling this method after their creation, right?

(I'm not asking for any particular change here, just making sure I understand.)

PR #1357, opened 2026-03-18
Pr/daniel noland/build system

daniel-noland:


Frostman (changes requested):
I wasn't really digging into the code changes, mainly trying it and looking at the DX. In general, build & test works fine.

1. `cargo build` doesn't work in nix-shell, you should set some default VERSION env var in the shell, maybe just `(devel)` or smth, main problem is that this is exactly the version that's going to be set when you do push so it should be working fine, alternatively you can wrap cargo with just to inject properly calculated version (which is IMO better idea)
2. `just push` is missing
3. matrix approach is VERY inefficient, need to decrease usage to max 8 runners in parallel
4. we shouldn't be building and pushing all the debug/sanitize by default in PRs

PR #1368, opened 2026-03-22
Fix flow-filter bypass and flow invalidation

Fredi-raspall:
This fixes https://github.com/githedgehog/fabricator/issues/1574

### DIFF
--- a/flow-filter/src/lib.rs
+++ b/flow-filter/src/lib.rs
@@ -95,6 +95,7 @@ impl FlowFilter {
             debug!("Flow-info is out-dated. New dst VPC is {dst_vpcd}");
             return true;
         }
+        drop(locked_info);
         if !packet.meta().requires_port_forwarding() && !packet.meta().requires_stateful_nat() {
             debug!("Flow {flowkey} no longer requires state. Will invalidate...");
             return true;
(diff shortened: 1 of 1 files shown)


flow-filter/src/lib.rs
```diff
@@ -57,6 +57,96 @@ impl FlowFilter {
...
+        }
+        // we could not invalidate despite the config change. This does not mean that the flow is
+        // valid (nor invalid). The NFs annotated in the requirements must determine. E.g. if we were
+        // masquerading with address A and, a new config, requires masquerading with address B, the above
+        // won't invalidate the flow, but the NF should (or update it accordingly).
+        false
+    }
+
+    /// Check if flow-info is up-to-date and allows bypassing the main filtering logic.
+    fn bypass_with_flow_info<Buf: PacketBufferMut>(
```
qmonnet: Nit: I'd like to keep `process_packet()` either at the top, or at the bottom, of the `FlowFilter` implementation. Can you please move this functions below it, or if you prefer to keep them at the top, could you please move `process_packet()` to the bottom?

qmonnet (changes requested):
Thanks!

Some small comments below.

To pass CI, you'll need to pick the first two commits from Daniel's PR: https://github.com/githedgehog/dataplane/pull/1304/commits

It would also be great to have unit tests in flow-filter for the bypass mechanism, in particular to validate the fix from the current PR.

mvachhar (approved):
I am approving this so that you can address @qmonnet comments in your morning and get this merged, but please create an issue to deal with the stale flow issues mentioned in this commit message.

PR #1380, opened 2026-03-23
wasm32 validator support

daniel-noland:
Adds support for wasm builds of the validator package.

I used claude a fair bit, especially to organize the many many rebases I needed to do here.

This was created with the help of Claude (Opus 4.6).

~~I have not yet signed off on the commits because I wan't to give them another once over before I do.  Everything seems fine, but I think the commit messages likely need some work.  That said, I'm taking this out of DRAFT for the moment so that copilot will take a look~~

I think this looks good and am now officially ready for a review :D

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -132,6 +132,7 @@ jobs:
           - tests.all
           - frr.dataplane
           - dataplane
+          - validator
         build:
           - name: "debug"
             profile: "debug"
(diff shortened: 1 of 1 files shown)


config/src/external/overlay/vpcpeering.rs
```diff
@@ -733,7 +733,11 @@ impl VpcPeering {
     }
 
     #[cfg(test)]
-    /// Validate A VpcPeering. Only used in tests. Dataplane validates `Peerings`
+    /// Validate A `VpcPeering`. Only used in tests. Dataplane validates `Peerings`
```
mvachhar: This backtick fix should probably be in the earlier commit that removes the allow for the lint.

PR #1398, opened 2026-03-26
cli fixups (AI assisted, undergoing review)

daniel-noland:


daniel-noland (commented):
The first commit in the series needs a "why" commit message rather than just a "what"

The motivation for the change revolved around the unmaintained status of bincode2

cli/bin/main.rs
```diff
@@ -63,6 +63,11 @@ fn ask_user(question: &str) -> bool {
 /// message (as 8 octets|u64) and then the message itself, in two writes.
 /// Therefore, here, we'll do 2 reads; one to figure out the length and a second
 /// one to received the actual message (response).
+/// Upper bound on a single CLI response payload.  Anything larger than this is
```
daniel-noland: this comment is injected in the wrong place.

It is obstructing / hiding the above rustdoc.  Please  move up

daniel-noland (commented):
NOTE: `allow(unused)` is an especially problematic lint disable because it suppresses unused `Result` types.  The `Result` enum is marked with `#[must_use]` but that is disabled by  `#[allow(unused)]` and for that reason the removal is more significant that it might seem.

cli/bin/terminal.rs
```diff
@@ -112,21 +110,24 @@ impl Terminal {
...
     pub fn get_helper(&self) -> Option<&CmdCompleter> {
         self.editor.helper()
     }
+    // Takes `&self` for API consistency even though the body only writes to stdout.
     #[allow(clippy::unused_self)]
     pub fn clear(&self) {
         print!("\x1b[H\x1b[2J");
         let _ = stdout().flush();
     }
+    // Takes `&self` for API consistency; parsing is pure but callers hold a Terminal.
```
daniel-noland: more "also" changes.  Move to another commit

daniel-noland (commented):
This series makes good progress in four different directions.

Those different directions should be covered by different PRs.

1. minor bug fixes
2. `allow($lint)` removals (with associated justification for the import of removing those `allow`s (especially `unused`)
3. future looking changes (admonitions regarding what should be done written without the intent that the changes be merged as such, just notes about important changes we would like)
4. the simplified `if` mechanics from the final commit in the sequence are substantial enough to justify their own PR (even tho they are technically bug fixes)

PR #1414, opened 2026-03-27
Stateful nat flow handling across reconfigs

Fredi-raspall:
Fixes https://github.com/githedgehog/internal/issues/342

There are further changes needed. I'll open follow-up PRs.

### DIFF
--- a/config/src/external/overlay/vpc.rs
+++ b/config/src/external/overlay/vpc.rs
@@ -314,17 +314,13 @@ impl Vpc {
             > 0
     }
 
-    /// Provide an iterator over all peerings that have either masquerade or port-forwarding exposes
+    /// Provide an iterator over all peerings that have either masquerade or port-forwarding exposes locally
     pub fn stateful_nat_peerings(&self) -> impl Iterator<Item = &Peering> {
         self.peerings.iter().filter(|p| {
             p.local
                 .exposes
                 .iter()
                 .any(|e| e.has_port_forwarding() || e.has_stateful_nat())
-                || p.remote
-                    .exposes
-                    .iter()
-                    .any(|e| e.has_port_forwarding() || e.has_stateful_nat())
         })
     }
 }
--- a/config/src/external/overlay/vpc.rs
+++ b/config/src/external/overlay/vpc.rs
@@ -313,6 +313,20 @@ impl Vpc {
             .count()
             > 0
     }
+
+    /// Provide an iterator over all peerings that have either masquerade or port-forwarding exposes
+    pub fn stateful_nat_peerings(&self) -> impl Iterator<Item = &Peering> {
+        self.peerings.iter().filter(|p| {
+            p.local
+                .exposes
+                .iter()
+                .any(|e| e.has_port_forwarding() || e.has_stateful_nat())
+                || p.remote
+                    .exposes
+                    .iter()
+                    .any(|e| e.has_port_forwarding() || e.has_stateful_nat())
+        })
+    }
 }
 
 #[derive(Clone, Debug, Default)]
--- a/config/src/external/overlay/vpc.rs
+++ b/config/src/external/overlay/vpc.rs
@@ -315,7 +315,7 @@ impl Vpc {
     }
 
     /// Provide an iterator over all peerings that have either masquerade or port-forwarding exposes locally
-    pub fn stateful_nat_peerings(&self) -> impl Iterator<Item = &Peering> {
+    pub fn local_stateful_nat_peerings(&self) -> impl Iterator<Item = &Peering> {
         self.peerings.iter().filter(|p| {
             p.local
                 .exposes
--- a/dataplane/src/packet_processor/mod.rs
+++ b/dataplane/src/packet_processor/mod.rs
@@ -37,6 +37,7 @@ where
 {
     pub router: Router,
     pub pipeline: Arc<dyn Send + Sync + Fn() -> DynPipeline<Buf>>,
+    pub flow_table: Arc<FlowTable>,
     pub vpcmapw: VpcMapWriter<VpcMapName>,
     pub nattablesw: NatTablesWriter,
     pub natallocatorw: NatAllocatorWriter,
@@ -44,7 +45,6 @@ where
     pub stats: StatsCollector,
     pub vpc_stats_store: Arc<VpcStatsStore>,
     pub portfw_w: PortFwTableWriter,
-    pub flow_table: Arc<FlowTable>,
 }
 
 /// Start a router and provide the associated pipeline
@@ -61,7 +61,6 @@ pub(crate) fn start_router<Buf: PacketBufferMut>(
 
     // create entities shared by management and data-path NFs
     let flow_table = Arc::new(FlowTable::default());
-    let flow_table_for_setup = flow_table.clone();
     let flowfiltertablesw = FlowFilterTableWriter::new();
     let flowfiltertablesr_factory = flowfiltertablesw.get_reader_factory();
     let nattablesw = NatTablesWriter::new();
@@ -90,8 +89,10 @@ pub(crate) fn start_router<Buf: PacketBufferMut>(
     let atabler_factory = router.get_atabler_factory();
 
     // create pipeline builder
+    let flow_table_clone = flow_table.clone();
     let pipeline_builder = move || {
         let pdata_clone = pdata.clone();
+
         // Build network functions
         let stage_ingress = Ingress::new("Ingress", iftr_factory.handle());
         let stage_egress = Egress::new("Egress", iftr_factory.handle(), atabler_factory.handle());
@@ -100,18 +101,18 @@ pub(crate) fn start_router<Buf: PacketBufferMut>(
         let stateless_nat = StatelessNat::with_reader("stateless-NAT", nattabler_factory.handle());
         let stateful_nat = StatefulNat::new(
             "stateful-NAT",
-            flow_table.clone(),
+            flow_table_clone.clone(),
             natallocator_factory.handle(),
         );
         let pktdump = PacketDumper::new("pipeline-end", true, None);
         let stats_stage = Stats::new("stats", stats_w.clone());
         let flow_filter = FlowFilter::new("flow-filter", flowfiltertablesr_factory.handle());
-        let icmp_error_handler = IcmpErrorHandler::new(flow_table.clone());
-        let flow_lookup = FlowLookup::new("flow-lookup", flow_table.clone());
+        let icmp_error_handler = IcmpErrorHandler::new(flow_table_clone.clone());
+        let flow_lookup = FlowLookup::new("flow-lookup", flow_table_clone.clone());
         let portfw = PortForwarder::new(
             "port-forwarder",
             portfw_factory.handle(),
-            flow_table.clone(),
+            flow_table_clone.clone(),
         );
         let pkt_stats_nf = PacketStatsNF::new(pkt_stats.clone());
 
@@ -137,6 +138,7 @@ pub(crate) fn start_router<Buf: PacketBufferMut>(
     Ok(InternalSetup {
         router,
         pipeline: Arc::new(pipeline_builder),
+        flow_table,
         vpcmapw,
         nattablesw,
         natallocatorw,
@@ -144,6 +146,5 @@ pub(crate) fn start_router<Buf: PacketBufferMut>(
         stats,
         vpc_stats_store,
         portfw_w,
-        flow_table: flow_table_for_setup,
     })
 }
--- a/flow-entry/src/flow_table/table.rs
+++ b/flow-entry/src/flow_table/table.rs
@@ -3,7 +3,7 @@
 
 use ahash::RandomState;
 use concurrency::sync::atomic::{AtomicUsize, Ordering};
-use concurrency::sync::{Arc, RwLock, RwLockReadGuard, Weak};
+use concurrency::sync::{Arc, RwLock, Weak};
 use dashmap::DashMap;
 use net::FlowKey;
 use net::flows::{FlowInfo, FlowStatus};
@@ -392,17 +392,6 @@ impl FlowTable {
                 .filter(|e| e.value().status() == FlowStatus::Active)
                 .count(),
         )
-    }
-
-    /// Take a read lock to the flow table
-    ///
-    /// # Panics
-    ///
-    /// This method panics if the `RwLock` is poisoned and may panic if, when called, the thread
-    /// already holds a lock to the table
-    #[allow(clippy::type_complexity)]
-    pub fn lock_read(&self) -> RwLockReadGuard<'_, DashMap<FlowKey, Arc<FlowInfo>, RandomState>> {
-        self.table.read().unwrap()
     }
 
     /// Execute a function for each flow in the table. This locks the table for reading.
--- a/flow-entry/src/flow_table/table.rs
+++ b/flow-entry/src/flow_table/table.rs
@@ -404,6 +404,84 @@ impl FlowTable {
     pub fn lock_read(&self) -> RwLockReadGuard<'_, DashMap<FlowKey, Arc<FlowInfo>, RandomState>> {
         self.table.read().unwrap()
     }
+
+    /// Execute a function for each flow in the table. This locks the table for reading.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if locking the table for reading fails
+    pub fn for_each_flow<F>(&self, mut func: F)
+    where
+        F: FnMut(&FlowKey, &FlowInfo),
+    {
+        let table = self.table.read().unwrap();
+        for flow in table.iter() {
+            func(flow.key(), &flow);
+        }
+    }
+
+    /// Same as `for_each_flow`, but allowing a filter to iterate only over the flows that match a predicate
+    ///
+    /// # Panics
+    ///
+    /// This function panics if locking the table for reading fails
+    pub fn for_each_flow_filtered<F, P>(&self, filter: P, mut func: F)
+    where
+        F: FnMut(&FlowKey, &FlowInfo),
+        P: Fn(&FlowKey, &FlowInfo) -> bool,
+    {
+        let table = self.table.read().unwrap();
+        for flow in table.iter().filter(|flow| filter(flow.key(), flow)) {
+            func(flow.key(), &flow);
+        }
+    }
+
+    /// Build an iterator of all flows in the table. Depending on how costly the processing of `f` in `for_each_flow`
+    /// is, taking a snapshot first may be faster. This is just possible because flow-info's are stored in 'Arc's.
+    /// There is little to no advantage of returning an iterator here because this method allocates anyway.
+    /// The snapshot can be restricted with the filter.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if locking the table for reading fails
+    pub fn snapshot<P>(&self, filter: P) -> impl Iterator<Item = Arc<FlowInfo>>
+    where
+        P: Fn(&FlowKey, &FlowInfo) -> bool,
+    {
+        let table = self.table.read().unwrap();
+        let v: Vec<_> = table
+            .iter()
+            .filter(|flow| filter(flow.key(), flow))
+            .map(|f| f.value().clone())
+            .collect();
+
+        v.into_iter()
+    }
+
+    /// FIXME: this does not provide any advantage
+    /// Need to interleave `reads()` with periods where we release lock/guard
+    /// I.e. need to chunk it
+    /// # Panics
+    ///
+    /// This function panics if locking the table for reading fails
+    pub fn for_each_flow_sharded<F>(&self, f: F)
+    where
+        F: Fn(&FlowKey, &FlowInfo),
+    {
+        let table = self.table.read().unwrap();
+        for shard in table.shards() {
+            let g = shard.read();
+            unsafe {
+                for (flowkey, flow_info) in g
+                    .iter()
+                    .map(|bucket| bucket.as_ref())
+                    .map(|(key, val)| (key, val.get().as_ref()))
+                {
+                    f(flowkey, flow_info);
+                }
+            }
+        }
+    }
 }
 
 #[cfg(test)]
--- a/flow-entry/src/flow_table/table.rs
+++ b/flow-entry/src/flow_table/table.rs
@@ -6,13 +6,13 @@ use dashmap::DashMap;
 use net::FlowKey;
 use std::fmt::Debug;
 use std::hash::Hash;
-use std::{borrow::Borrow, time::Duration};
-use tracing::debug;
 
 use concurrency::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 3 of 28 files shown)


qmonnet (changes requested):
Thanks a lot for this! The logics look good, I have mostly trivial comments.

You commented about “freezing the current allocator”, one concern is indeed how we ensure that no new flow is created in the flow table while we are busy validating the existing ones.

And we probably need some unit testing to mock a config change and check that we process the flows and allocations as expected.

dataplane/src/packet_processor/mod.rs
```diff
@@ -85,8 +86,10 @@ pub(crate) fn start_router<Buf: PacketBufferMut>(
     let atabler_factory = router.get_atabler_factory();
 
     // create pipeline builder
+    let flow_table_clone = flow_table.clone();
```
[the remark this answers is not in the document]
Fredi-raspall: It is needed because the clones() happen inside the closure, but we move the flow_table into it before those happen.

qmonnet:
Note: Compilation is broken due to changes in flow.rs between `feat(stateful-nat): add logic to invalidate / renew flows` and `feat(flow-table): let lock_read() panic`.

qmonnet:
I rebased the work in this branch on top of #1454 in https://github.com/githedgehog/dataplane/tree/pr/qmonnet/fredi-rebased-nat-fixes

flow-entry/src/flow_table/table.rs
```diff
@@ -349,6 +349,19 @@ impl FlowTable {
...
+    /// Returns `PoisonError` if the lock was poisoned.
+    #[allow(clippy::type_complexity)]
+    pub fn lock_read(
+        &self,
+    ) -> Result<
+        RwLockReadGuard<'_, DashMap<FlowKey, Weak<FlowInfo>, RandomState>>,
+        PoisonError<RwLockReadGuard<'_, DashMap<FlowKey, Weak<FlowInfo>, RandomState>>>,
+    > {
+        self.table.read()
+    }
```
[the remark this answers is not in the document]
Fredi-raspall: This is outdated and will be ignored.

nat/src/stateful/apalloc/alloc.rs
```diff
@@ -486,19 +451,11 @@ pub(crate) fn map_offset(
...
-        ))?;
-
-    Ok(prefix_offset
-        + u32::try_from(address.to_bits() - prefix_start_bits).map_err(|_| {
-            AllocatorError::InternalIssue("Failed to convert Ipv6 to offset".to_string())
-        })?)
+        .expect("This should never fail");
+
+    prefix_offset + u32::try_from(address.to_bits() - prefix_start_bits).unwrap()
 }
```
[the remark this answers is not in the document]
Fredi-raspall: This will be fixed. The whole point of the change is precisely getting rid of the results since these should never happen. Those are not things that can fail but bugs.

nat/src/stateful/flows.rs
```diff
@@ -0,0 +1,208 @@
...
+
+use crate::NatPort;
+use crate::stateful::allocator_writer::StatefulNatConfig;
+use crate::stateful::natip::NatIp;
+use crate::stateful::{NatDefaultAllocator, NatFlowState};
+
+use config::GenId;
+use flow_entry::flow_table::FlowTable;
+use lpm::prefix::IpRangeWithPorts;
+use lpm::prefix::PrefixWithOptionalPorts;
```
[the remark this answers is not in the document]
Fredi-raspall: This is outdated

nat/src/stateful/allocator_writer.rs
```diff
@@ -65,80 +119,48 @@ impl NatAllocatorWriter {
...
+
+        // if we transition to a config wo/ masquerading, flush allocator and remove flows
+        if nat_config.num_masquerading_peerings() == 0 {
+            if curr_allocator.is_some() {
+                debug!("No stateful NAT is required anymore: will invalidate flows");
+                invalidate_all_masquerading_flows(flow_table);
+                self.allocator.store(None);
+            }
+            return;
         }
```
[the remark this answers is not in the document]
Fredi-raspall: Yup. Forgot to add that. Good catch.

flow-entry/src/flow_table/table.rs
```diff
@@ -299,6 +299,84 @@ impl FlowTable {
...
+            let g = shard.read();
+            unsafe {
+                for (flowkey, flow_info) in g
+                    .iter()
+                    .map(|bucket| bucket.as_ref())
+                    .map(|(key, val)| (key, val.get().as_ref()))
+                {
+                    f(flowkey, flow_info);
+                }
+            }
```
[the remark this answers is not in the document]
Fredi-raspall: As annotated in the code, that method is unused / unusable atm. So, I will ignore this comment.

nat/src/stateful/flows.rs
```diff
@@ -0,0 +1,226 @@
...
+
+/// Main function called to deal with flows when masquerade configuration changes. This function:
+///   - locks the flow table
+///   - examines all masquerade flows to determine if they should continue or be invalidated
+///   - flows that continue get a new allocation with the same ip and port in the new allocator
+pub(crate) fn check_masquerading_flows(
+    flow_table: &FlowTable,
+    new_config: &StatefulNatConfig,
+    new_allocator: &mut NatAllocator,
+) {
```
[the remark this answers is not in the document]
Fredi-raspall: Very true. The whole test strategy of nat needs to be revisited. So, ignoring for now.

nat/src/stateful/flows.rs
```diff
@@ -0,0 +1,226 @@
...
+// Copyright Open Network Fabric Authors
+
+use crate::NatPort;
+use crate::stateful::NatFlowState;
+use crate::stateful::allocator_writer::StatefulNatConfig;
+use crate::stateful::apalloc::NatAllocator;
+
+use config::GenId;
+use flow_entry::flow_table::FlowTable;
+use lpm::prefix::IpRangeWithPorts;
```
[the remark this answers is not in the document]
Fredi-raspall: The type is actually used. Ignoring...

nat/src/stateful/allocator_writer.rs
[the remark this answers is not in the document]
Fredi-raspall: Added.

nat/src/stateful/flows.rs
```diff
@@ -0,0 +1,227 @@
...
+pub(crate) fn invalidate_all_stateful_nat_flows(flow_table: &FlowTable) {
+    debug!("INVALIDATING all masquerading flows...");
+    let mut count = 0;
+    let table_lock = flow_table.lock_read().unwrap();
+    for flow_info in table_lock.iter().filter_map(|flow| flow.upgrade()) {
+        if let Ok(locked) = flow_info.locked.read()
+            && locked.nat_state.as_ref().is_some()
+        {
+            flow_info.invalidate_pair();
+            count += 1;
```
qmonnet, line 30: It looks like you're processing all entries in the flow table, so maybe you don't need to invalidate by pairs (or you'll end up invalidating all flows twice)?

(Same comment for genid upgrade.)

nat/src/stateful/flows.rs
```diff
@@ -0,0 +1,227 @@
...
+        return None;
+    };
+    let data = src_alloc.translation_data();
+    Some((data.src_addr?, data.src_port?))
+}
+
+#[inline]
+fn flow_ipv6_masquerade_state(locked: &FlowInfoLocked) -> Option<(IpAddr, NatPort)> {
+    let state = locked.nat_state.as_ref();
+    let src_state = state.extract_ref::<NatFlowState<Ipv6Addr>>()?;
```
qmonnet, line 69: Does this fail for IPv4? I don't know how the `extract_ref()` method behaves, exactly.

nat/src/stateful/flows.rs
```diff
@@ -0,0 +1,227 @@
...
+        // used to masquerade the flow.
+        debug!("Peering exists between {src_vpcd} and {dst_vpcd}");
+        if nat_peering.peering.local.exposes.iter().any(|e| {
+            e.has_stateful_nat()
+                && e.nat.as_ref().is_some_and(|enat| {
+                    enat.as_range
+                        .iter()
+                        .any(|pfx| pfx.covers(&PrefixWithOptionalPorts::Prefix(ip.into())))
+                })
+        }) {
```
qmonnet: (Nit: We could consider moving this to a dedicated method in the `VpcExpose`.)

mgmt/src/processor/proc.rs
```diff
@@ -504,16 +504,16 @@ fn apply_stateless_nat_config(
...
 
 /// Update the config for stateful NAT.
+/// This is now infallible. Validation should ensure it is.
 fn apply_stateful_nat_config(
     vpc_table: &VpcTable,
     flow_table: &FlowTable,
     natallocatorw: &mut NatAllocatorWriter,
     genid: GenId,
-) -> ConfigResult {
+) {
```
qmonnet, line 513: Based on the recent discussions, should we drop this commit?

nat/src/stateful/allocator_writer.rs
```diff
@@ -22,12 +22,18 @@ pub(crate) struct StatefulNatPeering {
...
+    genid: GenId,
     peerings: Vec<StatefulNatPeering>,
     randomize: bool,
 }
+impl PartialEq for StatefulNatConfig {
+    fn eq(&self, other: &Self) -> bool {
+        // we exclude genid from comparison
+        self.peerings == other.peerings && self.randomize == other.randomize
+    }
+}
```
qmonnet, line 36: I've got mixed feelings about this, it would seem more natural to me to derive `PartialEq` so that it takes all fields into account, and to implement a special comparison method just to exclude the `genid`, because I can easily see someone trying to compare two `StatefulNatConfig` objects with the `==` operator in the future and introducing some bug because they may not realise the `genid` may not be the same.

Not blocking on this, though

flow-entry/src/flow_table/table.rs
```diff
@@ -396,16 +395,14 @@ impl FlowTable {
     }
 
     /// Take a read lock to the flow table
-    /// # Errors
-    /// Returns `PoisonError` if the lock was poisoned.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if the `RwLock` is poisoned and may panic if, when called, the thread
+    /// already holds a lock to the table
```
qmonnet: Ack

nat/src/stateful/apalloc/setup.rs
```diff
@@ -25,7 +25,7 @@ const DEFAULT_MASQUERADE_IDLE_TIMEOUT: Duration = Duration::from_mins(2);
 impl NatAllocator {
     /// Build a [`NatAllocator`] from a [`StatefulNatConfig`]
     pub(crate) fn from_config(config: &StatefulNatConfig) -> Self {
-        debug!("Building NAT allocator from config:\n{config:#?}");
+        debug!("Building stateful NAT allocator from config:\n{config:#?}");
```
qmonnet, line 28: Nit: `masquerading (NAT)` instead? Doesn't really matter anyway, we can fix when we do a larger “stateful NAT” -> “masquerading” clean-up.

config/src/external/overlay/vpc.rs
```diff
@@ -314,17 +314,13 @@ impl Vpc {
...
     pub fn stateful_nat_peerings(&self) -> impl Iterator<Item = &Peering> {
         self.peerings.iter().filter(|p| {
             p.local
                 .exposes
                 .iter()
                 .any(|e| e.has_port_forwarding() || e.has_stateful_nat())
-                || p.remote
-                    .exposes
-                    .iter()
-                    .any(|e| e.has_port_forwarding() || e.has_stateful_nat())
```
qmonnet: Please rename the method, it still reads like it returns all peerings with stateful NAT, it may be confusing if someone attempts to reuse it. `local_stateful_nat_peerings`? `stateful_nat_peering_pairs`?

nat/src/stateful/allocator_writer.rs
```diff
@@ -79,10 +79,7 @@ impl StatefulNatConfig {
         self.peerings
             .iter()
             .map(|p| &p.peering)
-            .filter(|p| {
-                p.local.exposes.iter().any(VpcExpose::has_stateful_nat)
-                    || p.remote.exposes.iter().any(VpcExpose::has_stateful_nat)
-            })
+            .filter(|p| p.local.exposes.iter().any(VpcExpose::has_stateful_nat))
             .count()
     }
```
qmonnet: Same, please rename the method to account for the fact we count only half of the peerings involved in masquerading.

nat/src/stateful/apalloc/setup.rs
```diff
@@ -239,10 +239,12 @@ fn build_reserved_prefixes_ports(
...
-        reserved_prefixes_ports.insert(
-            prefix.prefix().into(),
-            prefix.ports().expect("Missing ports. This is a bug"),
-        );
+        debug_assert!(prefix.ports().is_some());
+        let Some(ports) = prefix.ports() else {
+            error!("Stepped on a port-forwarding prefix without ports. This is a bug");
+            continue;
+        };
+        reserved_prefixes_ports.insert(prefix.prefix().into(), ports);
```
qmonnet, line 247: > Ideally, such conditions would be detected at validation time,
but this requires significant refactoring atm.

Do you mean introducing a validated type? Because just checking that prefixes for port forwarding come with a port number should already be part of the validation step, right?

nat/src/stateful/allocator_writer.rs
```diff
@@ -65,80 +119,50 @@ impl NatAllocatorWriter {
...
+        let mut allocator = NatAllocator::from_config(&nat_config);
+        if old_allocator.is_some() {
+            check_masquerading_flows(flow_table, &nat_config, &mut allocator);
+        }
+        // make new allocator visible
+        debug!("Installing new stateful NAT allocator...");
+        self.allocator.store(Some(Arc::new(allocator)));
+        self.config = nat_config;
+        debug!("Updated stateful NAT allocator");
     }
```
qmonnet, line 165: So what happens if:

1. Thread 1: Packet arrives to the stateful NAT pipeline stage
2. Thread 1: Packet gets allocated an IP and port from the old allocator, with the old config
3. Thread 2: We get and process a config change: we swap the allocator, we invalidate the flow table entries that are no longer valid
4. Thread 2: We release the `read` lock for the flow table
5. Thread 1 (Before or after we swap the allocator, it doesn't matter): We call `create_flow_pair()` in the NAT pipeline, using the new value from `self.pipeline_data.genid()`, and we insert the flow info, from the old allocator, now invalid, into the flow table, with the new `genid`?

Then any follow-up packets from the now-invalid flow (per the new config) are allowed to pass?

I think at the very least we need a check in stateful NAT's `create_flow_pair()` to ensure the `genid` is still the one we found when entering the `translate_packet()` function. Even better, it would be good to have a way to prevent the insertion in the flow table of entries with older `genid` values.

qmonnet (changes requested):
I've got several small comments; plus one larger concern about a potential (although not-super-likely) race condition, see [the last comment](https://github.com/githedgehog/dataplane/pull/1414#discussion_r3111948530) below.

I think these changes could really do with some tests, maybe even Shuttle tests to make sure we can't insert older and invalid entries into the flow table after a config change.

nat/src/stateful/allocator_writer.rs
```diff
@@ -65,80 +105,35 @@ impl NatAllocatorWriter {
...
+            self.allocator.store(None);
+            return;
+        }
 
-    fn build_new_allocator(config: &StatefulNatConfig) -> Result<NatDefaultAllocator, ConfigError> {
-        NatDefaultAllocator::build_nat_allocator(config)
-    }
+        // build a new allocator to replace the current
+        let mut allocator =
+            NatDefaultAllocator::from_config(&nat_config).set_randomize(nat_config.randomize);
```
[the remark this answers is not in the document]
qmonnet: Was this addressed? Please mark the conversation as resolved in that case.

nat/src/stateful/flows.rs
Fredi-raspall: Yes. I am aware of this. I wanted to make sure ALL were invalidated. I can change that.

nat/src/stateful/flows.rs
Fredi-raspall: It should "fail" in the sense that it should return None.

nat/src/stateful/flows.rs
Fredi-raspall: I am going to postpone changing this. True, we end up invalidating twice, but we invalidate both directions at the same time. I think the best approach here would be to actually change the invalidate method so that if a flow is invalid, it does not do anything, Will address this in the PR that follows.

nat/src/stateful/flows.rs
Fredi-raspall: I'd leave it as it is since it is only used here.

mgmt/src/processor/proc.rs
Fredi-raspall: I think I'd keep it as it is closer to what we want.
If we leave it as it is and there's a bug, it will lead to malfunction.
If we keep it fallible, we'll rollback to the prior config. However that's equally bad in our case since we'll remain there forever. Ideally, we would compute the state needed as part of the validation or extend the validation so that this would be impossible to happen if validation succeeded.

nat/src/stateful/allocator_writer.rs
Fredi-raspall: We do this in several other places and I believe it is a common practice. I'll think about it and change it in the following PR.

nat/src/stateful/apalloc/setup.rs
Fredi-raspall: Well, allocator is only for masquerading. None of the rest of NAT flavors use it so :shrug:

nat/src/stateful/allocator_writer.rs
Fredi-raspall: We're iterating over all peerings.

nat/src/stateful/apalloc/setup.rs
Fredi-raspall: No, I meant that this should not fail at this stage. We should have caught this sooner. If that's already the case, then we're good.

nat/src/stateful/allocator_writer.rs
Fredi-raspall: yes, we have several situations like that from the moment that the config of the distinct NFs aren't applied atomically. I will address it in the PR that follows.

nat/src/stateful/allocator_writer.rs
qmonnet: We're still only counting those that have stateful NAT on the local end, though?

Actually, looking at the code again, do you need to count at all? You seem to only compare the result with zero, so you could drop the `.count()` and leave the `.any(...)` return a `bool`, it would allow the code to return faster when it finds one peering with stateful NAT?

nat/src/stateful/apalloc/setup.rs
qmonnet: We have the relevant checks in the k8s converter, in `nat_expand_rules()`, so I guess we're good, OK.

mgmt/src/processor/proc.rs
qmonnet: > If we keep it fallible, we'll rollback to the prior config. However that's equally bad in our case since we'll remain there forever.

Not sure I follow, a rollback to a previous working config sounds better to me than a non-working update?

nat/src/stateful/allocator_writer.rs
qmonnet: How do you plan to address it, please?

qmonnet:
[The failure in the sanitizer](https://github.com/githedgehog/dataplane/actions/runs/24688622182/job/72205428105?pr=1414) on `fib::test::tests::test_concurrency_fibtable` looks unrelated to the PR (not touching `routing`), although we'll need to look into it at some point.

qmonnet (changes requested):
I still believe some tests would be useful to validate the new logic

mgmt/src/processor/proc.rs
Fredi-raspall: Generally speaking yes. Recall however, that in our case, we have no notification to k8s. That's the reason for the whole validation story.

nat/src/stateful/allocator_writer.rs
Fredi-raspall: > Same, please rename the method to account for the fact we count only half of the peerings involved in masquerading.

We're stepping on all masquerading exposes, even if we only check "local" because the "remote" are the "local" for other vpcs. Since we check ALL vpcs, we should only check "local" (or remote if you want).

> Actually, looking at the code again, do you need to count at all? You seem to only compare the result with zero, so you could drop the .count() and leave the .any(...) return a bool, it would allow the code to return faster when it finds one peering with stateful NAT?

Yes, that can be done.

nat/src/stateful/allocator_writer.rs
Fredi-raspall: There's multiple ways we may do so. However, I think it's pointless addressing this here given the changes in https://github.com/githedgehog/dataplane/pull/1475

Fredi-raspall:
> I still believe some tests would be useful to validate the new logic

Fair enough. Will add them in https://github.com/githedgehog/dataplane/pull/1475

nat/src/stateful/allocator_writer.rs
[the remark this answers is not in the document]
Fredi-raspall: I think it was!

nat/src/stateful/allocator_writer.rs
```diff
@@ -65,80 +118,50 @@ impl NatAllocatorWriter {
...
-        //
-        // One "option" is to process all entries in the session table, look at the new
-        // configuration (or the new allocator entries) to see if they're still valid, and then
-        // report them to the new allocator. However, the old allocator keeps being updated during
-        // this process.
-        Self::build_new_allocator(new_config)
+        // pull the current allocator out of the data path. While we build the new allocator,
+        // no reservation will be possible. However, this is better than adding any locking in data path.
+        // New flows requiring masquerading won't get any IP/port. That's fine, they will retry.
+        // We don't yet drop the old allocator: the flows that used it will in fact keep it alive until
```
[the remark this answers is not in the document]
Fredi-raspall: The comment is fine as it is. Packets will be dropped. Originators (they) will retry sending them.

net/src/packet/mod.rs
```diff
@@ -132,6 +132,13 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
         Some(flow_info)
     }
 
+    /// Invalidate the flow that a packet refers to if any, and the related flow
```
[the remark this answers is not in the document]
Fredi-raspall: I think it is clear as it is.

qmonnet (approved):
Looks good, but please make sure to follow up with the tests and fixes for the potential race in the flow table + allocator update, as we discussed. Thanks a lot!

Issue #1280, opened 2026-02-13
Peering config validation: Reject reserved prefixes, port 0


qmonnet:
See items 21 and 22 on https://github.com/githedgehog/dataplane/issues/1150

qmonnet:
This is not completed. We have addressed L4 port 0, but not the reserved IP prefixes.

sergeymatov:
@qmonnet do we need prevent LPM crate to return error if `PortRange::new()` allows `0` as `start` or stateful nat should manage it
Homework for all: there is an existing `well known list` of services with IANA defined ports. So the question is - do we need to avoid allocating ports from this range?

qmonnet:
> do we need prevent LPM crate to return error if `PortRange::new()` allows `0` as `start` or stateful nat should manage it

Sorry, I'm not sure I understand what you mean exactly? What function from LPM are you thinking of, exactly?

> Homework for all: there is an existing `well known list` of services with IANA defined ports. So the question is - do we need to avoid allocating ports from this range?

I don't think it matters too much, especially because the ports are allocated when initiating requests - so, on the client side, not server side, so presumably the port numbers don't matter as much? So I wouldn't restrict it for now.

sergeymatov:
> Sorry, I'm not sure I understand what you mean exactly? What function from LPM are you thinking of, exactly?

The `PortRange::new()` :) 

> I don't think it matters too much, especially because the ports are allocated when initiating requests - so, on the client side, not server side, so presumably the port numbers don't matter as much? So I wouldn't restrict it for now.

I think RFC7618 `Dynamic Allocation of Shared IPv4 Addresses` states that block from 0 to 1023 should be preserved from the dynamic allocation, or at least this is comes in my mind after reading chapter 8.

qmonnet:
> The `PortRange::new()` :)

I don't see why `PortRange::new()` itself should error when passing `0` for the start number. The object is not protocol-dependant, we could imagine using it for some protocol where 0 is valid. I don't remember for sure but we might be using it with ICMP, for the identifier field, where 0 is a valid value.

> I think RFC7618 `Dynamic Allocation of Shared IPv4 Addresses` states that block from 0 to 1023 should be preserved from the dynamic allocation, or at least this is comes in my mind after reading chapter 8.

This RFC is not related to NAT, though?

sergeymatov:
> This RFC is not related to NAT, though?

[...]

Issue #995, opened 2025-11-07
Stateful NAT: Improve ICMP support in allocator


qmonnet:
Right now, the stateful NAT allocator reuses the TCP/UDP ports fields for allocating values for ICMP identifiers. It would be better to have a cleaner design, and only allocate a single value in the case of ICMP.

Another question about ICMP, although not related to the allocator, is how we should translate (if at all) packets such as ICMP "Destination Error" messages when they _do not_ have any inner packet (right now we translate ICMP Query messages, and ICMP Error messages with inner packets - should we do something particular for other packets?)

PR #1419, opened 2026-03-27
Preliminary work for static NAT source/destination split

qmonnet:
As the title says, this is some preliminary work for splitting the current static NAT stage into two steps, source and destination NAT. This will be necessary for supporting static NAT + masquerade at opposite ends of a peering. Here we do not split the stage, but we prepare for that by moving source and destination translation into individual functions. The biggest chunk of these changes is the modifications brought to the code for translating ICMP Error messages' inner packets (and the biggest part of the diff is a not-so-related clean-up in NAT tests, but it's mostly an indentation change).

~Will need to be rebased on top of #1414 and possibly #1389.~ I'll handle conflicts if necessary.

### DIFF
--- a/nat/src/icmp_handler/icmp_error_msg.rs
+++ b/nat/src/icmp_handler/icmp_error_msg.rs
@@ -7,7 +7,6 @@
 use crate::NatPort;
 use crate::NatTranslationData;
 use net::buffer::PacketBufferMut;
-use std::num::NonZero;
 use net::checksum::{Checksum, ChecksumError};
 use net::headers::{
     EmbeddedTransport, TryEmbeddedHeaders, TryEmbeddedHeadersMut, TryEmbeddedTransportMut,
@@ -18,6 +17,7 @@ use net::icmp_any::{IcmpAnyChecksumErrorPlaceholder, IcmpAnyChecksumPayload};
 use net::ipv4::Ipv4;
 use net::packet::Packet;
 use std::net::IpAddr;
+use std::num::NonZero;
 
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 pub enum IcmpErrorMsgError {
@@ -39,23 +39,19 @@ pub enum IcmpErrorMsgError {
 
 // # Return
 //
+// * `Ok(())` if checksums are valid and we can translate the inner packet
 // * An error if we fail to validate relevant checksums and packet should be dropped
-// * `true` if checksums are valid and we need to translate the inner packet
-// * `false` if we don't need to translate the inner packet
 pub(crate) fn validate_checksums_icmp<Buf: PacketBufferMut>(
     packet: &Packet<Buf>,
-) -> Result<bool, IcmpErrorMsgError> {
+) -> Result<(), IcmpErrorMsgError> {
     let Some(net) = packet.try_ip() else {
-        // No network layer, no translation needed
-        return Ok(false);
+        return Err(IcmpErrorMsgError::BadIpHeader);
     };
     let Some(icmp) = packet.try_icmp_any() else {
-        // Not ICMPv4 or ICMPv6, no translation needed
-        return Ok(false);
+        return Err(IcmpErrorMsgError::NoTranslationPossible);
     };
     if !icmp.is_error_message() {
-        // Not an ICMP error message, no translation needed
-        return Ok(false);
+        return Err(IcmpErrorMsgError::NoTranslationPossible);
     }
 
     let icmp_payload =
@@ -71,7 +67,7 @@ pub(crate) fn validate_checksums_icmp<Buf: PacketBufferMut>(
 
     let Some(embedded_ip) = packet.embedded_headers() else {
         // No embedded IP packet to translate
-        return Ok(false);
+        return Err(IcmpErrorMsgError::NoTranslationPossible);
     };
 
     // From REQ-3 a) from RFC 5508, "NAT Behavioral Requirements for ICMP":
@@ -85,8 +81,7 @@ pub(crate) fn validate_checksums_icmp<Buf: PacketBufferMut>(
             .validate_checksum(&())
             .map_err(IcmpErrorMsgError::BadChecksumInnerIpv4)?;
     }
-
-    Ok(true)
+    Ok(())
 }
 
 pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
@@ -293,7 +288,7 @@ mod tests {
         let packet = Packet::new(buffer).unwrap();
 
         let result = validate_checksums_icmp(&packet);
-        assert_eq!(result, Ok(false));
+        assert_eq!(result, Err(IcmpErrorMsgError::BadIpHeader));
     }
 
     #[test]
@@ -312,7 +307,7 @@ mod tests {
         let packet = Packet::new(buffer).unwrap();
 
         let result = validate_checksums_icmp(&packet);
-        assert_eq!(result, Ok(false));
+        assert_eq!(result, Err(IcmpErrorMsgError::NoTranslationPossible));
     }
 
     #[test]
@@ -339,7 +334,7 @@ mod tests {
         let packet = Packet::new(buffer).unwrap();
 
         let result = validate_checksums_icmp(&packet);
-        assert_eq!(result, Ok(false));
+        assert_eq!(result, Err(IcmpErrorMsgError::NoTranslationPossible));
     }
 
     #[test]
@@ -363,7 +358,7 @@ mod tests {
         let packet = Packet::new(buffer).unwrap();
 
         let result = validate_checksums_icmp(&packet);
-        assert_eq!(result, Ok(false));
+        assert_eq!(result, Err(IcmpErrorMsgError::NoTranslationPossible));
     }
 
     #[test]
@@ -402,7 +397,7 @@ mod tests {
         let packet = Packet::new(buffer).unwrap();
 
         let result = validate_checksums_icmp(&packet);
-        assert_eq!(result, Ok(false));
+        assert_eq!(result, Err(IcmpErrorMsgError::NoTranslationPossible));
     }
 }
 
@@ -464,7 +459,7 @@ mod bolero_tests {
 
                 // Now, ICMP and inner IP headers checksums should be valid
                 let res = validate_checksums_icmp::<TestBuffer>(&icmp_error_msg_clone);
-                assert_eq!(res, Ok(true), "Checksum validation failed: {res:?}");
+                assert!(res.is_ok(), "Checksum validation failed: {res:?}");
 
                 // Also check outer IP header checksum, since we're at it
                 if let Some(ipv4) = icmp_error_msg_clone.headers().try_ipv4() {
@@ -608,7 +603,7 @@ mod bolero_tests {
                     // Update and validate checksums for inner IP header, ICMP header, and outer IP header
                     icmp_error_msg_clone.update_checksums();
                     let res = validate_checksums_icmp::<TestBuffer>(&icmp_error_msg_clone);
-                    assert_eq!(res, Ok(true), "Checksum validation failed: {res:?}");
+                    assert!(res.is_ok(), "Checksum validation failed: {res:?}");
                 },
             );
     }
--- a/nat/src/icmp_handler/icmp_error_msg.rs
+++ b/nat/src/icmp_handler/icmp_error_msg.rs
@@ -7,6 +7,7 @@
 use crate::NatPort;
 use crate::NatTranslationData;
 use net::buffer::PacketBufferMut;
+use std::num::NonZero;
 use net::checksum::{Checksum, ChecksumError};
 use net::headers::{
     EmbeddedTransport, TryEmbeddedHeaders, TryEmbeddedHeadersMut, TryEmbeddedTransportMut,
@@ -32,8 +33,8 @@ pub enum IcmpErrorMsgError {
     InvalidIpVersion,
     #[error("IP address {0} is not unicast")]
     NotUnicast(IpAddr),
-    #[error("no allocated identifier found for translation")]
-    NoIdentifier,
+    #[error("no translation possible")]
+    NoTranslationPossible,
 }
 
 // # Return
@@ -98,9 +99,6 @@ pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
         state.src_port,
         state.dst_port,
     );
-    let embedded_headers = packet
-        .embedded_headers_mut()
-        .ok_or(IcmpErrorMsgError::BadIpHeader)?;
 
     // From REQ-4 from RFC 5508, "NAT Behavioral Requirements for ICMP":
     //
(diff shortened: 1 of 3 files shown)


nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -32,29 +33,28 @@ pub enum IcmpErrorMsgError {
...
     let Some(icmp) = packet.try_icmp_any() else {
         // Not ICMPv4 or ICMPv6, no translation needed
-        return Ok(false);
+        return Err(IcmpErrorMsgError::NoTranslationPossible);
     };
     if !icmp.is_error_message() {
         // Not an ICMP error message, no translation needed
-        return Ok(false);
+        return Err(IcmpErrorMsgError::NoTranslationPossible);
     }
```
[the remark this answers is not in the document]
qmonnet: Right, I'll remove the comments.

nat/src/stateless/mod.rs
```diff
@@ -310,12 +275,37 @@ impl StatelessNat {
             }
         }
 
-        if packet.is_icmp() {
-            modified |= Self::translate_icmp_inner_packet_if_any(table, packet, dst_vni)?;
+        // ICMP Error messages
+        if packet.is_icmp_inner_translation_candidate() {
```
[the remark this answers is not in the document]
qmonnet: Certainly not. At this stage we've updated the inner destination address, but not the corresponding checksum, so the checksum is not valid at that point. This will need some additional care when uncoupling the function totally, but this is left as a future work.

nat/src/stateless/mod.rs
```diff
@@ -127,30 +125,19 @@ impl StatelessNat {
             .try_transport_mut()
             .ok_or(StatelessNatError::UnsupportedTranslation)?;
         match transport {
```
daniel-noland, line 125: This logic is a bug I think.

Just because the packet has a transport layer does not make it legit for nat.  e.g., if somebody sends us VLAN tagged traffic we will NAT it (we shouldn't even route it)

nat/src/stateless/mod.rs
daniel-noland, line 119: technically out of scope a bit since this line didn't change, bu as a private method shouldn't this accept a `NonZero<u16>` and then we push all the error handling to the outside?  The whole need for the `UnsupportedTranslation` seems to revolve around not checking that first

nat/src/stateless/mod.rs
```diff
@@ -127,30 +125,19 @@ impl StatelessNat {
...
-                );
-            }
-            Transport::Icmp4(_icmp4) => {
-                todo!()
+                packet
+                    .set_source_port(
+                        NonZero::try_from(new_port)
+                            .map_err(|_| StatelessNatError::UnsupportedTranslation)?,
+                    )
+                    .map_err(|_| StatelessNatError::UnsupportedTranslation)?;
```
daniel-noland: this isn't ideal.  We loose a lot of info in our error here.  We don't know if the port was zero of if we failed to translate because the ICMP xlate failed or what

nat/src/stateless/mod.rs
```diff
@@ -167,30 +154,19 @@ impl StatelessNat {
...
-                );
-            }
-            Transport::Icmp4(_icmp4) => {
-                todo!()
+                packet
+                    .set_destination_port(
+                        NonZero::try_from(new_port)
+                            .map_err(|_| StatelessNatError::UnsupportedTranslation)?,
+                    )
+                    .map_err(|_| StatelessNatError::UnsupportedTranslation)?;
```
daniel-noland: more error message data loss

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -331,13 +329,17 @@ mod tests {
...
+        let mut icmp = Icmp4::with_type(Icmp4Type::DestUnreachable(Icmp4DestUnreachable::Network));
+        icmp.update_checksum(&[]).unwrap();
 
+        headers.eth(Some(make_default_for_eth(EthType::IPV4)));
         headers.net(Some(Net::Ipv4(ipv4)));
         headers.transport(Some(Transport::Icmp4(icmp)));
 
         let headers = headers.build().unwrap();
-        let mut buffer = TestBuffer::new();
+        // Use a zeroed buffer so that ICMP payload that assumed null payload is correct
```
daniel-noland: not sure what this comment means

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -225,27 +225,22 @@ fn translate_inner_tcp_udp(
...
     use net::icmp4::{Icmp4DestUnreachable, Icmp4EchoRequest, Icmp4Type};
     use net::ip::NextHeader;
     use net::ipv4::Ipv4;
     use net::packet::Packet;
+    use net::packet::test_utils::make_default_for_eth;
     use net::parse::DeParse;
     use std::net::Ipv4Addr;
 
     #[test]
     fn test_validate_checksums_icmp_no_network_layer() {
```
daniel-noland, line 285: technically out of scope for this change, but this test is a great candidate for

1. fuzzing
2. the new packet builder

net/src/packet/utils.rs
```diff
@@ -181,6 +181,19 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
...
+        // If no network layer, no translation needed
+        self.try_ip().is_some()
+        // If not ICMPv4 or ICMPv6, no translation needed
+        && self.try_icmp_any().is_some_and(|icmp| {
+            // If not an ICMP error message, no translation needed
+            icmp.is_error_message()
+        })
+        // If no embedded IP packet, no translation needed
+        && self.embedded_headers().is_some()
+    }
```
daniel-noland: check does not account for vlan headers

nat/src/stateless/mod.rs
```diff
@@ -167,118 +135,89 @@ impl StatelessNat {
...
+            .ok_or(StatelessNatError::NoIpHeader)?
+            .src_addr();
+        let port = packet
+            .try_embedded_transport()
+            .and_then(|t| t.source().map(NonZero::get));
+        // Note: we assign the _destination_ mapping to the target _source_ address.
+        // We're sending the inner packet back without swapping source and destination in the
+        // header, so we need to swap the ranges we get from the tables lookup.
+        let Some((src_addr, src_port)) = table.find_dst_mapping(&addr, port) else {
+            return Err(StatelessNatError::UnsupportedTranslation);
```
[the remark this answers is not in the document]
qmonnet: Previous version of the function would translate source _or_ destination so we could have one direction that didn't require translation. Now that it translates the source only, we always need a mapping, or we should drop the packet; so this is expected.

nat/src/stateless/mod.rs
```diff
@@ -167,118 +135,89 @@ impl StatelessNat {
...
+            .ok_or(StatelessNatError::NoIpHeader)?
+            .dst_addr();
+        let port = packet
+            .try_embedded_transport()
+            .and_then(|t| t.destination().map(NonZero::get));
+        // Note: we assign the _source_ mapping to the target _destination_ address.
+        // We're sending the inner packet back without swapping source and destination in the
+        // header, so we need to swap the ranges we get from the tables lookup.
+        let Some((dst_addr, dst_port)) = table.find_src_mapping(&addr, port, dst_vni) else {
             return Err(StatelessNatError::UnsupportedTranslation);
```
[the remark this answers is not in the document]
qmonnet: Similar to comment above.

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -110,114 +108,162 @@ pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
...
-            target_src_ip
+
+    embedded_headers
+        .try_inner_ip_mut()
+        .ok_or(IcmpErrorMsgError::BadIpHeader)?
+        .try_set_source(
+            target_addr
                 .try_into()
-                .map_err(|_| IcmpErrorMsgError::NotUnicast(target_src_ip))?,
+                .map_err(|_| IcmpErrorMsgError::NotUnicast(target_addr))?,
```
daniel-noland, line 134: do we have a check to exclude `0.0.0.0` or `::`?

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -110,114 +108,162 @@ pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
...
+    }
+    if let Some(dst_addr) = target_dst_addr {
+        nat_translate_icmp_inner_dst(packet, dst_addr, target_dst_port)?;
+    }
+    Ok(())
+}
+
+pub(crate) fn nat_translate_icmp_inner_src<Buf: PacketBufferMut>(
+    packet: &mut Packet<Buf>,
+    target_addr: IpAddr,
```
daniel-noland, line 121: why not accept `UnicastIpAddr` instead of all the extra error handling and branches?

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -110,114 +108,162 @@ pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
...
+    let embedded_headers = packet
+        .embedded_headers_mut()
         .ok_or(IcmpErrorMsgError::BadIpHeader)?;
-    if let Some(target_src_ip) = target_src_addr {
-        net.try_set_source(
-            target_src_ip
+
+    embedded_headers
+        .try_inner_ip_mut()
+        .ok_or(IcmpErrorMsgError::BadIpHeader)?
```
daniel-noland: isn't this more of a "no ip header" situation?

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -110,114 +108,162 @@ pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
...
+    Ok(())
+}
 
-    let net = embedded_headers
-        .try_inner_ip_mut()
+pub(crate) fn nat_translate_icmp_inner_dst<Buf: PacketBufferMut>(
+    packet: &mut Packet<Buf>,
+    target_addr: IpAddr,
+    target_port: Option<NatPort>,
+) -> Result<(), IcmpErrorMsgError> {
```
daniel-noland, line 165: this type of thing would be _much_ easier as a method which accepted an `Ipv4`, an `Icmp4`  and a `UnicastIpv4Addr` + `Option<NatPort>` + another method with ipv6 equivalents

As is, this

1. requires the method to be generic over the buffer (which is not needed)
2. requires a bunch of error handling which could be avoided by moving it closer to the front of the pipeline
3. doesn't structurally catch errors regarding various packet shapes.

I don't really understand why we are trying to handle ipv4 and ipv6 at the same time.  That is making this much harder than is needed (and likely slowing it down a fair bit)

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -110,114 +108,162 @@ pub(crate) fn nat_translate_icmp_inner<Buf: PacketBufferMut>(
...
 
+fn translate_inner_tcp_udp_dst(
+    transport: &mut EmbeddedTransport,
+    target_port: NatPort,
+) -> Result<(), IcmpErrorMsgError> {
+    // Assume we have TCP or UDP, with destination port always present
+    let old_port = transport
+        .destination()
+        .unwrap_or_else(|| unreachable!())
+        .into();
```
daniel-noland, line 253: crash on invalid data!?!

nat/src/stateless/mod.rs
```diff
@@ -268,6 +233,32 @@ impl StatelessNat {
...
+        }
+
+        Ok(modified)
+    }
+
+    fn destination_nat<Buf: PacketBufferMut>(
+        &self,
+        table: &PerVniTable,
+        packet: &mut Packet<Buf>,
+    ) -> Result<bool, StatelessNatError> {
```
daniel-noland, line 278: why accept `Packet` instead of `Headers`?  Aren't we just making our lives hard here?

daniel-noland (dismissed):
I'm really confused by the whole approach taken here.  Maybe there is some reason for this approach I can't see, but I think that

1. completely refusing to use the qualified types from `net` and
2. dealing with ipv4 and ipv6 at the same time and
3. taking `Packet` where you could take `Headers`

has made life hard here for no real reason.

I would also really like to see fuzz tests to catch edge cases in this type of logic.

Granted, many of these problems (if problems they are, I may have missed something fundamental) predate this PR, but some of it does not.

I'm concerned about the whole direction here.  Especially as regards the global reasoning and use of `unreachable!()` when the methods are not

1. marked `unsafe` or
2. structurally defended against arbitrary crash by the type system.

`unreachable!()` isn't for code that is actually unreachable.  Not for code which is simply not reached in the current impl or tests.

That said, do let me know if I missed something major here.  I'm happy to chat about it

nat/src/stateless/mod.rs
qmonnet: This code is in the NAT pipeline stage. We're not processing just any packet here: If we reach this point, we've already identified that the packet needs to go through static NAT.

nat/src/stateless/mod.rs
qmonnet: The issue with `NonZero<u16>` is that, well, it couldn't be zero :) But the identifier may be zero in the case of ICMP.

The name "port" everywhere, and more generally the processing of TCP/UDP and ICMP via the same functions, are misleading and need to be fixed (this is tracked to some extent with #995).

nat/src/stateless/mod.rs
qmonnet: True, although we don't go that far down in the log level anyway. The error is translated into a `DoneReason` at some point, and I don't know if it makes sense to get such fine-grained info in the `DoneReason` enum. I can add them anyway, or add more log, if you think it's relevant to keep this info.

nat/src/stateless/mod.rs
qmonnet: (Same as above.)

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: The test builds an ICMP Error message with no embedded packet fragment, and validates its checksum.

To do so, we prepare a packet by building the headers, setting the checksum, then deparsing to a buffer. The checksum for ICMP is computed on the header and the data, and when we set it (`icmp.update_checksum(&[])`), we pass a pointer to the data. In this test, there is no embedded packet fragment, so the data is null (`&[]`).

When we deparse the packet, we write the headers to a buffer. If we create the buffer with `TestBuffer::new()`, it's not filled with zeroes, but with some data simulating some random payload. This does not correspond to what we want, as instead, we want a buffer filled with zeroes, so that any trailing data does not mess up with checksum computation. So we build a zero-filled buffer.

This is what the comment tries to say: when computing the checksum we assumed the ICMP data would be null, to make sure this is the case (or at least that this is equivalent, from a checksum-computation point of view), we fill the buffer with zeroes. Except the comment has a typo, it should be “so that the ICMP _checksum_ ...”. I'll fix it, I can expand the comment as well if you want. Or let me know if there is a simpler way to produce a buffer just the size of the headers with no trailing data.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: True. This PR was created before the new packet builder landed, but at some point it would be interesting to make a pass through the test and harmonise packet creation, moving towards the new builder.

For fuzzing, I'm struggling with it, in particular for finding ways to validate the behaviour without re-implementing the same logics as the code under tests in the fuzzing tests. It applies very well to parsing and de-parsing, for this test I'm not sure how to efficiently use it. We should talk about it.

net/src/packet/utils.rs
qmonnet: Nor is it ever called on packets with VLAN headers, but I see what you mean. I liked having it as a method for `Packet` but I don't see the point in accounting for VLAN headers if we never call the method on VLAN packets. I'll move it out of the `Packet` implementation and back into the `nat` crate, passing a pointer to a `Packet` as an argument.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: No, and we don't process them any differently than other IPs at this time.

This is tracked in #1280.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: It would move the same error handling (and branches) to the caller, possibly at multiple calling locations, and I'm not sure what the advantage of that would be ?

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: The error message is `"failure to get IP header"`, which I considered acceptable for both cases. I can add a distinct `NoIpHeader` error enum variant specific to ICMP Error messages inner packets if you want.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: Not sure I follow all of this. We need to chat.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: We need to have a discussion about this.

nat/src/stateless/mod.rs
qmonnet: I don't know. Are we? Without the `Packet` we need to pass down all the different header layers, plus the packet metadata, plus the attached flow information, to all subfunctions that need it, I think? And it doesn't sound easier, so I'm not sure I follow what you're suggesting. Let's chat?

nat/src/stateless/mod.rs
qmonnet: I added distinct enum variants and a log print for the resulting error in a new commit.

net/src/packet/utils.rs
qmonnet: I moved that function back to the `nat` crate.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: I expanded the comment.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: I added additional variants to this error enum in another commit.

daniel-noland (approved):
approved under the theory that we discuss and address the patterns in this code tomorrow.

We don't need to change instantly, but I would like this pattern to change

nat/src/stateless/mod.rs
```diff
@@ -167,30 +154,19 @@ impl StatelessNat {
...
+                packet
+                    .set_destination_port(
+                        NonZero::try_from(new_port)
+                            .map_err(|_| StatelessNatError::UnsupportedTranslation)?,
+                    )
+                    .map_err(|_| StatelessNatError::UnsupportedTranslation)?;
             }
-            Transport::Icmp6(_icmp6) => {
+            Transport::Icmp4(_) | Transport::Icmp6(_) => {
                 todo!()
```
Fredi-raspall: This will panic if we get an ICMP packet. Can't we replace it by an error?

Fredi-raspall: In both translate_dst_port and translate_src_port

Fredi-raspall (changes requested):
Looks good to me. The only concern I have relates to a todo!()

nat/src/stateless/mod.rs
qmonnet: @Fredi-raspall The `todo!()` was present before and this is not related to the PR, but OK, I added a commit to remove it.

Fredi-raspall: I know, but better if we remove it, right?

Fredi-raspall (approved):

PR #1431, opened 2026-04-02
feat(flowtable): Add FlowTable Capacity

sergeymatov:
If `FlowTable` size is reached configured capacity new flows won't be created and appropriate packet should be dropped.
Changes including NAT/PortFW error handling and appropriate packet drop.
Default value is set to 1M of flows.

### DIFF
--- a/config/src/external/mod.rs
+++ b/config/src/external/mod.rs
@@ -8,6 +8,8 @@ pub mod gwgroup;
 pub mod overlay;
 pub mod underlay;
 
+use std::num::NonZero;
+
 use crate::internal::device::DeviceConfig;
 use crate::{ConfigError, ConfigResult};
 use communities::PriorityCommunityTable;
@@ -31,6 +33,8 @@ pub struct ExternalConfig {
     pub overlay: Overlay, /* VPCs and peerings -- get highly developed in internal config */
     pub gwgroups: GwGroupTable, /* gateway group table */
     pub communities: PriorityCommunityTable, /* priority-to-community table */
+    #[builder(default)]
+    pub flow_table_capacity: Option<NonZero<usize>>, /* optional hard cap of flow table */
 }
 impl ExternalConfig {
     pub const BLANK_GENID: GenId = 0;
@@ -46,6 +50,7 @@ impl ExternalConfig {
             overlay: Overlay::default(),
             gwgroups: GwGroupTable::new(),
             communities: PriorityCommunityTable::new(),
+            flow_table_capacity: None,
         }
     }
 
--- a/flow-entry/src/flow_table/table.rs
+++ b/flow-entry/src/flow_table/table.rs
@@ -9,7 +9,8 @@ use std::hash::Hash;
 use std::{borrow::Borrow, time::Duration};
 use tracing::debug;
 
-use concurrency::sync::{Arc, RwLock};
+use concurrency::sync::atomic::{AtomicUsize, Ordering};
+use concurrency::sync::{Arc, RwLock, Weak};
 
 use net::flows::{FlowInfo, FlowStatus};
 
@@ -17,6 +18,8 @@ use net::flows::{FlowInfo, FlowStatus};
 pub enum FlowTableError {
     #[error("Invalid number of shards: {0}. Must be a power of two.")]
     InvalidShardCount(usize),
+    #[error("Flow table capacity exceeded")]
+    CapacityExceeded,
 }
 
 type Table = DashMap<FlowKey, Arc<FlowInfo>, RandomState>;
@@ -25,6 +28,7 @@ type Table = DashMap<FlowKey, Arc<FlowInfo>, RandomState>;
 pub struct FlowTable {
     // TODO(mvachhar) move this to a cross beam sharded lock
     pub(crate) table: Arc<RwLock<Table>>,
+    capacity: AtomicUsize,
 }
 
 impl Default for FlowTable {
@@ -40,6 +44,11 @@ fn hasher_state() -> &'static RandomState {
 }
 
 impl FlowTable {
+    /// Default capacity for the flow table: 10M.
+    ///
+    /// Use [`FlowTable::set_capacity`] to enforce a hard limit.
+    pub const DEFAULT_CAPACITY: usize = 10_000_000;
+
     #[must_use]
     pub fn new(num_shards: usize) -> Self {
         Self {
@@ -47,7 +56,16 @@ impl FlowTable {
                 hasher_state().clone(),
                 num_shards,
             ))),
+            capacity: AtomicUsize::new(Self::DEFAULT_CAPACITY),
         }
+    }
+
+    /// Set the hard capacity limit for the flow table.
+    ///
+    /// When the table reaches this limit, new flow insertions will fail with
+    /// [`FlowTableError::CapacityExceeded`] after one drain attempt.
+    pub fn set_capacity(&self, capacity: usize) {
+        self.capacity.store(capacity, Ordering::Relaxed);
     }
 
     /// Reshard the flow table into the given number of shards.
@@ -94,10 +112,17 @@ impl FlowTable {
     /// # Panics
     ///
     /// Panics if:
-    ///  - this thread already holds the read lock on the table orif the table lock is poisoned.
+    ///  - this thread already holds the read lock on the table or if the table lock is poisoned.
     ///  - if the `flow_info` to insert has a key different from `flow_key`
     ///
-    pub fn insert(&self, flow_key: FlowKey, mut flow_info: FlowInfo) -> Option<Arc<FlowInfo>> {
+    /// # Errors
+    ///
+    /// Returns [`FlowTableError::CapacityExceeded`] when the table has reached its hard limit.
+    pub fn insert(
+        &self,
+        flow_key: FlowKey,
+        mut flow_info: FlowInfo,
+    ) -> Result<Option<Arc<FlowInfo>>, FlowTableError> {
         // if the flow_info embeds its key already, it must match `flow_key`
         flow_info.flowkey().inspect(|key| {
             assert_eq!(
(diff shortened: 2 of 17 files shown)


flow-entry/src/flow_table/table.rs
```diff
@@ -10,6 +10,8 @@ use std::hash::Hash;
 use std::time::{Duration, Instant};
 use tracing::debug;
 
+use std::sync::atomic::{AtomicUsize, Ordering};
```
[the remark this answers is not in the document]
sergeymatov: done

nat/src/portfw/nf.rs
```diff
@@ -129,11 +129,33 @@ impl PortForwarder {
...
+            Ok(None) => {}
+            Err(FlowTableError::CapacityExceeded) => {
+                warn!(
+                    "Flow table capacity exceeded on reverse insert; dropping port-forwarded packet"
+                );
+                packet.done(DoneReason::NatOutOfResources);
+            }
+            Err(e) => {
+                warn!("Unexpected flow table error on reverse insert: {e}; dropping packet");
+                packet.done(DoneReason::InternalFailure);
```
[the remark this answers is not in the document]
sergeymatov: fixed

config/src/external/mod.rs
```diff
@@ -31,6 +31,8 @@ pub struct ExternalConfig {
     pub overlay: Overlay, /* VPCs and peerings -- get highly developed in internal config */
     pub gwgroups: GwGroupTable, /* gateway group table */
     pub communities: PriorityCommunityTable, /* priority-to-community table */
+    #[builder(default)]
+    pub flow_table_capacity: Option<usize>, /* optional hard cap of flow table */
```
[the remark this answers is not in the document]
sergeymatov: We can ignore this cause it's nitpicking

flow-entry/src/flow_table/table.rs
```diff
@@ -115,10 +127,17 @@ impl FlowTable {
...
     ///
-    pub fn insert(&self, flow_key: FlowKey, mut flow_info: FlowInfo) -> Option<Arc<FlowInfo>> {
+    /// # Errors
+    ///
+    /// Returns [`FlowTableError::CapacityExceeded`] when the table has reached its hard limit.
+    pub fn insert(
+        &self,
+        flow_key: FlowKey,
+        mut flow_info: FlowInfo,
+    ) -> Result<Option<Arc<FlowInfo>>, FlowTableError> {
```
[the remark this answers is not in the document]
sergeymatov: Fixed

Fredi-raspall (changes requested):

daniel-noland (commented):
I'm not done reviewing yet, but I think I spotted an issue.  Need to check.  I'm going to briefly mark this as don't-merge

flow-entry/src/flow_table/table.rs
```diff
@@ -40,16 +51,38 @@ fn hasher_state() -> &'static RandomState {
...
+    /// proactively purge all stale (Expired / Cancelled / deadline-passed) entries to
+    /// prevent unbounded memory growth.
+    pub const AGGRESSIVE_REAP_THRESHOLD: usize = 1_000_000;
+
+    /// When the table is over `reap_threshold`, run the O(n) drain scan at most once
+    /// every this many inserts.  Between scans the table may grow by up to this many
+    /// entries above the threshold, which is an acceptable bounded overshoot.
+    pub const DRAIN_EVERY_N_INSERTS: usize = 1_000;
+
+    pub const DEFAULT_CAPACITY: usize = 1_000_000;
```
mvachhar: This is way to small.  Our minimum system requirements are 32 GiB of RAM for < 50 Gb/s and 12 * 16 GiB for >= 50 Gb/s, so we should make this 16Mi (16*1024*1024), since even 1 kiB per flow would mean 16 GiB of RAM which leaves plenty for other things, and 1 kiB per flow is a lot.

flow-entry/src/flow_table/table.rs
```diff
@@ -151,97 +194,185 @@ impl FlowTable {
...
-                    () = tokio::time::sleep_until(tokio::time::Instant::from_std(deadline)) => {
-                        let status = flow_info.status();
-                        if status != FlowStatus::Active {
-                            debug!("Flow-timer[EXPIRED]: Flow {flow_key} is in status {status}");
+    fn insert_common(
+        &self,
+        flow_key: FlowKey,
+        val: &Arc<FlowInfo>,
+    ) -> Result<Option<Arc<FlowInfo>>, FlowTableError> {
+        // Try the insert; if it creates a new entry that pushes the count over the
```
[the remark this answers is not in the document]
sergeymatov: I've implemented approach explained here. Hope it's reflecting this idea

sergeymatov:
@daniel-noland Thank you for the review. I will mark several of your comments as unrelated since I've applied new logic to the flow insertion if corresponding flow is there and there was a rebase on top of main

flow-entry/src/flow_table/table.rs
```diff
@@ -40,16 +44,31 @@ fn hasher_state() -> &'static RandomState {
 }
 
 impl FlowTable {
+    /// Default capacity for the flow table: unlimited.
+    ///
+    /// When no explicit limit is configured the table can grow without bound.
```
Fredi-raspall: I am not sure this comment is aligned with the fact that there is a default value?

flow-entry/src/flow_table/table.rs
```diff
@@ -224,10 +261,42 @@ impl FlowTable {
...
+
+        // Enforce capacity for genuinely new entries only; replacements of an existing
+        // key are always allowed so that established flows can be refreshed when full.
+        //
+        // Additionally, a flow that completes a related pair (e.g. the reverse direction
+        // of a NAT session) is always admitted even if the table is at capacity.
+        // Leaving one side of a pair without its reverse is worse than briefly exceeding
+        // the limit by one entry.  Concurrent inserts from other workers may cause the
+        // count to drift slightly above the limit; that is acceptable.
+        if !table.contains_key(&flow_key) && table.len() >= capacity {
```
Fredi-raspall: I have two comments on this logic.
1) the comment of the code seems to indicate that we refresh flows by re-inserting them. That could be done and there actually exists a re-insert. However, we don't do that and wonder if we ever should since flow infos get attached extra info (e.g. NAT) that would be lost this way unless re-created too. In general, I'd say that we refresh only on activity (from packets references) and just by sliding the timeout forward. If we're sure we'll always go that way, the contains() call (which is a lookup) may be avoided.

2) later, you check if a flow has a related flow and that related can be upgraded and is in the table. That's another lookup. I wonder if we could avoid that second lookup by inspecting the status of the flow. Right now, IIRC, a flow is active if it lives in the table. The flow info constructor sets it to Active. If we changed that behavior so that a flow would be active iff it was in the flow table, then the table.get(k) below could be replaced by a check of the status. Ofc, checking the atomic may not be as reliable as the direct lookup, but it may be 99.999% of the times ? Food for thought....

nat/src/stateful/mod.rs
```diff
@@ -222,8 +224,24 @@ impl StatefulNat {
...
-        self.sessions.insert_from_arc(*flow_key, &forward);
-        self.sessions.insert_from_arc(reverse_key, &reverse);
+        self.sessions
+            .insert_from_arc(*flow_key, &forward)
+            .map_err(|e| match e {
+                FlowTableError::CapacityExceeded => StatefulNatError::CapacityExceeded,
+                FlowTableError::InvalidShardCount(_) => unreachable!(),
+            })?;
+
+        // The reverse insert is expected to always succeed: capacity enforcement
```
Fredi-raspall, line 212: In light of this comment, shouldn't we add a debug_assert or similar if inserting a "reverse" flow fails?

flow-entry/src/flow_table/table.rs
sergeymatov: Answering inline:
1 - I think you are right. contains_key is here to divide "new entry" from "replaced entry" . We could replace the contains_key + table.len() check with just table.len(), treating replacements as capacity-consuming too — which is conservative but correct and eliminates the lookup. But as a reminder `table.len()` returns not only Active flows present in a table

2 - I think we first mark flow Active (insert_common) and only then add it into the DashMap. I guess we can swap those things in the code, and `rel.status() == Active` can be a good check in this case

nat/src/stateful/mod.rs
sergeymatov: Oh, I think you are right, as well as in portfw/nf.rs

flow-entry/src/flow_table/table.rs
Fredi-raspall: Answers for answer ;-)
1- I think our goal is to protect us from memory exhaustion with many flows. So, if there are "many" flows, we should stop admitting more. Therefore, whether they are active or not is irrelevant. All consume the same.
2- Yeah. I think you could even use `rel.is_active()` IIRC.

flow-entry/src/flow_table/table.rs
```diff
@@ -224,10 +260,37 @@ impl FlowTable {
...
+        // If the flow key is not already in the table and we've reached capacity,
+        // we need to check if the related flow is still active before rejecting the insertion.
+        // This allows us to keep related flows together in the table even when at capacity,
+        // while still enforcing the capacity limit for new unrelated flows.
+        if !table.contains_key(&flow_key) && table.len() >= capacity {
+            let has_related_in_table = val
+                .related
+                .as_ref()
+                .and_then(Weak::upgrade)
+                .is_some_and(|rel| rel.status() == FlowStatus::Active);
```
Fredi-raspall: Here you could use `FlowInfo::is_valid()` (which should actually be renamed to is_active()).
Most importantly, I believe you should change `FlowInfo::new()` so that when FlowInfos are constructed, their state is NOT Active. Else, this check is not helpful since, by default, all flowInfos are created as Active. You could define a new state or actually use Detached (not an ideal name), but which means a flow that is not present in the table.

Fredi-raspall (changes requested):
LGTM. Just one minor thing that should be corrected IMO.

flow-entry/src/flow_table/table.rs
mvachhar: Can we stack a commit before this one for this PR that does the rename?

Fredi-raspall (approved):
LGTM Sergey!

PR #1441, opened 2026-04-08
Pr/fredi/fix asan complaints

Fredi-raspall:
* Fix potential leak due to cyclic references in next-hops.
* This fixes a test that was failing with address sanitizer
* ~~Making this a DRAFT since I'll opportunistically clean-up more stuff.~~
* Fix thread sanitizer complaint in test.
Fixes https://github.com/githedgehog/internal/issues/355

### DIFF
--- a/left-right-tlcache/src/lib.rs
+++ b/left-right-tlcache/src/lib.rs
@@ -106,10 +106,10 @@ impl<T: Identity<K>, K: PartialEq> ReadHandleEntry<T, K> {
         if self.rhandle.was_dropped() {
             return false;
         }
-        if *key == self.identity {
-            return true;
+        if self.version != provider.get_version() {
+            return false;
         }
-        if self.version == provider.get_version() {
+        if *key == self.identity {
             return true;
         }
         let Some(identity) = provider.get_identity(key) else {
--- a/routing/src/fib/fibtable.rs
+++ b/routing/src/fib/fibtable.rs
@@ -208,7 +208,7 @@ impl ReadHandleProvider for FibTable {
         &self,
     ) -> (
         u64,
-        impl Iterator<Item = (FibKey, &ReadHandleFactory<Fib>, FibKey)>,
+        impl Iterator<Item = (Self::Key, &ReadHandleFactory<Fib>, Self::Key)>,
     ) {
         let iter = self
             .entries
--- a/routing/src/fib/fibtable.rs
+++ b/routing/src/fib/fibtable.rs
@@ -141,6 +141,8 @@ impl FibTableWriter {
     pub fn unregister_vni(&mut self, vni: Vni) {
         self.0.append(FibTableChange::UnRegisterVni(vni));
         self.0.publish();
+        self.0.append(FibTableChange::UnRegisterVni(vni));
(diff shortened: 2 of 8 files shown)


routing/src/fib/fibtable.rs
```diff
@@ -141,6 +141,8 @@ impl FibTableWriter {
     pub fn unregister_vni(&mut self, vni: Vni) {
         self.0.append(FibTableChange::UnRegisterVni(vni));
         self.0.publish();
+        self.0.append(FibTableChange::UnRegisterVni(vni));
+        self.0.publish();
```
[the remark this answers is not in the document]
Fredi-raspall: No, this is intended.

routing/src/fib/fibtable.rs
```diff
@@ -149,6 +151,8 @@ impl FibTableWriter {
             self.0.append(FibTableChange::UnRegisterVni(vni));
         }
         self.0.publish();
+        self.0.append(FibTableChange::Del(fibid));
+        self.0.publish();
```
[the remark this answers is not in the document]
Fredi-raspall: same

routing/src/fib/test.rs
```diff
@@ -16,6 +16,7 @@ mod tests {
     use crate::fib::fibtype::FibWriter;
     use crate::rib::nexthop::NhopKey;
 
+    use left_right::Absorb;
```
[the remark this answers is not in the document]
Fredi-raspall: removed

routing/src/fib/fibtable.rs
qmonnet, line 145: Not sure I understand these changes, this is to unregister the FIB in both versions of the table (left and right) so that no version keeps allocated resources in memory, is that right? Do we expect a large performance impact from the double publication? (I guess not?)

left-right-tlcache/src/lib.rs
```diff
@@ -106,10 +106,10 @@ impl<T: Identity<K>, K: PartialEq> ReadHandleEntry<T, K> {
         if self.rhandle.was_dropped() {
             return false;
         }
-        if *key == self.identity {
-            return true;
+        if self.version != provider.get_version() {
+            return false;
         }
-        if self.version == provider.get_version() {
+        if *key == self.identity {
```
qmonnet, line 112: Might be worth a comment in the commit description to explain why the condition was invalid.

qmonnet (approved):
I'm not that familiar with the routing code and I don't have an in-depth understanding of all the changes in the PR; but I didn't see anything that seems concerning, so that's an ACK from me. Thanks!

routing/src/fib/fibtable.rs
Fredi-raspall: Yes, due to the way left-right works, the two copies may share an Arc in this case. This is to make sure that the fib does not remain. I'd say there's zero performance penalty (in data-path at least). The writer runs in the routing thread.

left-right-tlcache/src/lib.rs
Fredi-raspall: Added

PR #1451, opened 2026-04-13
remove etherparse types from tcp options handling :yawning_face: :yawning_face: 

daniel-noland:
Yet more tedious stuff needed for ACLs

### DIFF
--- a/net/src/tcp/option.rs
+++ b/net/src/tcp/option.rs
@@ -1,0 +1,365 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Individual TCP option types.
+
+use arrayvec::ArrayVec;
+use std::num::NonZero;
+
+/// A single TCP header option.
+///
+/// See [RFC 9293 section 3.1](https://datatracker.ietf.org/doc/html/rfc9293#section-3.1).
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    strum_macros::EnumCount,
+    strum_macros::Display,
+    strum_macros::IntoStaticStr,
+    strum_macros::AsRefStr,
+    strum_macros::EnumMessage,
+)]
+#[repr(u8)]
+pub enum TcpOption {
+    /// No-Operation padding (kind 1, RFC 9293).
+    Noop = 1,
+
+    /// Maximum Segment Size (kind 2, RFC 9293).
+    Mss(TcpMss) = 2,
+
+    /// Window Scale factor (kind 3, RFC 7323).
+    ///
+    /// Valid scale factors are 0..=14 per RFC 7323, but this is not currently
+    /// enforced at parse time.
+    WindowScale(u8) = 3,
+
+    /// SACK Permitted (kind 4, RFC 2018).
+    ///
+    /// Indicates the sender supports selective acknowledgement.
+    /// Only valid in SYN segments.
+    SackPermitted = 4,
+
+    /// Selective Acknowledgement blocks (kind 5, RFC 2018).
+    Sack(TcpSack) = 5,
+
+    /// Timestamps (kind 8, RFC 7323).
+    ///
+    /// First value is the sender timestamp (`TSval`), second is the echo reply
+    /// timestamp (`TSecr`).
+    Timestamp(u32, u32) = 8,
+}
+
+impl TcpOption {
+    /// Convert from an etherparse option element.
+    pub(in crate::tcp) fn from_etherparse(
+        element: &etherparse::TcpOptionElement,
+    ) -> Result<Self, TcpOptionParseError> {
+        use etherparse::TcpOptionElement;
+        match *element {
+            TcpOptionElement::Noop => Ok(TcpOption::Noop),
+            TcpOptionElement::MaximumSegmentSize(v) => {
+                let mss = TcpMss::new(v).map_err(|_| TcpOptionParseError::ZeroMss)?;
+                Ok(TcpOption::Mss(mss))
+            }
+            TcpOptionElement::WindowScale(scale) => Ok(TcpOption::WindowScale(scale)),
+            TcpOptionElement::SelectiveAcknowledgementPermitted => Ok(TcpOption::SackPermitted),
+            TcpOptionElement::SelectiveAcknowledgement(first, ref rest) => {
+                let mut buf = [TcpSackBlock::default(); 4];
+                buf[0] = TcpSackBlock::new(first.0, first.1);
+                let mut count = 1;
+                for (left, right) in rest.iter().flatten() {
+                    buf[count] = TcpSackBlock::new(*left, *right);
+                    count += 1;
+                }
+                // count is 1..=4 (1 mandatory + up to 3 optional), so this cannot fail.
+                let sack = TcpSack::new(&buf[..count]).unwrap_or_else(|_| unreachable!());
+                Ok(TcpOption::Sack(sack))
+            }
+            TcpOptionElement::Timestamp(val, echo) => Ok(TcpOption::Timestamp(val, echo)),
+        }
+    }
+
+    /// Convert to an etherparse option element.
+    pub(in crate::tcp) fn to_etherparse(&self) -> etherparse::TcpOptionElement {
+        use etherparse::TcpOptionElement;
+        match self {
+            TcpOption::Noop => TcpOptionElement::Noop,
+            TcpOption::Mss(mss) => TcpOptionElement::MaximumSegmentSize(mss.get()),
+            TcpOption::WindowScale(scale) => TcpOptionElement::WindowScale(*scale),
+            TcpOption::SackPermitted => TcpOptionElement::SelectiveAcknowledgementPermitted,
+            TcpOption::Sack(sack) => {
+                let b = sack.blocks();
+                let first = (b[0].left_edge(), b[0].right_edge());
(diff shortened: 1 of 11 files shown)


net/src/tcp/option.rs
```diff
@@ -0,0 +1,324 @@
...
+
+impl TcpSackBlock {
+    /// Create a new SACK block.
+    #[must_use]
+    pub const fn new(left_edge: u32, right_edge: u32) -> Self {
+        Self {
+            left_edge,
+            right_edge,
+        }
+    }
```
daniel-noland: do we need to assert that the left_edge is <=  right_edge?

daniel-noland: this does feel like something which should be "correct by construction" or rejected as nonsense

net/src/tcp/option.rs
```diff
@@ -0,0 +1,324 @@
...
+        self.right_edge
+    }
+}
+
+/// Selective Acknowledgement data containing 1 to 4 SACK blocks (RFC 2018).
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub struct TcpSack {
+    blocks: [TcpSackBlock; 4],
+    len: u8,
+}
```
daniel-noland, line 226: I don't understand the function of this len.  Shouldn't this be an ArrayVec?

daniel-noland: at minimum, the len should be `NonZero<u8>` (unless I'm totally lost)

daniel-noland (commented):
<notes for agent and self>

net/src/tcp/option.rs
daniel-noland: checked with the RFC and I hate it but this is correct

net/src/tcp/option.rs
daniel-noland: fixed by other means

net/Cargo.toml
```diff
@@ -31,6 +31,9 @@ multi_index_map = { workspace = true, default-features = false, features = ["ser
 rapidhash = { workspace = true }
 rkyv = { workspace = true, features = ["alloc", "bytecheck"]}
 serde = { workspace = true, features = ["derive", "std"] }
+static_assertions = { workspace = true }
+strum = { workspace = true }
+strum_macros = { workspace = true }
```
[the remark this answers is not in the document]
daniel-noland: this is kinda reasonable and kinda silly.  I'll think about it

net/src/tcp/options.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// Maximum number of bytes that can be stored in TCP options.
+    pub const MAX_LEN: usize = 40;
+
+    /// Returns the options as a byte slice, or `None` if empty.
+    #[must_use]
+    pub fn as_bytes(&self) -> Option<&[u8]> {
+        if self.0.is_empty() {
+            return None;
+        }
+        Some(self.0.as_slice())
```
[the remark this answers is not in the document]
daniel-noland: Is that valid here?  Don't the options have a hard and fast requirement to be multiples of 4 bytes?

net/src/tcp/option.rs
```diff
@@ -0,0 +1,433 @@
...
+            #[allow(clippy::cast_possible_truncation)] // const asserted to be safe (also repr(u8))
+            let kind = u.produce::<u8>()? % TcpOption::COUNT as u8;
+            match kind {
+                0 => Some(TcpOption::Noop),
+                1 => Some(TcpOption::Mss(u.produce()?)),
+                2 => Some(TcpOption::WindowScale(u.produce()?)),
+                3 => Some(TcpOption::SackPermitted),
+                4 => Some(TcpOption::Sack(u.produce()?)),
+                5 => Some(TcpOption::Timestamp(u.produce()?, u.produce()?)),
+                _ => unreachable!(),
```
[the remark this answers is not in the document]
daniel-noland: yeah, that's fair.

net/Cargo.toml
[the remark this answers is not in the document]
daniel-noland: yeah, I think static_assertions can be a normal dep because it really "doesn't exist" at runtime anyway

net/src/tcp/option.rs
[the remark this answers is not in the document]
daniel-noland: fixed

net/src/tcp/option.rs
```diff
@@ -0,0 +1,433 @@
...
+
+impl TcpOptionParseError {
+    pub(in crate::tcp) fn from_etherparse_read(e: &etherparse::TcpOptionReadError) -> Self {
+        use etherparse::TcpOptionReadError;
+        match *e {
+            TcpOptionReadError::UnknownId(id) => TcpOptionParseError::UnknownKind(id),
+            TcpOptionReadError::UnexpectedEndOfSlice { .. } => TcpOptionParseError::UnexpectedEnd,
+            TcpOptionReadError::UnexpectedSize { .. } => TcpOptionParseError::UnexpectedSize,
+        }
+    }
```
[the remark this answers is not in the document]
daniel-noland: fixed

net/src/tcp/options.rs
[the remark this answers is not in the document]
daniel-noland: yeah this is fine now

net/src/tcp/option.rs
```diff
@@ -0,0 +1,365 @@
...
+impl TcpSackBlock {
+    /// Create a new SACK block.
+    #[must_use]
+    pub fn new(left_edge: u32, right_edge: u32) -> Self {
+        if left_edge > right_edge {
+            tracing::debug!(
+                left_edge,
+                right_edge,
+                "SACK block wraps sequence number space (left > right); \
+                 legal per modular arithmetic but suspicious"
```
qmonnet, line 200: Nit: I wouldn't break strings, especially here where it's not that long.

net/src/tcp/option.rs
```diff
@@ -0,0 +1,365 @@
...
+    /// MSS value was zero.
+    #[error("zero MSS value")]
+    ZeroMss,
+    /// Unknown TCP option kind byte.
+    #[error("unknown TCP option kind: {0}")]
+    UnknownKind(u8),
+    /// Option data was truncated.
+    #[error(
+        "unexpected end of TCP option (kind {option_id}): \
+         need {expected_len} bytes, have {actual_len}"
```
qmonnet, line 286: Same nitpick.

net/src/tcp/options.rs
```diff
@@ -0,0 +1,111 @@
...
+use crate::tcp::option::{TcpOption, TcpOptionParseError};
+
+/// TCP header options.
+///
+/// Wraps [`etherparse::TcpOptions`] and provides an iterator over typed
+/// [`TcpOption`] elements.
+///
+// TODO: implement Parse/DeParse for standalone options parsing/serialization
+// TODO: add mutation API (add/remove/modify individual options)
+// TODO: add construction from a list of TcpOption values
```
qmonnet, line 15: Are all of these items significant work? If not, any reason why not addressing them in the same PR?

qmonnet (approved):
Looks OK from my side (with minor, non-blocking comments or questions below).

One question: I'm assuming most of the RFC lookup is being done by Claude in this work? If that's the case, do you have a way to validate that it's sourcing up-to-date RFCs? I'm concerned it might occasionally use one that has been superseded without realising it, for example. Note that I haven't spotted (or thoroughly searched for) any occurrence of it, it's just a concern at the back of my mind and I was wondering if you had anything to "validate" the info, in your prompt or in your review workflow, because I'd be interested as well.

net/src/tcp/options.rs
daniel-noland: actually some of this is rather significant.  It would be a fair bit more work, especially the mutation.  It also has a fair performance question on the parse.

qmonnet: OK, thanks

daniel-noland:
> Looks OK from my side (with minor, non-blocking comments or questions below).
> 
> One question: I'm assuming most of the RFC lookup is being done by Claude in this work? If that's the case, do you have a way to validate that it's sourcing up-to-date RFCs? I'm concerned it might occasionally use one that has been superseded without realising it, for example. Note that I haven't spotted (or thoroughly searched for) any occurrence of it, it's just a concern at the back of my mind and I was wondering if you had anything to "validate" the info, in your prompt or in your review workflow, because I'd be interested as well.

That is a fair question.  I did look to check that the RFCs were relevant, but I should double check that they aren't superseded.

The most likely bit would be tcp options and icmp types since that is regularly abused.

I'll give it a once over because that is a real concern

net/src/tcp/option.rs
daniel-noland: oh, it's just following the linter to a fault.  I tend to agree, but I don't argue with linters unless I really really need to.

daniel-noland: that said, the linter might not complain here.  It just generated with a line limit

qmonnet: OK, I don't think I've ever seen the linter complained on long strings, though :eyes:

daniel-noland:
> One question: I'm assuming most of the RFC lookup is being done by Claude in this work? If that's the case, do you have a way to validate that it's sourcing up-to-date RFCs?

Actually I did find that the TCP RFC 793 -> RFC 9293.  So this is a good catch.  Fixed the commit

PR #1457, opened 2026-04-13
Dependencies update

qmonnet:
```
name     old req compatible latest  new req
====     ======= ========== ======  =======
rand     0.10.0  0.10.1     0.10.1  0.10.1
rustls   0.23.37 0.23.38    0.23.38 0.23.38
tokio    1.51.0  1.51.1     1.51.1  1.51.1
```

Manual bump. The automatic workflow [failed](https://github.com/githedgehog/dataplane/actions/runs/24325609975/job/71020200811) because it also tried to upgrade ordermap from 1.1.0 to 1.2.0; this would pull indexmap 2.14.0, which is incompatible with skim 4.0.0 (pulled via rustyline), which pins indexmap to version 2.13.0.

In this PR, we leave the ordermap bump out. There is an open, recent Dependabot PR to bump the indexmap dependency in skim, so the hope is that the conflict will naturally solve itself soon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,7 +30,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aa514b7c2d4cb8ac9d54925c91900d21fdda508877694661858744211c11c69c"
 dependencies = [
  "futures-lite",
- "libc 0.2.184",
+ "libc 0.2.185",
  "tokio",
 ]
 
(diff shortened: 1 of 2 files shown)


qmonnet:
It looks like bumping `rand` also pulls the `ordermap` update somehow, I'm not sure how exactly. Then `tokio` also has the same issue in `skim`, pinned to 1.15.0 with a pending Dependabot update ([open PRs](https://github.com/skim-rs/skim/pulls)).

I'll update the PR to just do the rustls update. I should have left it to Dependabot, after all :facepalm:

qmonnet:
Closing in favour of #1460

PR #1464, opened 2026-04-14
feat(logging): Add configurable rate limit for logging

sergeymatov:
Introduce --tracing-rate-limit BURST:REPLENISH_PER_SECOND. if omitted, logging is unchanged and no throttling is applied. Initialize tracing explicitly from main() so CLI-provided throttle settings are applied before startup logs.

### DIFF
--- a/dataplane/src/main.rs
+++ b/dataplane/src/main.rs
@@ -21,7 +21,9 @@ use pyroscope::backend::{BackendConfig, PprofConfig, pprof_backend};
 use pyroscope::pyroscope::PyroscopeAgentBuilder;
 
 use routing::{BmpServerParams, RouterParamsBuilder};
-use tracectl::{custom_target, get_trace_ctl, trace_target};
+use tracectl::{
+    TracingControl, TracingRateLimitConfig, custom_target, get_trace_ctl, trace_target,
+};
 
 use tracing::{error, info, level_filters::LevelFilter};
 
@@ -50,15 +52,24 @@ fn init_name(args: &CmdArgs) -> Result<String, String> {
         Ok(name.to_string())
     }
 }
-fn init_logging(gwname: &str) {
+fn init_logging(args: &CmdArgs, gwname: &str) {
+    TracingControl::init_with_rate_limit(args.tracing_rate_limit().map(|rate_limit| {
+        TracingRateLimitConfig {
+            burst: rate_limit.burst,
+            replenish_per_second: rate_limit.replenish_per_second,
+        }
+    }));
+
     let tctl = get_trace_ctl();
     info!(
         " ━━━━━━ Starting dataplane for gateway '{gwname}' (Version = {}) ━━━━━━",
         option_env!("VERSION").unwrap_or("dev").to_string()
     );
 
-    tctl.set_default_level(LevelFilter::DEBUG)
-        .expect("Setting default loglevel failed");
+    if args.tracing().is_none() {
+        tctl.set_default_level(LevelFilter::DEBUG)
+            .expect("Setting default loglevel failed");
+    }
 }
 
 fn process_tracing_cmds(args: &CmdArgs) {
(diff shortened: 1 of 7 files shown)


tracectl/src/control.rs
```diff
@@ -339,18 +440,28 @@ impl TracingControl {
...
+    pub fn init_with_rate_limit(rate_limit_config: Option<TracingRateLimitConfig>) {
+        let mut initialized = false;
+        let has_rate_limit_config = rate_limit_config.is_some();
+        let _ = TRACING_CTL.get_or_init(|| {
+            initialized = true;
+            TracingControl::new(rate_limit_config)
+        });
+        if has_rate_limit_config && !initialized {
+            warn!("TracingControl already initialized; ignoring provided rate-limit config");
+        }
```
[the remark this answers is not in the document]
sergeymatov: This comment can be ignored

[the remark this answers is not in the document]
mvachhar: How hard is it to write a test to cover this instead of telling copilot to go away?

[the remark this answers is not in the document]
mvachhar: (And can AI write the test with some guidance?)

dataplane/src/main.rs
```diff
@@ -124,6 +122,17 @@ fn parse_bmp_params(args: &CmdArgs) -> (Option<BmpServerParams>, Option<BmpOptio
 #[allow(clippy::too_many_lines)]
 fn main() {
     let args = CmdArgs::parse();
+    TracingControl::init_with_rate_limit(args.tracing_rate_limit().map(|rate_limit| {
```
Fredi-raspall: Why can't this live in init_logging() whose purpose was to keep main() clean of those details?

tracectl/src/control.rs
```diff
@@ -278,32 +292,120 @@ impl TargetCfgDb {
...
+            .with_level(true)
+    }
+
+    // To get rid of spaghetti filters in the layers, we split the fmt layer into two:
+    // one for unthrottled levels and one for throttled levels.
+    fn unthrottled_level_filter<S>() -> impl Filter<S> + Send + Sync
+    where
+        S: Subscriber + for<'span> LookupSpan<'span>,
+    {
+        filter_fn(|meta| matches!(*meta.level(), Level::WARN | Level::ERROR))
```
Fredi-raspall: Aren't the conditions reversed? I mean, so far we've been using debug and such (instead of error) to protect ourselves from the case where a number of errors could be triggered. With this logic, you seem to be throttling debugs and infos, but not errors and warns. Isn't this the opposite of what we're willing to achieve?

tracectl/src/control.rs
```diff
@@ -278,32 +292,120 @@ impl TargetCfgDb {
...
+                .boxed()
+        }
+    }
+
+    fn build_rate_limit_layer(
+        rate_limit_config: Option<TracingRateLimitConfig>,
+    ) -> Option<TracingRateLimitLayer> {
+        let rate_limit_config = rate_limit_config?;
+
+        let policy = match Policy::token_bucket(
```
Fredi-raspall, line 371: Why token bucket and not exponential_backoff or time_window?
In other words, what has been the reason to choose token bucket?

tracectl/src/control.rs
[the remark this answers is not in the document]
Fredi-raspall: I agree. We should have unit tests.

Fredi-raspall (changes requested):

tracectl/src/control.rs
sergeymatov: docs for the `tracing-throttle` names token_bucket as the most stable and desirable policy. It allows an initial burst to come through and then enforces a steady average rate afterward

tracectl/src/control.rs
sergeymatov: It's done for reason. Once `debug`/`info` logging is enabled, those levels are usually the highest-volume source of noise, so we rate-limit them as a safety net. `warn`/`error` are left unthrottled to preserve visibility into actual failures.

sergeymatov: Alright, we need a bit more context here, we agreed to discuss with a bigger audience

Fredi-raspall (approved):
LGTM!

PR #1466, opened 2026-04-15
Small fixes cli

Fredi-raspall:
* use better menu for completion
* disable horrible coloring
* exit on ctrl +  D or ctrl + C


qmonnet:
Rebased on `main` to address the RUSTSEC issues that were blocking CI.

qmonnet (approved):
Looks fine by me. As discussed, My own preference would be to keep colours (ideally, with an improved colour scheme is the default one is not good) rather than no colours by default.

PR #1472, opened 2026-04-16
test(config.flow-filter,nat): Introduce VpcManifest::with_exposes()

qmonnet:
This is some preparatory commit for some more work around the `VpcManifest`s and other VPC-related objects. It's simple, straightforward, involves to functional change, so I'd like to get it out of the way while I'm working on the more consistent changes.

### DIFF
--- a/config/src/external/overlay/vpcpeering.rs
+++ b/config/src/external/overlay/vpcpeering.rs
@@ -546,6 +546,14 @@ impl VpcManifest {
             ..Default::default()
         }
     }
+
+    #[must_use]
+    pub fn with_exposes(vpc_name: &str, exposes: Vec<VpcExpose>) -> Self {
+        let mut manifest = Self::new(vpc_name);
+        manifest.add_exposes(exposes);
+        manifest
+    }
+
     #[must_use]
     pub fn has_host_prefixes(&self) -> bool {
         self.exposes.iter().any(VpcExpose::has_host_prefixes)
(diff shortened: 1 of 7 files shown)


config/src/external/overlay/vpcpeering.rs
```diff
@@ -546,6 +546,18 @@ impl VpcManifest {
...
+
+    // Should be used for tests only, but we can't restrict it to #[cfg(test)] because it's used in
+    // tests for other crates, too.
+    #[must_use]
+    pub fn with_exposes(vpc_name: &str, exposes: Vec<VpcExpose>) -> Self {
+        Self {
+            name: vpc_name.to_owned(),
+            exposes,
+            ..Default::default()
+        }
```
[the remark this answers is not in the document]
qmonnet: > `VpcManifest::with_exposes` is documented as “tests only” but [...]

Actually, it doesn't matter if the method is used outside of tests. I removed that comment entirely.

> Also, to avoid constructor logic diverging over time, implement `with_exposes` in terms of `new()` + `add_exposes()` instead of duplicating field initialization.

Ok, fixed.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -546,6 +546,14 @@ impl VpcManifest {
             ..Default::default()
         }
     }
+
+    #[must_use]
+    pub fn with_exposes(vpc_name: &str, exposes: Vec<VpcExpose>) -> Self {
```
Fredi-raspall, line 551: If this is only for testing, we should add #[cfg(test)]

Fredi-raspall (approved):
I'm not sure I understand the purpose of this, but LGTM.

config/src/external/overlay/vpcpeering.rs
qmonnet: I wish, but we can't call it in other crates (even if only for testing) if we do this. We'd need a specific feature, I don't think it's worth it.

PR #1474, opened 2026-04-17
ci: stop ignoring sanitize job failures

daniel-noland:
Remove continue-on-error from the build and sanitize matrices and the sanitize summary step so sanitize failures fail the workflow. Update the summary annotation from warning to error to match, and add sanitize to publish.needs so a red sanitize on a tag push blocks release.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -200,7 +200,6 @@ jobs:
         || github.event_name == 'workflow_dispatch'
       }}
     name: "${{matrix.nix-target}}/${{matrix.build.name}}"
-    continue-on-error: ${{ matrix.build.optional || false }}
     runs-on: lab
     needs:
       - check_changes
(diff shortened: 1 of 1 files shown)


.github/workflows/dev.yml
```diff
@@ -200,7 +200,6 @@ jobs:
         || github.event_name == 'workflow_dispatch'
       }}
     name: "${{matrix.nix-target}}/${{matrix.build.name}}"
-    continue-on-error: ${{ matrix.build.optional || false }}
```
qmonnet, line 203: This one was unused, right? I suppose `matrix.build.optional` always evaluated to `false` because it's not defined (as far as I can see)?

qmonnet (approved):

.github/workflows/dev.yml
daniel-noland: yup

PR #1475, opened 2026-04-17
Simplify stateful NAT

Fredi-raspall:
This goes on top of https://github.com/githedgehog/dataplane/pull/1414

This is an attempt to simplify the stateful nat code, which became significantly complex over time and due to the use of generics and traits.


nat/src/stateful/state.rs
```diff
@@ -1,206 +1,117 @@
...
-            Some((ip, port)) => write!(f, "{ip}:{})", port.as_u16()),
-            None => write!(f, "unchanged)"),
-        }?;
-        write!(f, "[{}s]", self.idle_timeout.as_secs())
+        write!(
+            f,
+            " {} ip: {} port|Id: {} timeout: {} {}",
+            self.action,
+            self.use_ip,
+            self.use_port.as_u16(),
```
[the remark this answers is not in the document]
Fredi-raspall: Fixed!

nat/src/stateful/apalloc/alloc.rs
```diff
@@ -115,7 +120,7 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
         if let Ok(port) = self.reuse_allocated_ip(allow_null) {
             return Ok(port);
         }
-
+        // here
```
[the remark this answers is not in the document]
Fredi-raspall: removed

nat/src/stateful/apalloc/mod.rs
```diff
@@ -181,74 +205,96 @@ pub struct NatAllocator {
...
-        Ok(AllocationResult {
-            src: src_mapping,
-            return_dst: reverse_dst_mapping,
-            idle_timeout: pool_src_opt.and_then(alloc::IpAllocator::idle_timeout),
-        })
+        // FIXME: In the case of ICMP, we're only interested in the IP allocated for src_mapping,
+        // not the port. However, both source and destination IP need to come with a "port" with
+        // the current architecture of the allocator, which means we also allocate a
+        // port/identifier value for the src_mapping, even though we'll never use it. (This does
+        // not apply to TCP or UDP, for which we need and use both ports).
```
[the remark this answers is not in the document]
Fredi-raspall: Removed.

qmonnet (commented):
Looks good so far

nat/src/stateful/nf.rs
```diff
@@ -0,0 +1,569 @@
...
+    InvalidPort(u16),
+    #[error("unexpected IP protocol key variant")]
+    UnexpectedKeyVariant,
+    #[error("flow table capacity exceeded")]
+    CapacityExceeded,
+    #[error("unsupported ICMP message category")]
+    IcmpUnsupportedCategory,
+    #[error("attempted to masquerade ICMP error message")]
+    IcmpError,
+    #[error("dropped the packet purposedly")]
```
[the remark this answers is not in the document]
Fredi-raspall: fixed

nat/src/stateful/allocator_writer.rs
```diff
@@ -94,74 +94,62 @@ impl StatefulNatConfig {
...
+        if curr_allocator.is_some() {
+            check_masquerading_flows(flow_table, &mut allocator);
         }
         // make new allocator visible
         debug!("Installing new stateful NAT allocator...");
-        self.allocator.store(Some(Arc::new(allocator)));
-        self.config = nat_config;
-        debug!("Updated stateful NAT allocator");
+        self.0.store(Some(Arc::new(allocator)));
+        debug!("NAT allocator is installed");
```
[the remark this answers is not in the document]
Fredi-raspall: That removal is intended. Those flows should be checked anyway

nat/src/stateful/nf.rs
```diff
@@ -0,0 +1,569 @@
...
+            packet.done(DoneReason::NotIp);
+            return;
+        }
+
+        // TODO: Check whether the packet is fragmented
+        // TODO: Check whether we need protocol-aware processing
+
+        if let Err(error) = self.masquerade_packet(packet) {
+            packet.done(translate_error(&error));
+            error!("Error masquerading packet: {error}");
```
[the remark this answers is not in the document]
Fredi-raspall: I will ignore this recommendation since we have log rate limiting.

nat/src/stateful/allocator_writer.rs
```diff
@@ -16,13 +16,13 @@ use crate::stateful::flows::check_masquerading_flows;
...
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub(crate) struct StatefulNatPeering {
     pub(crate) src_vpcd: VpcDiscriminant,
     pub(crate) dst_vpcd: VpcDiscriminant,
     pub(crate) peering: Peering,
 }
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
```
qmonnet: You have two commits entitled `feat(stateful-nat): embed config in allocator`, with identical description, the second one containing only these chunks. You probably want to squash them.

nat/src/stateful/mod.rs
```diff
@@ -44,6 +45,8 @@ trace_target!("stateful-nat", LevelFilter::INFO, &["nat", "pipeline"]);
 
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 enum StatefulNatError {
+    #[error("Unexpected failure")]
+    Bug,
```
qmonnet: We should probably pass some more context as parameter, a bit like `ConfigError::InternalFailure(String)`.

nat/src/stateful/mod.rs
```diff
@@ -68,6 +71,8 @@ enum StatefulNatError {
     IcmpUnsupportedCategory,
     #[error("attempted to masquerade ICMP error message")]
     IcmpError,
+    #[error("dropped the packet purposedly")]
+    IntendedDrop,
```
qmonnet: Here as well, it's “purposely” but we don't state the intended purpose, sounds frustrating when reading the logs. Provide more context?

nat/src/stateful/allocator_writer.rs
```diff
@@ -141,16 +141,9 @@ impl NatAllocatorWriter {
...
-        // New flows requiring masquerading won't get any IP/port. That's fine, their sources may retry.
-        let old_allocator = self.0.swap(None);
-        debug!("Disabled stateful NAT allocator");
-
         // build a new allocator. The allocator is not yet visible in data path
         let mut allocator = NatAllocator::new(nat_config);
-        if old_allocator.is_some() {
+        if curr_allocator.is_some() {
             check_masquerading_flows(flow_table, &mut allocator);
         }
```
qmonnet: I don't understand this change. If a packet is legit for config C and forbidden for new config C', and it arrives during the execution of `check_masquerading_flows()`, it may lead to the allocation of resources from the old allocator and insert it to the flow table, which is not what we want? It would not matter as much as before, because it would use the `genid` associated with old config C, so follow-up packets wouldn't bypass in flow-filter; but it would create leftover, non-invalidated entries in the flow table?

Or do you mean that these leftover entries are an acceptable trade-off to avoid dropping packets that are valid for both C and C' while we're going through `check_masquerading_flows()`? Still sounds risky to me to keep old, non-invalidated entries in the flow table, I could easily imagine this turning into a bug. Or am I missing something else?

nat/src/stateful/mod.rs
```diff
@@ -7,584 +7,21 @@ pub mod apalloc;
...
-        })
-    }
-
-    fn set_data(&mut self, data: Arc<PipelineData>) {
-        self.pipeline_data = data;
-    }
-}
-
 #[cfg(test)]
 mod tests {
```
qmonnet, line 26: These tests should follow the related code.

By the way, I'm not sure I see the point of moving the NF to a separate file and keep a nearly-empty `mod.rs`, that's one more file for no clear benefit. I'd prefer to wait until a potential reorganisation of the `nat` crate around a single NF and NAT modes as submodules/functions.

qmonnet (changes requested):

nat/src/stateful/mod.rs
Fredi-raspall: Hum... I thought I had fixed that typo...

nat/src/stateful/allocator_writer.rs
Fredi-raspall: I am not sure I understand your question either. Should be talk ?

Fredi-raspall: Ok. I think, the potential race has been fixed.

nat/src/stateful/allocator_writer.rs
Fredi-raspall: fixed

nat/src/stateful/mod.rs
Fredi-raspall: I've added a &'static str

nat/src/stateful/mod.rs
Fredi-raspall: Added a reason for the drop.

nat/src/stateful/nf.rs
```diff
@@ -0,0 +1,569 @@
...
+        ) => DoneReason::NatFailure,
+
+        StatefulNatError::InvalidIpVersion
+        | StatefulNatError::AllocationFailure(AllocatorError::InternalIssue(_)) => {
+            DoneReason::InternalFailure
+        }
+
+        StatefulNatError::AllocationFailure(AllocatorError::Denied)
+        | StatefulNatError::Bug(_)
+        | StatefulNatError::IntendedDrop => DoneReason::Filtered,
```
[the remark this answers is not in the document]
Fredi-raspall: fixed

nat/src/stateful/nf.rs
```diff
@@ -445,7 +445,7 @@ impl StatefulNat {
             None => {
                 // allocator got removed. Get rid of the flows and drop the packet.
                 installed.invalidate_pair();
-                Err(StatefulNatError::IntendedDrop)
+                Err(StatefulNatError::IntendedDrop("allocator  was removed"))
```
qmonnet: Nit: double space

```suggestion
                Err(StatefulNatError::IntendedDrop("allocator was removed"))
```

qmonnet (approved):
Thanks, I think the guard around the flow check + allocator swap is the fix we need to address the potential race, indeed. Thanks for this work!

nat/src/stateful/nf.rs
Fredi-raspall: fixed

PR #1490, opened 2026-04-22
NAT: Reject TCP/UDP port 0 at validation

sergeymatov:
Don't allow port `0` within `VpcExpose` `ips` ranges

### DIFF
--- a/config/src/external/overlay/vpcpeering.rs
+++ b/config/src/external/overlay/vpcpeering.rs
@@ -414,6 +414,21 @@ impl VpcExpose {
             self.as_range_or_empty(),
             self.not_as_or_empty(),
         ];
+
+        // Port 0 is not allowed in the exposed ranges. We do not check the excluded ranges here,
+        // as they are only used to remove prefixes/ports from the effective configuration.
+        for prefixes in [&self.ips, self.as_range_or_empty()] {
+            for prefix_with_ports in prefixes {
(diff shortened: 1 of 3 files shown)


qmonnet:
Partly addresses #1280

qmonnet (approved):
Looks good, thank you for this!

As we discussed on Slack, in theory we should accept port 0 in `ips`/`as_range` _iff_ it's also in a corresponding exclusion prefix+ports (making it absent of the final config, once exclusion prefixes have been applied); but this will be trivially addressed after the work I'm doing to get rid of exclusion prefixes early, so this is left out of this PR.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -414,6 +414,21 @@ impl VpcExpose {
             self.as_range_or_empty(),
             self.not_as_or_empty(),
         ];
+
+        // Port 0 is not allowed in the exposed ranges. We do not check the excluded ranges here,
+        // as they are only used to remove prefixes/ports from the effective configuration.
+        for prefixes in [&self.ips, self.not_as_or_empty()] {
```
qmonnet: Oh good call from Copilot, this should be `self.as_range_or_empty()`, not the exclusion prefixes

qmonnet (changes requested):

qmonnet (approved):

PR #1491, opened 2026-04-22
chore(nat): Improve AllocatorError variants and messages

qmonnet:
`AllocatorError::NoFreePort` is not about no port block being available, but not ports being available in a block. The message is confusing: let's fix it. It's been so confusing that we used the error at a location where the error is not related (passing it an incorrect argument): let's add a new variant for that location, to make it easier to troubleshoot errors.


nat/src/stateful/allocation.rs
```diff
@@ -16,10 +16,12 @@ pub enum AllocatorError {
     NoFreeIp,
     #[error("failed to allocate port block")]
     NoPortBlock,
-    #[error("no free port block available (base: {0})")]
+    #[error("no free port available in port block (block base index: {0})")]
     NoFreePort(u16),
     #[error("failed to allocate port: {0}")]
     PortAllocationFailed(NatPortError),
+    #[error("failed to reserve port: {0}")]
+    PortReservationFailed(u16),
```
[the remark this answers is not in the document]
qmonnet: Yep, forgot to save the change and `git add` it before creating my commit. Fixed now.

nat/src/stateful/allocation.rs
```diff
@@ -14,10 +14,12 @@ pub enum AllocatorError {
     NoFreeIp,
     #[error("failed to allocate port block")]
     NoPortBlock,
-    #[error("no free port block available (base: {0})")]
+    #[error("no free port available in port block (block base index: {0})")]
```
mvachhar, line 17: We should also degrade this case to a debug message not an error message since it is expected to happen frequently.  The thing to print as an error is when we have no ports left period.

mvachhar (changes requested):

nat/src/stateful/allocation.rs
qmonnet: @mvachhar I'm not sure I follow. I can ignore the error we meet and just log a debug message instead of returning `NoFreePort`, but I'm not sure this is what we want?

I'm not sure this is supposed to happen often, by the way. If the port block has no port available, the thread is supposed to pick a different port block. I don't fully understand why we've observed repeated messages related to this error recently, but it's not something I expected. Part of the motivation for this change is to pin down where the message comes from exactly (after the change from the PR, this error is only returned from a single location, vs. two different locations before).

Fredi-raspall (approved):

nat/src/stateful/allocation.rs
mvachhar: Oh, I see, I thought this was expected and then we move to the next block.  Ok, this is fine for now but we should understand why we see this at all then.

mvachhar (approved):

PR #1493, opened 2026-04-23
Remove some "allow(unused)" in nat

qmonnet:
Remove some unnecessary `#[allow(unused)]` and `#[allow(dead_code)]` in the `nat` crate.

Follow-up to #1414.

### DIFF
--- a/nat/src/portfw/portfwtable/portrange.rs
+++ b/nat/src/portfw/portfwtable/portrange.rs
@@ -3,8 +3,6 @@
 
 //! Object to represent port ranges for port-forwarding and methods
 
-#![allow(unused)] // Temporary
-
 use super::PortFwTableError;
 use std::fmt::Display;
 use std::num::NonZero;
--- a/nat/src/stateful/apalloc/alloc.rs
+++ b/nat/src/stateful/apalloc/alloc.rs
@@ -9,10 +9,6 @@
 //!
 //! See also the architecture diagram at the top of mod.rs.
 
-// We plan to re-use unused functions soon as part of the work to conserve valid flows across
-// configuration changes.
-#![allow(dead_code)]
-
 use super::{NatIpWithBitmap, port_alloc};
 use crate::port::NatPort;
 use crate::ranges::IpRange;
--- a/nat/src/stateful/apalloc/port_alloc.rs
+++ b/nat/src/stateful/apalloc/port_alloc.rs
@@ -8,10 +8,6 @@
 //!
 //! See also the architecture diagram at the top of mod.rs.
 
-// We plan to re-use unused functions soon as part of the work to conserve valid flows across
-// configuration changes.
-#![allow(dead_code)]
-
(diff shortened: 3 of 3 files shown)


nat/src/portfw/portfwtable/access.rs
```diff
@@ -8,9 +8,6 @@ use super::PortFwTableError;
 use super::objects::{PortFwEntry, PortFwTable};
 use left_right::{Absorb, ReadGuard, ReadHandle, ReadHandleFactory, WriteHandle};
 
-#[allow(unused)]
```
Fredi-raspall, line 11: this is intended

nat/src/portfw/flow_state.rs
```diff
@@ -21,8 +21,7 @@ use flow_entry::flow_table::FlowInfo;
 use crate::portfw::PortFwEntry;
 use crate::portfw::protocol::{AtomicPortFwFlowStatus, PortFwFlowStatus, next_flow_status};
 
-#[allow(unused)]
```
Fredi-raspall, line 24: same here

nat/src/portfw/nf.rs
```diff
@@ -24,7 +24,6 @@ use crate::portfw::flow_state::setup_forward_flow;
 use crate::portfw::flow_state::setup_reverse_flow;
 use crate::portfw::packet::{dnat_packet, nat_packet};
 
-#[allow(unused)]
```
Fredi-raspall, line 27: and here too

nat/src/stateful/mod.rs
```diff
@@ -33,7 +33,6 @@ use std::fmt::Debug;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::time::{Duration, Instant};
 
-#[allow(unused)]
```
Fredi-raspall, line 36: and that too

nat/src/portfw/portfwtable/access.rs
qmonnet: @Fredi-raspall I can remove these from the PR, but can you please explain _why_ it's intended? What do these imports bring if they're unused, and why do we want to allow them to be unused if they're not?

Fredi-raspall: Efficiency, so that each time we decide to change the severity of a trace we don't have to change the imports.
Importing those macros has zero cost in runtime, binary size and since they are simple macros they should minimally increment build time.

qmonnet: > Efficiency, so that each time we decide to change the severity of a trace we don't have to change the imports.

OK, thank you.

I wonder if we can wrap this import (with the `allow(unused)` into some dedicated macro, `import_tracing_macro!()` or something like that, it would be nicer. Not very important, though.

I adjusted the commit, please take another look :pray:

Fredi-raspall (approved):

PR #1499, opened 2026-04-27
fix: fix user in Dockerfile

Fredi-raspall:
Counter the fact that root may not appear in /etc/passwd or the latter not be present.


daniel-noland (approved):
yup

PR #1503, opened 2026-04-29
chore(nat): Preserve 0..1023 ports from allocation in stateful NAT

sergeymatov:
Port 0 is illegal, ports 1..1023 are IANA system well-known ports. For UDP and TCP port range allocation we will exclude this port range from `PortRange` allocation.

### DIFF
--- a/nat/src/stateful/apalloc/port_alloc.rs
+++ b/nat/src/stateful/apalloc/port_alloc.rs
@@ -86,10 +86,22 @@ pub(crate) struct PortAllocator<I: NatIpWithBitmap> {
     thread_blocks: ThreadPortMap,
     allocated_blocks: AllocatedPortBlockMap<I>,
     reserved_port_range: Option<PortRange>,
+    exclude_wellknown_ports: bool,
 }
 
+/// Ports 0..=1023 cover the IANA system/well-known range and should not be
+/// allocated by masquerade NAT for TCP or UDP.
+const IANA_WELLKNOWN_PORT_LIMIT: u16 = 1024;
+
+/// Number of 256-port blocks covering the IANA well-known port range (0-1023).
+const IANA_WELLKNOWN_BLOCKS: u16 = IANA_WELLKNOWN_PORT_LIMIT / 256;
+
 impl<I: NatIpWithBitmap> PortAllocator<I> {
-    pub(crate) fn new(reserved_port_range: Option<PortRange>, randomize: bool) -> Self {
+    pub(crate) fn new(
+        reserved_port_range: Option<PortRange>,
+        randomize: bool,
+        exclude_wellknown_ports: bool,
+    ) -> Self {
         let mut base_ports = (0..=255).collect::<Vec<_>>();
 
         // Shuffle the list of port blocks for the port allocator. This way, we can pick blocks in a
@@ -99,21 +111,39 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
         if randomize {
             Self::shuffle_slice(&mut base_ports);
         }
-        let blocks = std::array::from_fn(|i| AllocatorPortBlock::new(base_ports[i]));
-
+        let blocks = std::array::from_fn(|i| {
+            let block = AllocatorPortBlock::new(base_ports[i]);
+            // Pre-mark IANA well-known port blocks (0-1023) as permanently non-free so they are
+            // never handed out by masquerade NAT for TCP or UDP.
+            if exclude_wellknown_ports && block.to_port_number() < IANA_WELLKNOWN_PORT_LIMIT {
+                block
+                    .free
+                    .store(false, concurrency::sync::atomic::Ordering::Relaxed);
+            }
+            block
+        });
+        let usable_blocks = if exclude_wellknown_ports {
+            256 - IANA_WELLKNOWN_BLOCKS
+        } else {
+            256
+        };
         Self {
             blocks,
-            usable_blocks: AtomicU16::new(256),
+            usable_blocks: AtomicU16::new(usable_blocks),
             current_alloc_index: AtomicUsize::new(0),
             thread_blocks: ThreadPortMap::new(),
             allocated_blocks: AllocatedPortBlockMap::new(),
             reserved_port_range,
+            exclude_wellknown_ports,
         }
     }
 
     #[cfg(test)]
-    pub(crate) fn new_no_randomness(reserved_port_range: Option<PortRange>) -> Self {
-        Self::new(reserved_port_range, false)
+    pub(crate) fn new_no_randomness(
+        reserved_port_range: Option<PortRange>,
+        exclude_wellknown_ports: bool,
+    ) -> Self {
+        Self::new(reserved_port_range, false, exclude_wellknown_ports)
     }
 
     #[concurrency_mode(std)]
@@ -345,6 +375,16 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
         ip: Arc<AllocatedIp<I>>,
         port: NatPort,
     ) -> Result<AllocatedPort<I>, AllocatorError> {
+        // Reject explicit reservations into the IANA system/well-known range up front so callers
+        // get a policy-oriented error rather than a misleading resource-exhaustion error from the
+        // pre-excluded low-port blocks.
+        if self.exclude_wellknown_ports && port.as_u16() < IANA_WELLKNOWN_PORT_LIMIT {
+            debug!(
+                "Explicit reservation for well-known port {} denied by allocator policy",
+                port.as_u16()
+            );
+            return Err(AllocatorError::Denied);
+        }
         let block = self.find_block_for_port(ip, port)?;
         block.reserve_port_from_block(port)
     }
@@ -1119,7 +1159,7 @@ mod tests {
 
     #[test]
     fn pick_available_block_no_reserved_range() {
-        let allocator = PortAllocator::<Ipv4Addr>::new_no_randomness(None);
+        let allocator = PortAllocator::<Ipv4Addr>::new_no_randomness(None, false);
         let (index, base_port) = allocator.pick_available_block().unwrap();
         assert_eq!(index, 0);
         assert_eq!(base_port, 0);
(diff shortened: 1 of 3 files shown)


nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -345,6 +384,13 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
         ip: Arc<AllocatedIp<I>>,
         port: NatPort,
     ) -> Result<AllocatedPort<I>, AllocatorError> {
+        // Reject explicit reservations into the IANA system/well-known range up front so callers
+        // get a policy error rather than an internal "block not free" error from the pre-excluded
+        // low-port blocks.
+        if self.exclude_wellknown_ports && port.as_u16() < IANA_WELLKNOWN_PORT_LIMIT {
+            return Err(AllocatorError::NoFreePort(port.as_u16()));
```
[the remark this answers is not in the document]
sergeymatov: error type fixed

nat/src/stateful/apalloc/test_alloc.rs
```diff
@@ -447,4 +447,51 @@ mod tests_shuttle {
             assert!(in_use.front().unwrap().upgrade().is_none()); // Weak references in list no longer resolve
         });
     }
+
+    // TCP and UDP masquerade NAT must never hand out IANA well-known ports (0-1023).
+    // Allocate enough ports to exhaust the first usable block (256 ports) plus some of the
+    // second, verifying that every allocated port is >= 1024.
+    #[test]
+    fn tcp_allocations_never_use_wellknown_ports() {
+        let allocator = build_allocator();
```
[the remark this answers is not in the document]
sergeymatov: My bad, I thought I am adding std test but picked a wrong place to put it. I copied this into the default concurrency mode code section while fixed and added shuttle-related

nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -345,6 +384,13 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
...
         port: NatPort,
     ) -> Result<AllocatedPort<I>, AllocatorError> {
+        // Reject explicit reservations into the IANA system/well-known range up front so callers
+        // get a policy error rather than an internal "block not free" error from the pre-excluded
+        // low-port blocks.
+        if self.exclude_wellknown_ports && port.as_u16() < IANA_WELLKNOWN_PORT_LIMIT {
+            return Err(AllocatorError::Denied);
+        }
+
         let block = self.find_block_for_port(ip, port)?;
```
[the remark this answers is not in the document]
sergeymatov: I am confused. If the port was allocated and it's < 1024 - this flow will be invalidated and it's desired behavior. I think copilot is wrong here because it tries to save the wrongly allocated flow rather then kill it  to fit the goal of the PR

nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -86,10 +86,22 @@ pub(crate) struct PortAllocator<I: NatIpWithBitmap> {
     thread_blocks: ThreadPortMap,
     allocated_blocks: AllocatedPortBlockMap<I>,
     reserved_port_range: Option<PortRange>,
+    exclude_wellknown_ports: bool,
 }
 
+/// Ports 0..=1023 cover the IANA system/well-known range and must not be
```
qmonnet: Why “must”? Please change.

nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -159,6 +189,15 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
...
         // lookup in the list.
         //
+        // IANA system/well-known blocks are permanently excluded from TCP/UDP masquerade
+        // allocation. If a caller reaches deallocation for one of these blocks, keep it excluded
+        // instead of making it available again.
+        if self.exclude_wellknown_ports
+            && self.blocks[index].to_port_number() < IANA_WELLKNOWN_PORT_LIMIT
+        {
+            return;
+        }
```
qmonnet: When would this condition ever evaluate to `true`, given that we cannot allocate from < 1024 in the first place?

nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -345,6 +384,13 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
         ip: Arc<AllocatedIp<I>>,
         port: NatPort,
     ) -> Result<AllocatedPort<I>, AllocatorError> {
+        // Reject explicit reservations into the IANA system/well-known range up front so callers
+        // get a policy error rather than an internal "block not free" error from the pre-excluded
+        // low-port blocks.
+        if self.exclude_wellknown_ports && port.as_u16() < IANA_WELLKNOWN_PORT_LIMIT {
+            return Err(AllocatorError::Denied);
```
qmonnet, line 386: This needs a debug log message

nat/src/stateful/apalloc/setup.rs
```diff
@@ -75,23 +75,29 @@ fn build_nat_pool_generic<'a, I: NatIpWithBitmap, J: NatIpWithBitmap, F, FIter,
             .idle_timeout()
             .unwrap_or(DEFAULT_MASQUERADE_IDLE_TIMEOUT);
 
+        // TCP/UDP masquerade allocators must avoid the IANA system/well-known range
```
qmonnet: Why “must”?

nat/src/stateful/apalloc/test_alloc.rs
```diff
@@ -333,6 +333,51 @@ mod std_tests {
         assert_eq!(in_use.len(), 1); // 1 allocated, in use
     }
 
+    // TCP and UDP masquerade NAT must never hand out IANA well-known ports (0-1023).
+    // Allocate enough ports to exhaust the first usable block (256 ports) plus some of the
+    // second, verifying that every allocated port is >= 1024.
```
qmonnet: Given that you've got good coverage for `pick_available_block()` already, I'd probably drop these tests rather than doing 298 “useless” allocations?

nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -99,21 +111,39 @@ impl<I: NatIpWithBitmap> PortAllocator<I> {
...
+                    .free
+                    .store(false, concurrency::sync::atomic::Ordering::Relaxed);
+            }
+            block
+        });
+        let usable_blocks = if exclude_wellknown_ports {
+            256 - IANA_WELLKNOWN_BLOCKS
+        } else {
+            256
+        };
```
qmonnet, line 129: Why not removing the block with base index 0 from `blocks` entirely? This way you don't have to mark it as non-free but keep it in memory, or to deal with deallocation, or to hack the `usable_blocks` length count?

nat/src/stateful/apalloc/port_alloc.rs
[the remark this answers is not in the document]
qmonnet: I'm not sure that killing existing connections with ports < 1024 is desirable, the change should only apply to new connections, in my opinion.

This being said, we don't currently persist flow table entries across dataplane reboots, meaning that when users launch a version with the < 1024 allocation blocking, they'll start with a fresh (empty) flow table, and they won't be able to allocate a port < 1024 in the first place, so I believe this is not a concern.

qmonnet (changes requested):
Thanks! Please find some observations and requested changes below.

Please improve the commit description: at the moment the commit reads like there's some obvious rule in place that forbids allocating from ports < 1024. The description should mention that we've not found any explicit rule in the RFCs, but existing implementations... etc.

I remember that @Fredi-raspall also required that disabling these allocations would be configurable, but this is not present in the PR.

nat/src/stateful/apalloc/port_alloc.rs
sergeymatov: PortAllocator itself operates fixed-size arrays. I see your concern here, but to make this happen there should be a refactoring done.

nat/src/stateful/apalloc/port_alloc.rs
sergeymatov: fixed. Sorry mixing up `must` and `should`

nat/src/stateful/apalloc/port_alloc.rs
sergeymatov: This is correct concern. Fixed this along with dedicated test

nat/src/stateful/apalloc/setup.rs
sergeymatov: fixed

nat/src/stateful/apalloc/port_alloc.rs
sergeymatov: fixed

nat/src/stateful/apalloc/test_alloc.rs
sergeymatov: You are right. Removed four tests (std and shuttle)

nat/src/stateful/apalloc/setup.rs
qmonnet: not fixed 🙃

nat/src/stateful/apalloc/port_alloc.rs
```diff
@@ -499,7 +539,7 @@ impl<I: NatIpWithBitmap> AllocatedPortBlock<I> {
                     )
                 })?,
             )
-            .map_err(|()| AllocatorError::PortReservationFailed(port.as_u16()))?;
+            .map_err(|()| AllocatorError::NoFreePort(port.as_u16()))?;
```
qmonnet: I don't understand this change. This is a partial revert from recent commit ee219dc46a5208cfde5af7dfa7699b261206a641, and I don't understand why you'd switch back to `NoFreePort`. One of the motivation to have a different (and more accurate) error here was to pin down more precisely where the `NoFreePort`-related we've been observing recently are coming from. Would you mind changing back, please?

qmonnet (changes requested):
Thanks for addressing the feedback! Just two small things below.

qmonnet (approved):
Thank you!

PR #1504, opened 2026-04-29
ICMP error handling enchilada 🌮🫜🫑

### LINKED ISSUE #1500: Dataplane: fix broken ICMP error support

Fredi-raspall:
* Fix ICMP error handling, fixes: https://github.com/githedgehog/dataplane/issues/1500
* Fix ICMP error handling with masquerade, fixes https://github.com/githedgehog/internal/issues/367
* Reorganize code to so that ICMP error handling is independent of the flow-filter.
* Invalidate flows on certain ICMP errors like port/host unreachable since sources may not retry. This helps in:
    * expediting removal of flows, without needing to wait for timeouts
    * in masquerading, releasing allocated ports faster.


nat/src/icmp_handler/nf.rs
```diff
@@ -84,73 +83,41 @@ fn icmp_checksums_ok<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> bool {
...
 
-/// Build a `FlowKey` for the embedded "offending" packet within an ICMP error packet
-fn get_icmp_inner_pkt_flowkey<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> Option<FlowKey> {
-    // get the data for the embedded, offending packet
-    let Some(inner) = packet.embedded_headers() else {
-        debug!("ICMP error message does not contain inner packet");
-        return None;
+/// Tell if the given ICMP error packet is "terminating" in the sense that any flow related to it *could* be removed
+/// because the source may no longer send packets on reception of the error
+fn is_icmp_unrecoverable<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> (bool, Option<&str>) {
```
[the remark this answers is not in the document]
Fredi-raspall: rephrased

nat/src/icmp_handler/nf.rs
```diff
@@ -84,73 +83,41 @@ fn icmp_checksums_ok<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> bool {
...
+            (true, unreach.get_message())
         }
-        EmbeddedTransport::Icmp6(icmp6) => {
-            debug_assert_eq!(proto, NextHeader::ICMP6);
-            // Icmp errors are not sent for icmp errors
-            let icmp_id = icmp6.identifier()?;
-            IpProtoKey::Icmp(IcmpProtoKey::QueryMsgData(icmp_id))
+        IcmpAny::V6(icmp6) if let Icmp6Type::PacketTooBig(_) = icmp6.icmp_type() => {
+            (false, Some("Packet too big"))
         }
```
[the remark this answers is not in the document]
Fredi-raspall: fixed

net/src/headers/embedded.rs
```diff
@@ -740,11 +740,11 @@ impl<T> AbstractEmbeddedHeadersMut for T where
...
 
 pub trait TryEmbeddedHeaders {
-    fn embedded_headers(&self) -> Option<&impl AbstractEmbeddedHeaders>;
+    fn embedded_headers(&self) -> Option<&EmbeddedHeaders>;
 }
 
 pub trait TryEmbeddedHeadersMut {
-    fn embedded_headers_mut(&mut self) -> Option<&mut impl AbstractEmbeddedHeadersMut>;
+    fn embedded_headers_mut(&mut self) -> Option<&mut EmbeddedHeaders>;
 }
```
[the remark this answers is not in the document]
Fredi-raspall: that's fine.

flow-entry/src/flow_table/table.rs
```diff
@@ -301,9 +301,9 @@ impl FlowTable {
     pub fn lookup<Q>(&self, flow_key: &Q) -> Option<Arc<FlowInfo>>
     where
         FlowKey: Borrow<Q>,
-        Q: Hash + Eq + ?Sized + Debug,
+        Q: Hash + Eq + ?Sized + Debug + Display,
     {
-        debug!("lookup: Looking up flow key {:?}", flow_key);
+        debug!("lookup: Looking up flow key {flow_key}");
```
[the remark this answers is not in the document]
Fredi-raspall: that's fine

net/src/flows/flow_key.rs
```diff
@@ -677,6 +684,64 @@ impl<Buf: PacketBufferMut> TryFrom<Uni<&Packet<Buf>>> for FlowKey {
...
+///
+/// This function returns a `FlowKey` on success and `FlowKeyError` otherwise.
+///
+pub fn flowkey_embedded_in_icmp_error<Buf: PacketBufferMut>(
+    packet: &mut Packet<Buf>,
+) -> Result<FlowKey, FlowKeyError> {
+    // we currently restrict this to ICMP error packets
+    if !packet.is_icmp_error() {
+        return Err(FlowKeyError::NotIcmpError);
+    }
```
qmonnet, line 704: This is where it would be interesting to have more vetted wrappers I suppose - if we could have some wrapper for ICMP error packets, we could pass the right object immediately and avoid double-checking that it's indeed ICMP error in here.

(Outside of the scope for this PR, obviously)

net/src/flows/flow_key.rs
```diff
@@ -677,6 +684,64 @@ impl<Buf: PacketBufferMut> TryFrom<Uni<&Packet<Buf>>> for FlowKey {
...
+        .ok_or(FlowKeyError::EmbeddedMissingHeader("ip"))?;
+
+    let src_ip = net.src_addr();
+    let dst_ip = net.dst_addr();
+    let embedded_transport = inner
+        .try_embedded_transport()
+        .ok_or(FlowKeyError::EmbeddedMissingHeader("transport"))?;
+
+    // build the protocol key
+    let proto_key = match embedded_transport {
```
qmonnet, line 723: Or implement `From<EmbeddedTransport> for IpProtoKey`?

nat/src/icmp_handler/nf.rs
```diff
@@ -84,73 +81,17 @@ fn icmp_checksums_ok<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> bool {
...
-
-    let net = inner.try_inner_ip()?;
-    let src_ip = net.src_addr();
-    let dst_ip = net.dst_addr();
-    let proto = net.next_header();
-
-    let embedded_transport = inner.try_embedded_transport()?;
-
-    // FIXME(fredi): see if we should do a TryFrom<EmbeddedTransport> for IpProtoKey
-    // as suggested by Quentin
```
qmonnet, line 103: Oh, looks like it's not the first time I'm suggesting this 😅

nat/src/icmp_handler/nf.rs
```diff
@@ -84,73 +81,17 @@ fn icmp_checksums_ok<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> bool {
...
-        let id = embedded_transport.icmp_identifier()?;
-        debug!("ICMP offending packet: ({proto}) src:{src_ip} dst:{dst_ip} id: {id}");
-    }
-    let flow_key = FlowKey::uni(None, src_ip, dst_ip, proto_key);
-    Some(flow_key)
-}
-
 impl IcmpErrorHandler {
-    /// Handle an ICMP error packet. This processing follows requirements REQ-4 and REQ-5
-    /// of RFC 5508, "NAT Behavioral Requirements for ICMP".
```
qmonnet, line 140: Why dropping this comment?

nat/src/icmp_handler/nf.rs
```diff
@@ -161,46 +102,54 @@ impl IcmpErrorHandler {
...
 
-        let Some(flow_key) = get_icmp_inner_pkt_flowkey(packet) else {
-            debug!("Could not build flow key for ICMP-error embedded packet");
-            packet.done(DoneReason::IcmpErrorIncomplete);
-            return;
+        let flow_key = match flowkey_embedded_in_icmp_error(packet) {
+            Ok(flow_key) => flow_key,
+            Err(e) => {
+                debug!("Could not build flow key for ICMP-error inner packet: {e}");
+                packet.done(DoneReason::IcmpErrorIncomplete);
```
qmonnet, line 139: The function can now return different errors (it no longer fails only if the header is incomplete - which makes me suspicious about the interest of moving and changing `get_icmp_inner_pkt_flowkey()` into `flowkey_embedded_in_icmp_error()`, by the way - so you cannot just use `IcmpErrorIncomplete` here.

nat/src/icmp_handler/nf.rs
```diff
@@ -161,46 +102,54 @@ impl IcmpErrorHandler {
...
         };
 
-        // reverse the flow key and look up for a flow in the flow table.
+        // reverse the flow key, set src vpc and look up for a flow in the flow table.
         let rev_flow_key = flow_key.reverse(Some(icmp_src_vpcd));
         let Some(flow) = self.flow_table.lookup(&rev_flow_key) else {
-            debug!("Found no flow for key={rev_flow_key}");
-            packet.done(DoneReason::Filtered);
+            // There is no flow for the provided flow key. This is not necessarily an error.
+            // the ICMP error packet may correspond to a communication that uses static NAT or no NAT, or
```
qmonnet: I see, hmm but we need to verify whether the embedded packet relates to a flow that is supposed to be stateful NAT. If it does, we SHOULD drop it when there's no entry in the flow table, see REQ-4 / REQ-5 in [RFC 5508](https://datatracker.ietf.org/doc/html/rfc5508).

flow-entry/src/flow_table/nf_lookup.rs
```diff
@@ -39,7 +39,7 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for FlowLookup {
     ) -> impl Iterator<Item = Packet<Buf>> + 'a {
         input.filter_map(move |mut packet| {
             let nfi = &self.name;
-            if !packet.is_done() && packet.meta().is_overlay() {
+            if !packet.is_done() && packet.meta().is_overlay() && packet.meta().dst_vpcd.is_none() {
```
qmonnet, line 42: Please add a comment here, because outside of the context of the commit it's really not obvious why we want `dst_vpcd` to be `None`.

flow-filter/src/lib.rs
```diff
@@ -470,7 +458,10 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for FlowFilter {
     ) -> impl Iterator<Item = Packet<Buf>> + 'a {
         input.filter_map(|mut packet| {
             if let Some(tablesr) = &self.tablesr.enter() {
-                if !packet.is_done() && packet.meta().is_overlay() {
+                if !packet.is_done()
+                    && packet.meta().is_overlay()
+                    && packet.meta().dst_vpcd.is_none()
```
qmonnet, line 463: Same, please comment.

nat/src/stateful/test.rs
```diff
@@ -30,16 +32,95 @@ use net::tcp::TruncatedTcp;
...
 use std::str::FromStr;
 use std::time::Duration;
 use tracectl::get_trace_ctl;
+use tracing::debug;
 use tracing_test::traced_test;
 
 const ONE_MINUTE: Duration = Duration::from_mins(1);
 use crate::stateless::test::build_gwconfig_from_overlay;
 
+fn case(msg: &str) {
```
qmonnet, line 47: Why `case`? Sounds like a switch/case thing 😮

nat/src/stateful/test.rs
```diff
@@ -495,6 +576,33 @@ fn check_packet_icmp_echo(
     )
 }
 
+fn check_packet_icmp_echo_new(
```
qmonnet, line 579: We should change this name. Or did you check whether we could use this new version instead of the older `check_packet_icmp_echo()` for the other test that relies on it, so that we could keep just one version?

nat/src/stateful/test.rs
qmonnet: Oh I see, that's a test case. How about `step` instead, or `test_case` at least?

nat/src/icmp_handler/nf.rs
```diff
@@ -140,18 +140,20 @@ impl IcmpErrorHandler {
...
-            // This can't happen atm since the only NFs that create flows are stateful NAT
-            // and port-forwarding. If we hit a flow that does not have either of those set
-            // that's, atm, a bug.
-            packet.done(DoneReason::Filtered);
+            Err(DoneReason::Filtered)
+        };
+
+        // drop packet if could not translate it
+        if let Err(reason) = result {
+            packet.done(reason);
```
qmonnet: With regards to my previous comments on the RFC requirements for dropping - is this how you address it? I'm not sure I follow exactly what happens in the case of no NAT / static NAT vs. stateful NAT 🤔

nat/src/icmp_handler/nf.rs
```diff
@@ -140,18 +140,20 @@ impl IcmpErrorHandler {
...
             debug!("Icmp error is for vpc {dst_vpcd}. Will process with port-forwarding state");
-            handle_icmp_error_port_forwarding(packet, flow.as_ref());
+            handle_icmp_error_port_forwarding(packet, flow.as_ref())
         } else {
             warn!("Found no NAT state to process ICMP error message. Dropping...");
-            // This can't happen atm since the only NFs that create flows are stateful NAT
-            // and port-forwarding. If we hit a flow that does not have either of those set
-            // that's, atm, a bug.
-            packet.done(DoneReason::Filtered);
+            Err(DoneReason::Filtered)
```
qmonnet, line 184: Nit: I'd find it easier to follow if we just marked the packet as done and returned from the function here, rather than returning an `Err()` for `result` and comparing it right after.

nat/src/icmp_handler/nf.rs
```diff
@@ -81,6 +83,28 @@ fn icmp_checksums_ok<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> bool {
     true
 }
 
+/// Tell if the given ICMP error packet is "unrecoverable" in the sense that any flow related to it *may*
+/// be removed because the source may no longer send packets after receiving the error message.
+fn is_icmp_unrecoverable<Buf: PacketBufferMut>(packet: &mut Packet<Buf>) -> (bool, Option<&str>) {
```
qmonnet, line 89: Great idea, but doesn't this introduce a risk that someone could DDoS by making use drop valid entries from the flow table?

qmonnet (changes requested):
Thanks for this!

I've got a bunch of minor suggestions. Aside from that, I've got two bigger concerns:

- I'd like to make sure we keep dropping ICMP error packets related to packets that should go through stateful NAT, if no related entry is found in the flow table. I thought you removed that, but going further in the PR I think you may in fact cover it. I'd like to chat with you to go through this, if you don't mind.
- The eviction of flows related to some irrecoverable ICMP errors might open a path to DDoS by making use drop valid entries from the flow table?

qmonnet:
Also, the PR misses unit tests to detect regressions on the new fixes, please add some.

nat/src/icmp_handler/nf.rs
```diff
@@ -161,46 +126,71 @@ impl IcmpErrorHandler {
...
         let Some(flow) = self.flow_table.lookup(&rev_flow_key) else {
-            debug!("Found no flow for key={rev_flow_key}");
-            packet.done(DoneReason::Filtered);
+            debug!("Found no flow for key={rev_flow_key}. Letting packet through...");
+            // There is no flow for the provided flow key. This is not necessarily an error since the
+            // ICMP error packet could correspond to a flow that uses static NAT or no NAT at all. It
+            // could correspond to a masqueraded / port-forwarded flow that was just expired (unlikely).
+            // In either case, leave the packet through so that the flow filter deals with it.
             return;
         };
```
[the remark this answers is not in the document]
Fredi-raspall: Fixed. I doubt this has any implication in practice however and requires checking an atomic.

nat/src/icmp_handler/nf.rs
Fredi-raspall: Yes, I moved it to flow_key. I thought of implementing From<EmbeddedTransport> , but the use is so niche that I think it was just clearer to leave it like that.

nat/src/icmp_handler/nf.rs
Fredi-raspall: Because it's written in multiple places and adds little info IMO.

nat/src/icmp_handler/nf.rs
Fredi-raspall: Sorry, I don't understand the comment. Can you explain?

nat/src/stateful/test.rs
Fredi-raspall: I wanted to reuse it, but there are many tests there and very hard to update right. I was actually going to open an issue to get all the tests fixed because it took me nearly as much (if not more) to fix the tests than to code the whole PR. I will open it.

nat/src/icmp_handler/nf.rs
Fredi-raspall: If the packet belongs to a communication without NAT or static NAT, it will not be handled by the ICMP error handler. Is that the confusion?

nat/src/icmp_handler/nf.rs
Fredi-raspall: let me have a look....

nat/src/icmp_handler/nf.rs
Fredi-raspall: I thought about it, but now I realize I forgot to add a sanity required to avoid that.

Fredi-raspall:
> Also, the PR misses unit tests to detect regressions on the new fixes, please add some.

I adjusted some of the tests that should cover regressions now. As mentioned, I think that the existing tests need some work outside the scope of this PR to make that possible.

nat/src/icmp_handler/nf.rs
qmonnet: It's niche, but it would look cleaner to me. That's not the only place where we'd implement a trait for a single consumer.

nat/src/icmp_handler/nf.rs
qmonnet: I find it useful to keep track of the RFC we base the logic on. I can only find two mentions of it, in different files :confused:

nat/src/icmp_handler/nf.rs
qmonnet: The message associated with `IcmpErrorIncomplete` is `Dropping ICMP error packet: embedded packet missing`. But `flowkey_embedded_in_icmp_error()` can also fail if `!packet.is_icmp_error()`, in which case the message is not consistent with the actual code path that led to the error.

nat/src/icmp_handler/nf.rs
Fredi-raspall: Added sanity

nat/src/icmp_handler/nf.rs
Fredi-raspall: I checked. Note that as is, the result catches the outcome of all possible cases.

nat/src/icmp_handler/nf.rs
qmonnet: Discussed offline, it's addressed elsewhere

nat/src/icmp_handler/nf.rs
qmonnet: We chatted about that - we're good here.

flow-entry/src/flow_table/nf_lookup.rs
qmonnet: Please follow-up on the comment here, or in two months we won't remember why we check the destination VPC here and someone might try to remove it by mistake.

flow-filter/src/lib.rs
qmonnet: Same here.

qmonnet (approved):
We discussed about my main concerns, the need to drop flows for stateful NAT when we find no entries in the flow table and the potential DoS via the invalidation of entries on ICMP errors, and these have been addressed.

I'd like to see some of the other comments addressed, including the tests, but we agreed we'd go ahead to have these fixes in 26.02.

Thanks a lot for these fixes!

PR #1512, opened 2026-05-04
extend stateful nat tests + small fixes

Fredi-raspall:
fixes https://github.com/githedgehog/internal/issues/374

### DIFF
--- a/nat/src/stateful/nf.rs
+++ b/nat/src/stateful/nf.rs
@@ -13,7 +13,7 @@ use crate::stateful::flows::check_masquerading_flow;
 use crate::stateful::packet::{NatPacketError, NatTranslate, masquerade};
 use crate::stateful::protocol::next_flow_status;
 use crate::stateful::state::MasqueradeState;
-use concurrency::sync::Arc;
+use concurrency::sync::{Arc, Weak};
 use flow_entry::flow_table::table::{FlowTable, FlowTableError};
 use net::buffer::PacketBufferMut;
 use net::flow_key::{IcmpProtoKey, Uni};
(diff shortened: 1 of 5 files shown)


nat/src/stateful/nf.rs
```diff
@@ -152,6 +152,19 @@ impl StatefulNat {
...
+        };
+
+        // extend the duration of the flow according to the new status
+        if let Some(extend_by) = extend_by {
+            let _ = flow_info.reset_expiry_unchecked(extend_by);
+            // if we transition to established, let the related flow get the configured timeout too
+            if current != new_status
+                && new_status == NatFlowStatus::Established
+                && let Some(related) = flow_info.related.as_ref()
+                && let Some(related) = related.upgrade().as_ref()
```
mvachhar: Can we restructure this using `and_then`?

nat/src/stateful/nf.rs
```diff
@@ -362,6 +362,13 @@ impl StatefulNat {
             return Err(StatefulNatError::NoAllocator);
         };
 
+        // if packet contains TCP, do not create flows nor translate state unless it is a SYN
+        if let Some(tcp) = packet.try_tcp()
+            && (!tcp.syn() || tcp.ack() || tcp.rst())
```
mvachhar: I think tcp.syn() is fine with a TODO comment to state that we should ensure that this is a well formed SYN packet.

nat/src/stateful/test.rs
```diff
@@ -1547,3 +1579,295 @@ async fn test_full_config_unidirectional_nat_overlapping_exposes_for_single_peer
...
+    let src_port = packet.transport_src_port().unwrap();
+    let dst_port = packet.transport_dst_port().unwrap();
+
+    let mut reply = packet.clone();
+    reply.meta_reset();
+    reply.meta_mut().src_vpcd = dst_vpcd;
+    reply.meta_mut().set_stateful_nat(true);
+    reply.meta_mut().set_overlay(true);
+
+    reply.set_eth_source(dst_mac).unwrap();
```
mvachhar, line 1652: Not needed now, but the new builder is probably a better approach to handle this construction.

mvachhar (changes requested):

mvachhar (approved):
In the interest of time I'm going to approve this, but please make the changes I requested before merging.

nat/src/stateful/nf.rs
Fredi-raspall: Ok. I've left the || ack() since that would 100% be a bad case.

nat/src/stateful/nf.rs
Fredi-raspall: Done

PR #1514, opened 2026-05-04
address opengrep finding

daniel-noland:
With this all opengrep findings are fixed and future findings are marked as CI fail.

This tool has proven to be low spam so far, so I think this is a solid idea


qmonnet (approved):
Looks OK

sergeymatov (approved):

PR #1527, opened 2026-05-11
refactor(lpm): Merge traits IpPrefix{Covering,Colliding} into IpPrefix

qmonnet:
There's no good reason to keep these traits separate, let's merge them into IpPrefix.

This is a small clean-up that I came up with while looking at larger changes for the IP prefixes types. I'm not sure to complete this larger piece of work, so here's at least the small clean-up.


daniel-noland (approved):
yup, that looks reasonable

PR #1528, opened 2026-05-11
POC: musl libc would work :muscle:

daniel-noland:
Not for merge.  This is just a mechanism to show that the musl libc build works as a review aide for #1524

> [!NOTE]
> I am NOT proposing this switch.  This is intended as a stress test for our cross compile env.


daniel-noland:
closing since this is just a proof.  No real intention to merge from the start

PR #1530, opened 2026-05-12
Fix performance of tracing rate limiter

sergeymatov:
Move to own atomic operations for per-level filter for callsites. Reloading of interest is done via ArcSwap rather then RwLock

Layered structure of throtteled/unthrotteled filters produces atomic reloading of `Interest` for each callsite. `Layered<EnvFilter>::enable()` invokes every time on each of tracing message, but `tracing_subscriber::reload` path that was previously used to check atomic state spawned `RwLock::read_contended ` on each of such calls.
Pyroscope flame graph analysis showed that those operations were causing extreme cost of CPU operations.

Proposed solution is to use `ArcSwap` and precisely update callsite Interest for each of the `Layer`

### DIFF
--- a/tracectl/src/control.rs
+++ b/tracectl/src/control.rs
@@ -5,17 +5,31 @@
 
 use ordermap::OrderMap;
 use std::str::FromStr;
-use std::sync::{Arc, LazyLock, Mutex};
-use std::{collections::HashSet, sync::MutexGuard};
+use std::sync::{Arc, Mutex, OnceLock};
+use std::{collections::HashSet, sync::MutexGuard, time::Duration};
 use thiserror::Error;
 #[allow(unused)]
-use tracing::{debug, error, info, warn};
-use tracing_subscriber::{EnvFilter, Registry, filter::LevelFilter, prelude::*, reload};
+use tracing::{Level, Subscriber, debug, error, info, warn};
+use tracing_subscriber::{
+    EnvFilter, Registry,
+    filter::{FilterExt, LevelFilter, filter_fn},
+    layer::{Filter, Layer},
+    prelude::*,
+    registry::LookupSpan,
+    reload,
+};
+use tracing_throttle::{Policy, TracingRateLimitLayer};
 
 use crate::display::TargetCfgDbByTag;
 use crate::targets::{TRACING_TAG_ALL, TRACING_TARGETS};
 use crate::trace_target;
 trace_target!("tracectl", LevelFilter::INFO, &[]);
+
+#[derive(Copy, Clone, Debug)]
+pub struct TracingRateLimitConfig {
+    pub burst: u32,
+    pub replenish_per_second: u32,
+}
 
 #[derive(Clone, Debug, Error, PartialEq)]
 pub enum TraceCtlError {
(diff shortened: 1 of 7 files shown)


tracectl/src/control.rs
```diff
@@ -3,20 +3,21 @@
...
+use tracing::{Level, Subscriber, callsite, debug, error, info, warn};
 use tracing_subscriber::{
-    EnvFilter, Registry,
+    EnvFilter,
     filter::{FilterExt, LevelFilter, filter_fn},
-    layer::{Filter, Layer},
+    layer::{Context, Filter, Layer},
     prelude::*,
     registry::LookupSpan,
-    reload,
```
mvachhar: This package implicityl has the RwLock being referenced, but it is not obvious that this is the case.  Can you please expand the commit message to explain where the removed RwLock was and what was done to remove it.

sergeymatov: Done in the commit description

tracectl/src/control.rs
```diff
@@ -278,38 +297,230 @@ impl TargetCfgDb {
...
+    fn throttled_level_filter<S>() -> impl Filter<S> + Send + Sync
+    where
+        S: Subscriber + for<'span> LookupSpan<'span>,
+    {
+        filter_fn(|meta| {
+            matches!(
+                *meta.level(),
+                Level::INFO | Level::TRACE | Level::ERROR | Level::WARN
+            )
+        })
```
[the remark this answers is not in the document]
sergeymatov: As designed, we `do not throttle DEBUG traces`

tracectl/src/control.rs
```diff
@@ -292,13 +297,109 @@ impl TargetCfgDb {
...
-type BoxedTracingLayer<S> = Box<dyn Layer<S> + Send + Sync>;
+/// `AtomicEnvFilter` wraps an `EnvFilter` in an `ArcSwap` to allow atomic swapping of the filter
+struct AtomicEnvFilter {
+    inner: Arc<ArcSwap<EnvFilter>>,
+}
+
+/// Handle used to atomically swap in a new `EnvFilter` from outside the
+/// subscriber. Mirrors `reload::Handle`.
+#[derive(Clone)]
+struct AtomicEnvFilterHandle {
```
Fredi-raspall: This type is identical to `AtomicEnvFilter`, except that it implements Clone.
Why do we need two types that are the same?

Fredi-raspall: Asking because both types are private to this crate?

tracectl/src/control.rs
```diff
@@ -428,10 +529,14 @@ impl TracingControl {
             .lock()
             .map_err(|e| TraceCtlError::LockFailure(e.to_string()))
     }
+    /// Reload the active `EnvFilter`. Infallible with the `ArcSwap`-based
+    /// handle: there's no lock to poison and no subscriber-gone case. The
+    /// `Result` return type is kept for API stability with the previous
+    /// version that used `reload::Handle::reload`.
+    #[allow(clippy::unnecessary_wraps)]
```
Fredi-raspall: If this is not going to be fallible anymore, I'd add a third commit removing the errors and their handling.

tracectl/src/control.rs
```diff
@@ -3,23 +3,42 @@
...
 
+#[derive(Copy, Clone, Debug)]
+pub struct TracingRateLimitConfig {
+    pub burst: u32,
+    pub replenish_per_second: u32,
+}
+
 #[derive(Clone, Debug, Error, PartialEq)]
 pub enum TraceCtlError {
+    // Preserved for API compatibility. With the `ArcSwap`-based filter swap
```
Fredi-raspall: I'd remove this in a third commit.

Fredi-raspall (changes requested):
Looks good to me overall, but I wish the PR included several changes.
1) why two types for AtomicEnvFilter? Any chance a single type suffices?
2) I'd remove the no longer used Result cases.
3) I would like to see a test where a bunch of logs are produced and rate-limitted. There is a way to know if an event would be traced under a given config. Unsure if that is extensible when you have rate-limiting. It would be good to at least be able to see the effect.
4) Super ideally: I wonder if two subscribers (one with rate limiting, one without) could be compared (profiled) to be able to see the impact in performance of the rate limiter. Unsure if this can be done in any meaningful / reliable way.

tracectl/src/control.rs
sergeymatov: done

tracectl/src/control.rs
sergeymatov: done

sergeymatov:
I will add test as separate commit. Thanks @Fredi-raspall

sergeymatov:
@Fredi-raspall There is a simple test that measures emitted logs among throttled

Fredi-raspall (approved):
LGTM @sergeymatov !
If anything, in the new test, I would also emit DEBUG traces and check that all of them make it and that they do not affect the throttling of other traces. I believe this is an easy addition, but will approve this PR and let it up to you to add this.

PR #1537, opened 2026-05-14
perf(miri): trim heavy tests so the miri job finishes in reasonable time

daniel-noland:
Three independent miri-only adjustments collected for the per-CI-run budget. All are no-ops outside `cfg(miri)`:

* `config/src/utils/collapse.rs`: the bolero generator for `test_bolero_collapse_prefix_lists` shrinks from (prefix_max=100, exclude_max=100, addr_count=1000) to (10, 10, 100) under miri. Miri's borrow-stack cost on prefix-trie traversal is ~100-1000x; the algorithmic coverage is largely the same with the smaller shape.
* `nat/src/stateless/test.rs`: `test_full_config` was dominated by `#[traced_test]` subscriber dispatch (formatting, TLS atomics, layer walk, capture mutex) on per-packet `debug!` calls; hundreds of events per run, each disproportionately expensive under miri. Skip the subscriber install under miri, gate out the `println!` debug dump, and use `[packet]` rather than `vec![packet]` to drop one per-call allocation.
* `miri.just`: weak_failure_rate 0.15 -> 0.05. The 0.15 value was inflating miri's weak-memory failure injection beyond what the protocol invariants are tuned for; 0.05 still exercises the path and gives shuttle headroom for other interleavings.  We should still (sometimes) run with higher weak failure rates, but no need to burn endless resources on it.

### DIFF
--- a/config/src/utils/collapse.rs
+++ b/config/src/utils/collapse.rs
@@ -548,10 +548,21 @@ mod tests {
 
     #[test]
     fn test_bolero_collapse_prefix_lists() {
-        let generator = PrefixExcludeAddrsGenerator {
-            prefix_max: 100,
-            exclude_max: 100,
-            addr_count: 1000,
+        let generator = cfg_select! {
+           miri => {
+               PrefixExcludeAddrsGenerator {
+                    prefix_max: 10,
+                    exclude_max: 10,
+                    addr_count: 100,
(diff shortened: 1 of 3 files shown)


Fredi-raspall (approved):
Looks good to me, thanks!

PR #1540, opened 2026-05-15
feat(concurrency): poison-as-panic wrapper for std::sync

daniel-noland:
## Summary

- New `concurrency::sync` module exposing `Mutex` / `RwLock` with a `parking_lot`-shaped naked-guard surface backed by a poison-as-panic wrapper around `std::sync`. Workspace policy treats a crashed thread as a crashed process: poison surfacing inside a lock acquire panics through `concurrency::sync::poisoned()` rather than handing back a possibly-torn `LockResult`.
- One cold `#[inline(never)] fn poisoned() -> !` keeps the hot path branch-free; the wrapper is a single `match` per acquire, all `Send/Sync`/`Default`/`From`/`Display` impls forward to the wrapped type.
- `RwLock::upgradable_read` is implemented as exclusive `write()` for now; sound but lossy. Documented at the backend module level.
- `Arc`, `Weak`, atomics, `Condvar`, `mpsc`, `OnceLock`, etc. are re-exported from `std::sync` unchanged.
- `concurrency::quiescent` migrates from `std::sync::{Arc, Mutex}` to `crate::sync::{Arc, Mutex}` so the QSBR primitive consumes the facade.

PR 2 of 6. Depends on **#1539 — absorb-quiescent**. The next PR layers `parking_lot` as the default backend on top of this.

### DIFF
--- a/concurrency/src/lib.rs
+++ b/concurrency/src/lib.rs
@@ -13,12 +13,10 @@
 #![allow(missing_docs)]
 
 pub mod macros;
+pub mod sync;
 
 #[cfg(all(miri, any(feature = "shuttle", feature = "loom")))]
 compile_error!("miri does not meaningfully support 'loom' or 'shuttle'");
-
-#[cfg(not(any(feature = "loom", feature = "shuttle")))]
-pub use std::sync;
 
 #[cfg(not(any(feature = "loom", feature = "shuttle")))]
 pub use std::thread;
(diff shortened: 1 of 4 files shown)


daniel-noland:
Folded into #1542 — the std-sync wrapper, parking_lot default, and consumer sweep can't be split without leaving the workspace uncompilable between them, so this stack collapses 2+3+4 into a single PR.

PR #1541, opened 2026-05-15
feat(concurrency): use parking_lot for the default backend

daniel-noland:
## Summary

- Add a `parking_lot` Cargo feature (enabled by default) and a `concurrency/src/sync/parking_lot_backend.rs` that is a zero-cost re-export of `parking_lot::{Mutex, RwLock}` plus the std re-exports for everything `parking_lot` doesn't ship (`Arc`, `Weak`, atomics, `Condvar`, ...).
- Backend routing in `concurrency/src/sync/mod.rs` picks `parking_lot_backend` when `feature = "parking_lot"` is on, falls back to `std_backend` otherwise.
- `_strict_provenance` feature overrides `parking_lot` and routes through `std_backend`: `parking_lot_core::word_lock` uses integer-to-pointer casts that `-Zmiri-strict-provenance` rejects, and the CI miri job (`--features=_strict_provenance`) exercises the fallback slot, which needs `std::sync` underneath.
- Workspace `Cargo.toml` gets a `tokio` parking_lot override entry so the `tokio` parking_lot feature is enabled workspace-wide.

PR 3 of 6. Depends on **#1540 — facade-std**. The next PR sweeps the workspace data-path crates to consume the facade.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1236,6 +1236,7 @@ dependencies = [
  "bolero",
  "dataplane-concurrency-macros",
  "loom",
+ "parking_lot",
  "shuttle",
  "static_assertions",
 ]
@@ -5629,6 +5630,7 @@ dependencies = [
  "bytes",
  "libc 0.2.186",
  "mio",
+ "parking_lot",
  "pin-project-lite",
  "signal-hook-registry",
  "socket2",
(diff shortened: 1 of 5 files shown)


daniel-noland:
Folded into #1542 — same reason as #1540.

PR #1542, opened 2026-05-15
(2+3+4) concurrency / dpdk

daniel-noland:
## Summary

Introduce a `parking_lot`-shaped facade over `Mutex` / `RwLock`, ship the four backends together (std / parking_lot / shuttle / loom), and rewire the data-path crates through it in a single atomic step.

The four pieces are bundled because a partial sweep leaves the workspace in a state where the facade and its consumers disagree on whether `Mutex::lock` returns `LockResult<Guard>` or a naked guard. The shuttle/loom wrappers strip `LockResult` in exactly the same poison-as-panic shape as the std wrapper, so splitting them re-introduces that disagreement under the `test/shuttle` and `test/loom` CI jobs.

### std::sync wrapper (poison-as-panic)

- New `concurrency::sync` module exposing `Mutex` / `RwLock` with a `parking_lot`-shaped naked-guard surface backed by a poison-as-panic wrapper around `std::sync`. Workspace policy treats a crashed thread as a crashed process: poison surfacing inside a lock acquire panics through `concurrency::sync::poisoned()` rather than handing back a possibly-torn `LockResult`.
- One cold `#[inline(never)] fn poisoned() -> !` keeps the hot path branch-free; the wrapper is a single `match` per acquire.
- `RwLock::upgradable_read` is implemented as exclusive `write()` for now -- sound but lossy. Documented at the backend module level.
- `Arc`, `Weak`, atomics, `Condvar`, `mpsc`, `OnceLock`, etc. are re-exported from `std::sync` unchanged.

### parking_lot as the production default

- Adds a `parking_lot` Cargo feature (enabled by default) and `concurrency/src/sync/parking_lot_backend.rs` -- a zero-cost re-export of `parking_lot::{Mutex, RwLock}` plus the std re-exports for everything `parking_lot` doesn't ship.
- Backend routing in `concurrency/src/sync/mod.rs` picks `parking_lot_backend` when `feature = "parking_lot"` is on, falls back to `std_backend` otherwise.
- `_strict_provenance` feature overrides `parking_lot` and routes through `std_backend`: `parking_lot_core::word_lock` uses integer-to-pointer casts that `-Zmiri-strict-provenance` rejects, and the CI miri job (`--features=_strict_provenance`) exercises the fallback slot, which needs `std::sync` underneath.
- Workspace `Cargo.toml` gets a `tokio` parking_lot override entry so the `tokio` parking_lot feature is enabled workspace-wide.

### shuttle backend

- `concurrency/src/sync/shuttle_backend.rs` mirrors `std_backend.rs`: a thin poison-as-panic wrap around `shuttle::sync::{Mutex, RwLock}` returning naked guards, plus re-exports for `Arc` / `Weak` / atomics / `Condvar` / `mpsc` that pass through unchanged. `OnceLock` falls back to `std::sync` because shuttle doesn't ship one.
- Three feature flags share one `dep:shuttle`, arranged as a chain (`shuttle_dfs` -> `shuttle_pct` -> `shuttle`):
  - `shuttle` -- random scheduler (default first choice).
  - `shuttle_pct` -- PCT, biases toward rare interleavings.
  - `shuttle_dfs` -- DFS, exhaustive small-state exploration.
  Same backend, scheduler chosen at runtime (by `concurrency::stress`, added in the follow-on PR). The chain means a single `feature = "shuttle"` cfg check is true under every variant, so the routing in `sync/mod.rs` collapses to single-feature checks (no `any(shuttle, shuttle_pct, shuttle_dfs)` chains).

### loom backend

- `concurrency/src/sync/loom_backend.rs` mirrors the std and shuttle wrappers: poison-as-panic around `loom::sync::{Mutex, RwLock}` returning naked guards, with `LockResult` / `TryLockError` / `OnceLock` etc. re-exported from `std::sync` because loom doesn't ship them.
- Loom 0.7 has no `Weak<T>` and no `Arc::downgrade`. The backend adds a local `Arc<T>` wrapper that adds `downgrade` plus a `Weak<T>` shim that holds a *strong* clone of the inner `loom::sync::Arc` until the `Weak` itself drops. Documented quirks (`Weak::upgrade` always succeeds after a `downgrade`, `Arc::weak_count` panics rather than silently returning 0, last `Arc` drop does not free when a `Weak` is live) are explicit at the module level -- silent fallbacks would let assertions pass for the wrong reason on every backend.
- `sync/mod.rs` routes the loom feature ahead of every other backend.

### Workspace transition

- Sweep the data-path crates (`flow-entry`, `flow-filter`, `nat`, `net`, `pipeline`, `routing`, `stats`) to consume `dataplane_concurrency::sync::{Arc, Mutex, RwLock, atomic, ...}` instead of `std::sync::*` / `parking_lot::*` directly.
- Call sites drop the `.unwrap()` / `.expect("poisoned")` noise the `LockResult`-shaped surface forced, since the facade returns naked guards. Same shape as the existing `parking_lot::Mutex` consumers.
- `concurrency_mode(std)` annotations at QSBR-touching test sites in `nat::stateful::apalloc::test_alloc`, `routing::fib::test`, and `flow-entry::flow_table::table::tests` keep the existing std-only feature gating intact.
- `nat::portfw::flow_state::get_packet_port_fw_state` simplification: the explicit `drop(guard)` dance before `debug!` calls is removed (per the recorded author intent that it isn't worth the code complexity); the guard now drops implicitly at the end of scope.

The sweep is mostly mechanical (`use` rewrites and `.unwrap()` removal). The semantic change is one step: every consumer now gets the facade-routed primitives, so flipping the loom/shuttle features at the top of the workspace is now meaningful.

PR 2 of 3 in the concurrency facade stack (the original 4-PR plan collapsed the shuttle-only and loom-only PRs into this one because the wrappers, the workspace sweep, and the consumer changes can't be independently merged without leaving CI red). Depends on **#1539 -- absorb-quiescent**. The follow-on PR (#1544) adds `#[concurrency::test]`, the `stress` dispatcher, and the `thread::scope` shim.


mvachhar (changes requested):
You need to resolve merge conflicts here so while you are doing that, you should also clean up the commits a bit.  There are some changes related to the move of QSBR into concurrency in the second commit when they should be in the first.  I think AI should be able to fix that pretty easily so it is worth doing.  If it takes too long, it isn't worht it and we can merge as is.

daniel-noland:
> You need to resolve merge conflicts here so while you are doing that, you should also clean up the commits a bit. There are some changes related to the move of QSBR into concurrency in the second commit when they should be in the first. I think AI should be able to fix that pretty easily so it is worth doing. If it takes too long, it isn't worht it and we can merge as is.

~~Sure.  Can scrub it a bit~~

Err, sorry, #1539 (absorb) is already merged, the QSBR-move-related hunks that leaked into commit 2 are stuck here. There's no commit 1 left on this branch to fold them into

mvachhar (approved):

concurrency/src/slot.rs
```diff
@@ -3,15 +3,28 @@
 
 //! Single-slot atomic publication.
 //!
-//! In production this is `arc_swap::ArcSwap` -- lock-free read fast path,
-//! which is what makes [`Subscriber::snapshot`] cheap on the data-plane.
+//! In production these are `arc_swap::ArcSwap` / `ArcSwapOption` --
```
Fredi-raspall, line 6: What is the reason to call it slot / slotoption instead of the original ?

PR #1544, opened 2026-05-15
(3) #[concurrency::test] macro and thread::scope shim

daniel-noland:
## Summary

The capstone PR. Ship the test-time infrastructure that lets a single source file run under every backend the facade (in #1542) routes to.

Three pieces, plus the test suite that exercises them.

### `concurrency::stress` + `#[concurrency::test]`

- `concurrency::stress(body)` dispatches to `loom::model` / `shuttle::check_random` / `_pct` / `_dfs` / direct call based on the active feature. The dispatch has a defensive `_ => compile_error!` arm: if a new model-checker feature is added to the exclusion list but no matching dispatch arm is wired up, the build fails loudly rather than silently skipping exploration.
- `#[concurrency::test]` proc-macro expands to `#[test]` plus `stress(|| body)`. Under model-checker backends it emits a nested `mod <fn_name> { mod concurrency_model { #[test] fn <backend>() { ... } } }` so test reports / JUnit show the active backend as the leaf name (e.g. `snapshot_observes::concurrency_model::loom`). Default backend stays flat. `#[should_panic]` / `#[ignore]` thread through; `async fn` and `fn(args)` are rejected at parse time.
- Crate path is resolved at expansion time via `proc-macro-crate`, so consumers under any alias work without setup; integration tests inside this crate use `extern crate dataplane_concurrency as concurrency;` since cargo rejects self-deps.
- `stress` is re-exported at the crate root for the macro expansion but hidden from rustdoc (`#[doc(hidden)]`) -- users land on the macro.

### `thread::scope` shim

- `concurrency::thread::scope` works on every backend. Std and shuttle re-export theirs; loom 0.7 doesn't have `scope`, so we ship `concurrency/src/thread/loom_scope.rs`: built on `loom::spawn` plus an `Arc<Mutex<Option<T>>>` keepalive that preserves the same drop-affinity guarantee `std::thread::scope` gives.
- Drop-affinity is enforced by an explicit `take_payload` on main during teardown, not by an Arc-count assertion. Loom's `JoinHandle::join` synchronises only on `notify`, which can fire before the spawned thread's wrapper has finished dropping its captures -- so the count-based check fires spuriously in some schedules. Manual `take` on main gives the same guarantee unconditionally.
- One `mem::transmute` to lift the spawned closure's `'scope` to `'static` (loom 0.7 has no `spawn_unchecked`); the keepalive trait object keeps its honest `'scope` and `ScopeInner<'scope>` lives in `ManuallyDrop` so dropck doesn't constrain `'scope` past `scope()`'s body. Per-site SAFETY comments at both the transmute and the `ManuallyDrop::drop`.

### Tests

- `tests/quiescent_model.rs` replaces the loom-only `quiescent_loom.rs`. One source, every backend, via `#[concurrency::test]` + `thread::scope`.
- `tests/thread_scope.rs` -- contract tests for `thread::scope` (single/multi-spawn, drop affinity, nested-scope re-entry).
- `tests/scope_property.rs` -- bolero x shuttle conservation property over generated spawn/op plans.
- `tests/stress_dispatch.rs` -- `stress` invokes body once on default, more than once on model-check backends; plus `#[should_panic]` / `#[ignore]` attribute pass-through.
- `tests/arc_weak.rs` -- `Arc` / `Weak` contract (passes under every backend, plus loom-only quirk tests). File-level `#![cfg(not(feature = "shuttle_pct"))]` -- PCT can't run single-threaded protocol checks.
- `justfile` nextest filter: empty under loom (the archived binaries are already feature-filtered), `shuttle` substring under shuttle (matches both `quiescent_shuttle` and the new `concurrency_model::shuttle` leaves).

### Documentation

- `lib.rs` gains a "Compiles under loom != exhaustively checked under loom" section listing each documented shim limitation (Weak strong-clone semantics, `weak_count` panic, RwLock upgradable_read, OnceLock ordering invisibility, Mutex::new const divergence) with the workspace consequence -- most notably calling out NAT's `Weak::upgrade().is_none()` patterns and why NAT isn't in the loom matrix today.
- `sync/mod.rs` gains a "Portability footguns the facade does not paper over" section.

PR 3 of 3 in the concurrency facade stack. Depends on **#1542 -- sync facade + backends + workspace sweep**. After this lands, the stack reaches feature parity with the original PR.


mvachhar (approved):
One question here, if `::stress` enables loom everywhere (it currently doesn't because of `Weak`), aren't there many cases where the model checker will fail to terminate or consume too much time/memory?  Again, no change needed if this is not currently and issue, but it is a concern.

daniel-noland:
> One question here, if `::stress` enables loom everywhere (it currently doesn't because of `Weak`), aren't there many cases where the model checker will fail to terminate or consume too much time/memory? Again, no change needed if this is not currently and issue, but it is a concern.

I really have not seen that so far.  The more common issue is just failing to compile under loom.  If you give loom an appropriate domain then it finishes very quickly

PR #1548, opened 2026-05-20
Static NAT + Masquerading / Static NAT + Port Forwarding

### LINKED ISSUE #998: NAT: Investigate stateless + stateful NAT
We should be able to support stateless (destination NAT) + stateful (source NAT) for traffic. This needs some investigation and testing; configuration validation; documentation.

### LINKED ISSUE #1273: Split stateless NAT stage into two
There are certain combinations of NAT within a peering that will not work today the way the network functions are organized, since the execution order matters. One solution to the problem is splitting the stateless NAT NF into two (one for source, one for destination) and placing them at the opposite sides of the pipeline (sourceNAT + NF + NF ... + DstNAT).

qmonnet:
The first commit are some preparatory clean-ups for the static NAT code:

- **refactor(nat): Move static NAT network function to dedicated file**
- **refactor(nat): Move unicast check closer to source IP mapping lookup**
- **refactor(nat): Handle zero-port check at port mapping lookup**
- **feat(net): Add wrappers for transport with types (TCP/UDP)**
- **refactor(nat): Use TcpUdp view in static NAT to simplify port update**
- **refactor(net): Simplify metadata flag toggling**

Then we add some finer control for running static NAT on a packet:

- **chore(flow-filter): Split requires_stateless_nat() into source/dest**
- **feat(flow-filter): Tag packets for src/dst static NAT**
- **feat(nat): Mark packets as NAT-ed for source or destination**
- **feat(nat): For static NAT, only apply NAT for relevant direction(s)**

We proceed to the changes we need to support combinations of NAT modes:

- **feat(dataplane): Move static NAT stage before port forwarding**
- **feat(flow-filter,nat): Store static NAT requirements in flow info**
- **feat(net): Embed flow key in packet metadata**
- **feat(nat): Use initial IP addresses for forward flow table entry**

We add support for ICMP Error messages with the NAT combinations (thanks Fredi!)

- **fix(nat): Remove NoMappingFound error**
- **refactor(nat): Move NAT tables reader entering to process_packet()**
- **feat(nat): Pass a NatTablesReaderFactory to the IcmpErrorHandler NF**
- **feat(nat): Add static NAT support in IcmpErrorHandler**

We update the validation step and add tests:

- **feat(config): Allow static + stateful NAT, on opposite ends**
- **test(nat): Add tests for static NAT + {masquerade, port forwarding}**

Please refer to individual commit description for details.

## Summary

This PR refactors static NAT into a dedicated StatelessNat NF and extends NAT to support directional static NAT combined with masquerading and port forwarding, plus many related fixes, refactors, and tests.

## Key Changes

- Core refactor & new NF
  - Introduced StatelessNat network function (nat/src/stateless/nf.rs) with NatTablesReader/Writer and constructors (new/with_reader).
  - Moved static NAT logic into a dedicated module and adjusted module re-exports (nat/src/stateless/mod.rs).

- Transport & header helpers
  - Added TCP/UDP transport wrappers and helpers (net/src/tcp_udp.rs) and exported tcp_udp module (net/src/lib.rs) to provide unified port access for TCP/UDP.

- Packet metadata & directional NAT
  - Replaced the single stateless-NAT predicate with directional requirements: requires_static_nat_src / requires_static_nat_dst in PacketMeta and flow-filter logic.
  - Replaced single natted/req-stateless flags with per-direction flags: is_src_natted/is_dst_natted and set_static_nat_src/set_static_nat_dst; MetaFlags now include NATTED_SRC/NATTED_DST and REQ_STATIC_NAT_SRC/REQ_STATIC_NAT_DST.
  - PacketMeta now carries an optional flow_key: Option<Box<FlowKey>> so stages can preserve the packet's initial flow key when creating forward flow-table entries.

- Pipeline & flow-info changes
  - Inserted the static NAT stage before port-forwarding in the packet-processing pipeline (dataplane/src/packet_processor/mod.rs).
  - Added FlowInfoFlags bitflags and stored them in FlowInfo; FlowInfo::related_pair accepts flags to carry static-NAT requirements into flow state (net/src/flows/flow_info.rs).
  - Flow creation for masquerade and port-forwarding now distinguishes initial vs current flow keys and prefers packet.meta().flow_key when present (nat/src/stateful/nf.rs, nat/src/portfw/flow_state.rs).

- Stateless NAT implementation & ICMP
  - StatelessNat performs directional source/destination static NAT using per-VNI NAT tables; supports address+port mappings and uses NonZero for translated ports; source mappings validate unicast IPs (nat/src/stateless/setup/tables.rs).
  - StatelessNat handles embedded inner headers inside ICMP error messages and maps translation failures to DoneReason values (nat/src/stateless/nf.rs).
  - ICMP error handling updated to set static-NAT metadata flags and refresh checksums where needed (nat/src/icmp_handler/nf.rs). Icmp error translation errors are mapped into DoneReason (nat/src/icmp_handler/icmp_error_msg.rs).

- Port-forwarding & stateful changes
  - Port-forwarding flow-key construction is now fallible; build_portfw_flow_keys returns Result and do_port_forwarding handles failures without panics (nat/src/portfw/*).
  - PortFwTableWriter can be populated from a validated VPC table (update_from_vpc_table) (nat/src/portfw/portfwtable/access.rs).
  - Masquerade flow creation preserves initial/current key distinction and uses the initial key for allocation/flow binding (nat/src/stateful/nf.rs).
  - Packet NAT application sets src/dst natted flags when modifications occur; debug_assert! guards prevent double NAT application (nat/src/portfw/packet.rs, nat/src/stateful/packet.rs).

- Configuration & tests
  - Relaxed overlay peering NAT compatibility validation to allow static NAT on one end with stateful NAT or port-forwarding on the other; updated validation logic and tests (config/src/external/overlay/vpc.rs and validation tests).
  - Updated tests and helpers across flow-filter and nat to use directional static-NAT metadata and the new FlowKey API.
  - Added extensive end-to-end tests (nat/src/test.rs) verifying:
    - Static NAT + masquerade (UDP, session establishment)
    - Static NAT + port forwarding (UDP, port mapping)
    - ICMP destination-unreachable handling for both combinations (outer and embedded header translations)
  - Adjusted long_dyn_pipeline test from 1000 to 999 stages to avoid stack exhaustion caused by increased packet/meta size (pipeline/src/pipeline.rs).

- FlowKey / public-surface & other refactors
  - Refactored FlowKey from enum/wrapper (Uni/FlowKeyData) to a direct struct with FlowKey::new(...) and TryFrom<&Packet> implementations; removed FlowKeyData, Uni wrapper, and related public accessors (net/src/flows/flow_key.rs).
  - Updated many tests and modules to construct FlowKey via FlowKey::new or TryFrom<&Packet> (flow-entry, flow-filter, flow-entry tests, flow-table tests, stateful flows, etc.).
  - Display logic updated to render FlowKey fields directly (net/src/flows/display.rs).
  - Added FlowInfoFlags and exposed FlowInfo::get_flags; extended FlowInfo::related_pair signature to accept flags.

- API & public-surface changes
  - Added public types/methods: FlowInfoFlags, FlowInfo::get_flags, extended FlowInfo::related_pair signature, PacketMeta.flow_key and per-direction APIs, TcpUdp types and trait impls, StatelessNat struct and nf exports.
  - Removed: ValidatedExpose::has_nat helper, PacketMeta requires_stateless_nat/set_stateless_nat accessors, and the previous NatTablesReader re-export from stateless mod.

- Error mapping & minor utilities
  - Added From<&IcmpErrorMsgError> for DoneReason and From<&AllocatorError> for DoneReason to map internal errors to termination reasons.
  - Updated various modules to use the new TryTcpUdp helpers and flow-key APIs.

- Performance / size note
  - Embedding Option<FlowKey> in PacketMeta increased Packet and PacketMeta sizes (Packet<TestBuffer> from ~544 B → ~624 B; PacketMeta grew ~80 B). This caused stack pressure in a deep-pipeline test; the test was adjusted to 999 stages. The author intends to explore reducing metadata size.

## Tests
- Numerous unit and integration tests added/updated across nat, flow-filter, config, pipeline, and net to cover directional static NAT, combinations with masquerade and port-forwarding, and ICMP error translations.


nat/src/portfw/packet.rs
```diff
@@ -92,6 +97,11 @@ fn dnat_packet<Buf: PacketBufferMut>(
     new_dst_ip: IpAddr,
     new_dst_port: NonZero<u16>,
 ) -> Result<bool, NatPacketError> {
+    debug_assert!(
+        !packet.meta().is_dst_natted(),
+        "Trying to apply double source NAT to packet!"
```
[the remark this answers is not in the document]
qmonnet: fixed

nat/src/portfw/packet.rs
```diff
@@ -46,6 +46,11 @@ fn snat_packet<Buf: PacketBufferMut>(
     new_src_ip: UnicastIpAddr,
     new_src_port: NonZero<u16>,
 ) -> Result<bool, NatPacketError> {
+    debug_assert!(
+        !packet.meta().is_src_natted(),
+        "Trying to apply double source NAT to packet!"
+    );
```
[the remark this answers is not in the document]
qmonnet: fixed

nat/src/stateless/setup/tables.rs
```diff
@@ -95,24 +97,34 @@ impl PerVniTable {
...
+                debug!("No src mapping found for {addr}:{port:?}, dst_vni: {dst_vni}");
+                None
+            }
             Some((ipaddr, opt_port)) => {
-                debug!("Found src mapping; ipdaddr: {ipaddr} port: {opt_port:?}");
+                if let Ok(unicast_ipaddr) = UnicastIpAddr::try_from(ipaddr) {
+                    debug!("Found src mapping; ipdaddr: {ipaddr} port: {opt_port:?}");
+                    Some((unicast_ipaddr, opt_port))
+                } else {
+                    debug!(
```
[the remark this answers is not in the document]
qmonnet: fixed

qmonnet:
Adding the flow key to the packet metadata increases the size of the metadata and hence the packet. With its 1,000 stages each adding a copy to the stack, `long_dyn_pipeline()` exhausts its stack. Here's the difference:
```
  ┌────────────────────┬────────┬────────┐
  │      size_of       │ before │ after  │
  ├────────────────────┼────────┼────────┤
  │ Packet<TestBuffer> │ 544 B  │ 624 B  │
  ├────────────────────┼────────┼────────┤
  │ PacketMeta         │ 72 B   │ 152 B  │
  ├────────────────────┼────────┼────────┤
  │ Option<FlowKey>    │ —      │ 80 B   │
  └────────────────────┴────────┴────────┘
```

I'll look into reducing the size of the metadata tomorrow.

config/src/external/overlay/tests.rs
```diff
@@ -420,11 +420,7 @@ pub mod test {
 
         // Build overlay object and validate it
         let overlay = Overlay::new(vpc_table, peering_table);
-        assert!(
-            overlay
-                .validate()
-                .is_err_and(|e| e == ConfigError::IncompatibleNatModes("Peering-1".to_owned()))
-        );
+        assert!(overlay.validate().is_ok());
```
daniel-noland, line 423: super tiny nit: this should work with `assert_matches!` as of rust 1.96.0 (very new).  That will give you better error messages than the true / false this returns.

nat/src/icmp_handler/nf.rs
```diff
@@ -20,6 +20,7 @@ use strum::EnumMessage;
 use tracectl::trace_target;
 use tracing::{debug, warn};
 
+use crate::StatelessNat;
```
daniel-noland: Claude spotted this on this commit:


> **safety rationale is inaccurate — fix the comment.** `nat/src/icmp_handler/nf.rs` sets both `set_static_nat_src(true)` and `set_static_nat_dst(true)` then runs `process_packet`, justified by: "we do mark packets as source-NAT-ed or dest-NAT-ed when we process them for masquerade or port forwarding, and the static NAT stage does not re-apply translation… so there's no risk to incur double translations."* But `handle_icmp_error_masquerading` and `handle_icmp_error_port_forwarding` do **not** set `src_natted`/`dst_natted` (only the non-ICMP `snat`/`dnat` paths in `stateful/packet.rs` and `portfw/packet.rs` do — confirmed by grep). So in the ICMP path, *both* `source_nat` and `destination_nat` actually run. The reason it's still safe is the **commit-15 change**: a missing mapping now returns `Ok(false)` (no-op) instead of `NoMappingFound`, and the opposite-end address ranges don't match the wrong direction. The tests pass because of *that*, not because of the natted flags. Please correct the comment to state the real invariant — relying on a documented-but-false guarantee is a maintenance trap for the next person who touches this.

Looking over this, that seems to be true, but I'm not done reviewing yet

daniel-noland (approved):
other than some truly tiny nits this is ready to roll.  I don't even think those nits need to be addressed in this PR, just noted for future.  Nice work.

config/src/external/overlay/tests.rs
qmonnet: Thanks! Better messages would be welcome indeed, at some locations I've been using `assert_eq!()` or `assert!(X.is_ok(), "{X}")` just to get the actual value when the test fails :disappointed:

This pattern is all over the place in tests I've written though, I'll update them in a follow-up PR.

nat/src/icmp_handler/nf.rs
qmonnet: I think Claude is wrong on this one:
- `handle_icmp_error_port_forwarding()` calls `nat_packet()`,
- which in turns calls `dnat_packet()` or `snat_packet()`,
- running `packet.meta_mut().dst_natted(true);` or `packet.meta_mut().src_natted(true);` on success, respectively.

Similarly: `handle_icmp_error_masquerading()` -> `masquerade()` -> `snat()`/`dnat()` from `stateful/packet.rs` setting the flag.

nat/src/portfw/packet.rs
```diff
@@ -46,6 +46,11 @@ fn snat_packet<Buf: PacketBufferMut>(
     new_src_ip: UnicastIpAddr,
     new_src_port: NonZero<u16>,
 ) -> Result<bool, NatPacketError> {
+    debug_assert!(
```
Fredi-raspall, line 49: I would not keep these debug_asserts here.
Whether masquerading can be used with port-forwarding or not is not something that port-forwarding (or masquerading) should enforce. Also, this will bail, but in release, allow the action.
IMO:
   * we should remove this from here (and masquerading)
   * put the constraint (which will hopefully be gone soon) at a validation place where we can err.

nat/src/stateful/nf.rs
```diff
@@ -514,13 +514,12 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for StatefulNat {
...
-            if !packet.is_done()
-                && packet.meta().requires_stateful_nat()
-                && !packet.is_icmp_error()
-                && !packet.meta().is_natted()
+            if !packet.is_done() && packet.meta().requires_stateful_nat() && !packet.is_icmp_error()
             {
                 // Packet should never be marked for NAT and reach this point if it is not overlay
                 debug_assert!(packet.meta().is_overlay());
+                // Packet should never go through both masquerading and port forwarding
+                debug_assert!(!packet.meta().requires_port_forwarding());
```
Fredi-raspall, line 525: Same here

net/src/packet/mod.rs
```diff
@@ -93,6 +95,11 @@ impl<Buf: PacketBufferMut> Packet<Buf> {
         self.headers.set_eth(eth);
     }
 
+    /// Update the flow key of this packet based on its current state
+    pub fn update_flow_key(&mut self) {
```
Fredi-raspall: nit: I would call this set_flow_key and probably forbid setting it twice.
The "update" scared me as a flow key should never change even if we mutate the packet with the corresponding flow entry/info?

nat/src/portfw/flow_state.rs
```diff
@@ -105,23 +105,32 @@ pub(crate) fn build_portfw_flow_keys<Buf: PacketBufferMut>(
...
+
+    let mut key_forward_dnated = current_flow_key;
     key_forward_dnated.data_mut().set_dst_ip(new_dst_ip.inner());
     key_forward_dnated
         .data_mut()
         .set_ip_proto_key(IpProtoKey::from((proto, src_port, new_dst_port)));
     let key_reverse = key_forward_dnated.reverse(Some(dst_vpcd));
 
-    (key_forward, key_reverse)
+    Ok((initial_flow_key, key_reverse))
```
Fredi-raspall, line 130: Why is this now fallible? If this could fail, that means that the validation of the packet that was made to allow the packet to be port-forwarded did not do a good job.

nat/src/portfw/nf.rs
```diff
@@ -105,8 +105,15 @@ impl PortForwarder {
         debug!("Will translate {dst_ip}:{dst_port} -> {new_dst_ip}:{new_dst_port} as per {entry}");
 
         // build keys for the FORWARD and REVERSE flows
-        let (fw_key, rev_key) =
-            build_portfw_flow_keys(packet, new_dst_ip, new_dst_port, entry.dst_vpcd);
+        let Ok((fw_key, rev_key)) =
```
Fredi-raspall, line 108: Related to the above comment.

nat/src/icmp_handler/nf.rs
```diff
@@ -213,6 +214,20 @@ impl IcmpErrorHandler {
...
+        // We first need to update the checksums after the previous NAT changes, or the static NAT
+        // processor will fail to validate checksums and won't proceed.
+        let Some(tables_factory) = &self.tables_factory else {
+            return;
+        };
+        packet.update_checksums();
+        packet.meta_mut().set_static_nat_src(true);
+        packet.meta_mut().set_static_nat_dst(true);
+        let nat_processor =
+            StatelessNat::with_reader("icmp_error_static_nat", tables_factory.handle());
```
Fredi-raspall: I do not like this solution too much. This requires creating a left-right handle per packet, which isn't ideal. Quoting the left-right documentation "Note that this _internally_ takes a lock whenever you call [`ReadHandleFactory::handle`], so you should not expect producing new handles rapidly to scale well."

nat/src/test.rs
```diff
@@ -0,0 +1,627 @@
...
+
+fn pwp(s: &str, start: u16, end: u16) -> PrefixWithOptionalPorts {
+    PrefixWithOptionalPorts::new(s.into(), Some(PortRange::new(start, end).unwrap()))
+}
+
+fn vni(vni: u32) -> Vni {
+    Vni::new_checked(vni).expect("Failed to create VNI")
+}
+
+fn build_packet(
```
Fredi-raspall, line 46: We should not be deferring this PR because of this, but we should really try to use proper, well-formed packets from now on?

Fredi-raspall (changes requested):
I'm mostly fine with the changes @qmonnet, but have one serious concern about the read-handles in the icmp error handling stage. Let's talk about it!

nat/src/portfw/packet.rs
qmonnet: > I would not keep these debug_asserts here. Whether masquerading can be used with port-forwarding or not is not something that port-forwarding (or masquerading) should enforce.

I think there's confusion? Or I don't follow. This is not about supporting port forwarding + masquerading on _different sides_ of a peering. This is a check on the _source_ address here, before we apply source NAT: in what scenario would we want to NAT the source address with two different modes at the same time? It would require configuring both masquerading and port forwarding _on the same end_ of a peering and apply them consecutively.

> Also, this will bail, but in release, allow the action. IMO:

Yes, the idea is to hopefully catch it if something weird happens in CI or tests and the `debug_assert()` triggers.

>  * we should remove this from here (and masquerading)
>  * put the constraint (which will hopefully be gone soon) at a validation place where we can err.

Not sure we can “validate” for that, it's excluded by construction: an expose block can only have up to one NAT mode, per definition. Or are you thinking of something else?

nat/src/stateful/nf.rs
qmonnet: Ack, same reply, marking as resolved to avoid a duplicate discussion.

net/src/packet/mod.rs
qmonnet: Thanks for catching this, it looks like this method is actually unused - some remnant from a previous version. I'll remove it.

nat/src/portfw/flow_state.rs
qmonnet: We've had this discussion in team meetings recently, we agreed `unreachable!()` should be for cases where the code is actually unreachable, not when we assume that the surrounding context makes it impossible to reach that point; and we said we'd fix these progressively moving forward. When we call `build_portfw_flow_keys()` on a packet, there's no guarantee that we're able to extract this flow key, so I made it return a `Result` - even though we know we always call it when we have a packet with a valid flow key at the moment. If we want to make it infallible, we need to make an infallible `FlowKey::from`.

nat/src/portfw/nf.rs
qmonnet: Ack, same discussion, closing this one.

nat/src/test.rs
qmonnet: Right, I reused the code from existing NAT tests. I'll try to find a better example to reuse next time.

nat/src/icmp_handler/nf.rs
qmonnet: Looking again, this commit is silly. There's a much easier way to make the packets go through static NAT in addition with masquerading or port forarding:

```diff
diff --git i/nat/src/stateless/nf.rs w/nat/src/stateless/nf.rs
index a4dac63d3b7d..ba290147e5bc 100644
--- i/nat/src/stateless/nf.rs
+++ w/nat/src/stateless/nf.rs
@@ -372,7 +372,7 @@ impl<Buf: PacketBufferMut> NetworkFunction<Buf> for StatelessNat {
         input.filter_map(|mut packet| {
             if !packet.is_done()
                 && packet.meta().requires_static_nat()
-                && !packet.meta().is_natted()
+                && (!packet.meta().is_natted() || packet.is_icmp_error())
             {
                 // Packet should never be marked for NAT and reach this point if it is not overlay
                 debug_assert!(packet.meta().is_overlay());
```

And get rid of the other commit. I'll change it.

qmonnet:
[This was a simple rebase, to get a clear diff when I push actual changes]

qmonnet:
@Fredi-raspall Can you please look at [the latest changes](https://github.com/githedgehog/dataplane/compare/116940238c1c97791b125630f0161fef02384d26..3cc328fbf1acbf7b2dc687b89df087c7eec237ac)?

nat/src/portfw/packet.rs
Fredi-raspall: Sorry, I meant static NAT.

nat/src/portfw/flow_state.rs
Fredi-raspall: I remember that discussion, but I am not sure if I agree on its applicability (not just here, but in general) or how large that context needs to be. Let me give you another example: in whatever NAT flavor, if we've checked that a packet is IP and could then be potentially NATed, should we attempt to catch each possible failure when fetching, say, the source IP address or just unwrap/unreachable? IMO if a packet has qualified already, any manipulation (e.g. within a stage) of the packet beyond that point should be able to confidently assume that the ip address is there. I'm not super strong at this, but I think that that saves a lot of boilerplate and error-handling noise. .... and if tryIp or the like succeeded and we later can't retrieve an IP address, then we deserve crashing because we have a fundamental bug. In any case, leave it as it is but we should add this to the list of topics to discuss.

nat/src/portfw/packet.rs
qmonnet: I'm still not sure I follow?

flow-filter/src/tests.rs
```diff
@@ -44,25 +44,25 @@ fn vpcd(id: u32) -> VpcDiscriminant {
...
         && !packet.meta().requires_stateful_nat()
-        && !packet.meta().requires_stateless_nat()
+        && !packet.meta().requires_static_nat()
 }
 
 fn needs_no_nat<Buf: PacketBufferMut>(packet: &Packet<Buf>) -> bool {
     !packet.meta().requires_stateful_nat()
-        && !packet.meta().requires_stateless_nat()
+        && !packet.meta().requires_static_nat()
         && !packet.meta().requires_port_forwarding()
```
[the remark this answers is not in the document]
qmonnet: This is only for tests, and in these tests, checking exclusively is what we want.

nat/src/stateless/nf.rs
```diff
@@ -0,0 +1,384 @@
...
+            {
+                self.translate_dst_port(&mut transport, new_dst_port);
+                modified = true;
+            }
+        }
+
+        // ICMP Error messages
+        if Self::is_icmp_inner_translation_candidate(packet) {
+            modified |= Self::translate_icmp_inner_packet_src_if_any(table, packet)?;
+        }
```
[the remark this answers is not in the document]
qmonnet: We should validate the checksum for destination NAT only, although we can't do it here, because `destination_nat()` runs (potentially) after `source_nat()` which may have changed the packet and invalidated the checksum. We should probably move the check to the parent `translate()` function. But this is not related to this PR, I'll address separately.

nat/src/stateless/nf.rs
```diff
@@ -0,0 +1,387 @@
...
+                self.translate_src_port(&mut transport, new_src_port);
+                modified = true;
+            }
+        }
+
+        // ICMP Error messages
+        if Self::is_icmp_inner_translation_candidate(packet) {
+            validate_checksums_icmp(packet).map_err(StatelessNatError::IcmpErrorMsg)?;
+            modified |= Self::translate_icmp_inner_packet_dst_if_any(table, packet, dst_vni)?;
+        }
```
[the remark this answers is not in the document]
qmonnet: Should also be addressed by moving checksum validation to the parent function.

nat/src/stateless/nf.rs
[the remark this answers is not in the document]
qmonnet: Yes, open this issue and assign me to it. Add a joke at the end of the PR description.

nat/src/portfw/flow_state.rs
mvachhar: Looking at this, I can't reason locally that the error case is unreachable!().  What are the invariants we want to assume on entry to this function?  Does every packet have a flow key?  Especially in this case, I'd rather log an error and drop the packet than crash the entire dataplane.  I think crashing in the debug build is appropriate though.

flow-filter/src/lib.rs
```diff
@@ -319,22 +323,51 @@ impl FlowFilter {
...
+            && packet.meta().requires_static_nat())
+        {
+            return;
+        }
+
+        let Ok(flow_key) = FlowKey::try_from(net::flow_key::Uni(&*packet)) else {
+            return;
+        };
+
+        packet.meta_mut().flow_key = Some(Box::new(flow_key));
```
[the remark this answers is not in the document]
qmonnet: Good catch, fixed.

nat/src/portfw/packet.rs
Fredi-raspall: Ok. Let me rephrase. What does that debug assert protect us against and what does it have to do with port forwarding? If I understand the code correctly, you don't want to nat twice the same address. But  that's something that: 1) may happen with portfw + masq or other combos, right? 2) should be better checked somewhere else?

qmonnet: > 1) may happen with portfw + masq or other combos, right?

I don't think so? You'd have portfw + masq on separate ends of a peering, not applying to the same side: so masquerade would translate, say, the source address, and port forward would translate its destination (or the other way around for the reply). I don't think you ever want masquerade and port forwarding to translate the same address?

> 2) should be better checked somewhere else?

Where would be the right place to check whether we're doing a double NAT operation on a given address of the packet regardless of the provided configuration?

Fredi-raspall: Resolving as per our talk.

Fredi-raspall (approved):

qmonnet:
Merging in #1550, which was independently approved, on top of this branch.

PR #1558, opened 2026-05-22
(2-6) qemu cross tests

daniel-noland:
# run cross-compiled tests under qemu-user! 

- A `ci:+cross` or `ci:+cross/full` label exercises the musl/gnu x aarch64/bluefield3 matrix end-to-end instead of stopping at "it linked".
- Put the new qemu-user path and the existing Miri path behind a single `emulated` cfg, so timing-sensitive / filesystem-bound tests can declare "skip me under emulation" once instead of carrying parallel `cfg_attr(miri, ...)` and `cfg_attr(qemu_user, ...)` arms.

~~The `cross` CI job is advisory (job-level `continue-on-error: true`): its results post as a sticky PR comment via a small composite action, rather than turning the workflow red.~~

**Auto-bump PRs ship the `ci:+cross/full` label so upstream regressions surface there too.**

**The goal:** detect cross compile regressions more easily without blocking or slowing down CI in day-to-day operations.  In the future we will likely want to make at least BF3 tests gating for PRs, but we are not quite there yet.  For now I'll settle for getting a big warning when/if we, e.g., start depending on a crate which is x86_64 only.

> [!NOTE]
> We run cross compiled tests in debug mode **only**!  `qemu-user` isn't always able to emulate "higher" instructions for aarch64 (e.g. some NEON instructions) which can and will show up under the release build.  I don't have a real solution to this problem so I'm going to ignore it and take what I can get :shrug: 

## Components

- `scripts/test-runner.sh` -- wired in via `.cargo/config.toml`'s `[target.<linux>]` runners. Fast path: when the host arch matches the target and the host kernel is Linux, `exec` the test directly (zero qemu cost on native runs). Otherwise `exec qemu-<arch>`. Also forwards to `.cargo-miri-wrapped` when `MIRI_SYSROOT` is set, so the runner is a no-op under Miri.
- `nix/profiles.nix` -- gains a `host-arch` parameter, threaded from `default.nix`. For test builds where `arch != host-arch`, the profile adds `--cfg=emulated`. `--check-cfg=cfg(emulated)` is always registered so `cfg_attr(emulated, ...)` sites don't trip `unexpected_cfgs` on native builds.
- `miri.just` -- adds `--cfg=emulated` to `RUSTFLAGS` so the Miri path shares the umbrella.
- `.config/nextest.toml` -- new `cross-qemu` nextest profile with `slow-timeout = { period = "60s", terminate-after = 5 }` and `failure-output = immediate-final` so a hung test in qemu does not hold the run indefinitely.
- `.github/workflows/dev.yml` -- new `cross` job and an updated `summary` step. `continue-on-error: true` at the job level means `needs.cross.result` reports `success` to dependents regardless of outcome; `summary` therefore lists per-cell conclusions via the Actions API and only posts the sticky comment when at least one cell failed.
- `.github/actions/sticky-pr-comment/` -- composite action that create-or-updates a marker-keyed PR comment, with input validation and try/catch so a 5xx from the API does not fail the calling job.

## Notes for reviewers

- `emulated` deliberately conflates qemu-user with Miri. If a test wants "skip on Miri only", keep `cfg_attr(miri, ignore)`; if it wants "skip on qemu only", check the target arch alongside. The umbrella exists for the common case (timing / fs / proc / UDS).
~~- Cross failures on `push` / `merge_group` are not surfaced (sticky comments are PR-only); if main breakage tracking matters, we can add an issue or channel hook later.~~


.github/workflows/dev.yml
```diff
@@ -713,6 +737,22 @@ jobs:
...
           exit 1
+      # Cross is advisory (continue-on-error), so don't exit 1 here --
+      # surface a sticky PR comment instead.
+      - name: "Surface cross failures via sticky PR comment"
+        if: >-
+          ${{
+            github.event_name == 'pull_request'
+            && needs.cross.result != 'success'
+            && needs.cross.result != 'skipped'
+          }}
```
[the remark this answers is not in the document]
daniel-noland: sorted.

scripts/test-runner.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Open Network Fabric Authors
+
+set -euo pipefail
+set -o allexport
+
```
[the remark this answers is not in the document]
daniel-noland: sorted

nix/profiles.nix
```diff
@@ -38,6 +42,9 @@ let
   ];
   common.RUSTFLAGS = [
     "--cfg=tokio_unstable"
+    # Register `emulated` so `#[cfg_attr(emulated, ...)]` never trips
+    # `unexpected_cfgs`; only *set* for is-emulated-test and miri.
+    "--check-cfg=cfg(emulated)"
```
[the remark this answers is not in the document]
daniel-noland: nah, miri is a "well known" cfg, so this is fine

.cargo/config.toml
```diff
@@ -12,13 +12,26 @@ CARGO_LLVM_COV_TARGET_DIR = { value = "target/llvm-cov/build", relative = true,
 CARGO_LLVM_COV_BUILD_DIR = { value = "target/llvm-cov/target", relative = true, force = false }
 
 [build]
-rustflags = ["--cfg=tokio_unstable"]
+# Register `emulated` so cfg_attr sites don't trip unexpected_cfgs natively.
+rustflags = ["--cfg=tokio_unstable", "--check-cfg=cfg(emulated)"]
```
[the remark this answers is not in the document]
daniel-noland: nah, miri is a "well known" cfg, so this is fine

daniel-noland:
see PR #1560 for an illustration of how the ci:+cross/full is now applied to bump jobs

scripts/test-runner.sh
```diff
@@ -0,0 +1,37 @@
...
+declare host_machine
+host_machine="$(uname --machine)"
+declare -r host_machine
+
+declare host_kernel_name
+host_kernel_name="$(uname --kernel-name)"
+host_kernel_name="${host_kernel_name,,}" # convert to lower case
+declare -r host_kernel_name
+
+declare -r target_machine="${1}"
```
qmonnet, line 30: Nit: Do we really need all the `declare`? I'm not sure the size of the script justifies the read-only precaution and they make it quite verbose.

flow-entry/src/flow_table/table.rs
```diff
@@ -567,7 +567,7 @@ mod tests {
         }
 
         #[tokio::test]
-        #[traced_test]
+        #[cfg_attr(not(emulated), traced_test)]
```
qmonnet, line 570: Would it make sense to redefine something like a `#[my_new_traced_test]` config macro that implements `#[cfg_attr(not(emulated), traced_test)]`, to avoid the `cfg_attr()` exclusion everywhere?

Also how do you catch it when someone adds a new `#[traced_test]` without the exclusion on `emulated`, will it crash CI or just make it (silently) very slow? Should there be a semantic grep check for this?

.github/workflows/dev.yml
```diff
@@ -713,6 +720,76 @@ jobs:
...
+        uses: "./.github/actions/sticky-pr-comment"
+        with:
+          marker: "<!-- ci:dev:cross-advisory -->"
+          body: |
+            ## :warning: Cross build advisory: failure detected :warning:
+
+            [Please investigate before merging](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
+
+            Failed legs: ${{ steps.cross_status.outputs.failed_names }}
```
qmonnet: Personally, I'm not a fan of automated comments in the PR's discussion - I prefer to keep that space for reviews and discussions around the PR and it's already hard enough to navigate through the UI like this. Plus comments generate email notifications, which I don't need for every CI failure.

On failure I'd expect the workflow to show red (with the final summary green), which would be enough to detect that something has gone wrong?

nix/pkgs/dplane-plugin/default.nix
```diff
@@ -20,15 +20,6 @@ stdenv.mkDerivation (finalAttrs: {
...
 
-  # workaround: src/hh_dp_msg.c reaches into a glibc-internal anonymous
-  # union name (`.__in6_u.__u6_addr8`) on struct in6_addr.  musl exposes
-  # the POSIX-standard `.s6_addr` member directly without that wrapping
-  # union, so the access fails to compile.
-  # remove once fixed upstream in githedgehog/dplane-plugin.
-  postPatch = ''
-    sed -i 's/\.__in6_u\.__u6_addr8/.s6_addr/g' src/hh_dp_msg.c
-  '';
-
```
qmonnet, line 31: Can you please explain in the commit description why this is OK to remove, now?

.github/workflows/dev.yml
```diff
@@ -489,6 +490,25 @@ jobs:
...
       - *nix-setup
+      - name: "test"
+        if: >-
+          ${{
+            matrix.recipe.args == 'dataplane'
+            && matrix.libc == 'musl'
+            && github.event_name == 'pull_request'
+            && (
+              contains(github.event.pull_request.labels.*.name, 'ci:+cross')
+              || contains(github.event.pull_request.labels.*.name, 'ci:+cross/full')
```
qmonnet: Also: How about `ci:+cross/full` and `ci:+cross/partial`, to make the distinction clear?

Actually, looking again, it's not clear to me where you use `ci:+cross/full` but not `ci:+cross`? The two labels seem to be used at the same locations?

scripts/test-runner.sh
daniel-noland: This I will defend.  bash is an absolutely horrific language and absolutely every single thing you can do to prevent the infinity mirror of bugs inherent to this trash language should be taken.  In every case.  All of them.  Seriously.

flow-entry/src/flow_table/table.rs
daniel-noland: just slow.  And yes, we could add a macro to a test helper crate

.github/workflows/dev.yml
daniel-noland: unless I did something silly, the `/full` actually runs the cross compiled tests where `+cross` just does the compile

.github/workflows/dev.yml
daniel-noland: dropped

.github/workflows/dev.yml
daniel-noland: yeah, I copy pasted wrong.  It is fixed now.

nix/pkgs/dplane-plugin/default.nix
daniel-noland: sure?

scripts/test-runner.sh
qmonnet: I agree regarding bash and bugs. Maybe we should use another language altogether, in that case?

Fine by me if you prefer to keep them.

.github/workflows/dev.yml
qmonnet: Thank you :pray:

flow-entry/src/flow_table/table.rs
```diff
@@ -702,7 +702,7 @@ mod tests {
 
         #[allow(clippy::too_many_lines)]
         #[concurrency::test]
-        #[tracing_test::traced_test]
+        #[cfg_attr(not(emulated), tracing_test::traced_test)]
```
qmonnet, line 705: Nit: it looks like we pull `tracing_test` in the file already.

```suggestion
        #[cfg_attr(not(emulated), traced_test)]
```

config/src/display.rs
```diff
@@ -79,7 +79,7 @@ impl Display for VpcExpose {
             if !nat.as_range.is_empty() {
                 write!(f, "{SEP}       as:")?;
                 nat.as_range.iter().for_each(|pfx| {
-                    let _ = write!(f, " {pfx} proto: {:?} NAT:{}", nat.proto, &nat.config);
+                    let _ = write!(f, " {pfx} proto: {:?} NAT:{}", nat.proto, nat.config);
```
qmonnet, line 82: Do you know why we haven't caught these Clippy reports in CI so far?

.github/workflows/dev.yml
qmonnet: I'd still go with `ci:+cross/partial` (or some variant) rather than just `ci:+cross`, to better mark the distinction with `ci:+cross/full`, but this is not blocking.

flow-entry/src/flow_table/table.rs
qmonnet: If re-introducing `#[traced_test]` would make the CI slow, can you please add a semantic grep check to make sure we don't do that? (Can be in a follow-up).

qmonnet (approved):
My only remaining concern would be the risk to accidentally re-introduce `#[traced_test]` and make the CI slow without noticing, I'd like if we had something to detect and reject this.

Other than this, the PR looks good to me, thanks a lot!

config/src/display.rs
daniel-noland: I actually don't.  It is a conditional compile thing I think.  I was going to investigate but I didn't want to get side tracked

PR #1564, opened 2026-05-24
LSM scratch work

daniel-noland:


daniel-noland:
this did the thing for the moment and is too big to review as such. Closing in favor of adding this logic in a different order

PR #1573, opened 2026-05-31
(9) feat: fixed-size + lookup foundation trait crates

daniel-noland:
Stack (9). Base: `pr/daniel-noland/dpdk-test-eal`.

The two foundational leaf trait crates the match-action / ACL layers build on.
No DPDK, no `unsafe`:

- `feat(fixed-size)`: `FixedSize` trait crate + `impl`s for `net` wire newtypes.
- `feat(lookup)`: `Lookup<K, A>` + `Projection<T>` trait crate.
---

**Review stack** (merge bottom -> top):
- (7) #1555 threading rewrite
- (8) #1572 dpdk test EAL harness
- (9) #1573 fixed-size + lookup
- (10) #1574 match-action
- (11) #1575 acl reference backend
- (12) #1576 acl rte_acl backend
- (13) #1567 cascade

PR #1567, opened 2026-05-26
(13) feat(cascade): in-memory LSM match-action store + real-shape ACL tests

daniel-noland:
Stack (13), this PR (retargeted). Base: `pr/daniel-noland/acl-dpdk`.

In-memory LSM-shaped match-action storage primitive, plus tests that drive it
with a real ACL-shaped consumer.

- `feat(cascade)`: in-memory LSM-shaped match-action storage primitive.
- `test(cascade)`: real-shape ACL consumer + bolero + subscribe tests.
---

**Review stack** (merge bottom -> top):
- (7) #1555 threading rewrite
- (8) #1572 dpdk test EAL harness
- (9) #1573 fixed-size + lookup
- (10) #1574 match-action
- (11) #1575 acl reference backend
- (12) #1576 acl rte_acl backend
- (13) #1567 cascade

PR #1575, opened 2026-05-31
(11) feat(acl): crate scaffolding + software reference backend

daniel-noland:
Stack (11). Base: `pr/daniel-noland/match-action`.

The ACL crate scaffolding and the **software reference backend** -- the readable
oracle that defines ACL semantics and the public API surface.

- `feat(acl)`: crate scaffolding + software reference backend.

The DPDK backend (proven equivalent to this oracle via a differential test)
lands in the next PR.
---

**Review stack** (merge bottom -> top):
- (7) #1555 threading rewrite
- (8) #1572 dpdk test EAL harness
- (9) #1573 fixed-size + lookup
- (10) #1574 match-action
- (11) #1575 acl reference backend
- (12) #1576 acl rte_acl backend
- (13) #1567 cascade

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1143,6 +1143,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "dataplane-acl"
+version = "0.21.0"
+dependencies = [
+ "arrayvec",
+ "dataplane-concurrency",
+ "dataplane-lookup",
+ "dataplane-match-action",
+ "dataplane-net",
+ "thiserror",
+]
+
+[[package]]
 name = "dataplane-args"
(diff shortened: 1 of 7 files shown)


mvachhar (approved):
I feel like I have reviewed a bunch of this before, perhaps a prior branch got merged onto this one?  In any case, the stuff I don't remember looks good to go.  I'm a bit torn on the `dataplane-acl` name since there is a user facing feature for that, which may ore may not end up in this crate (probably not).  But we can always rename it if we want.

mvachhar (approved):

PR #1576, opened 2026-05-31
(12) feat(acl): rte_acl backend + differential-vs-reference + benches

daniel-noland:
Stack (12). Base: `pr/daniel-noland/acl-reference`.

The `rte_acl` backend and everything proving it matches the reference oracle.
This is the largest PR -- the irreducible DPDK lowering complexity:

- `feat(acl/dpdk)`: layout planner + rule lowering.
- `feat(acl/dpdk)`: install + `DpdkAclLookup` + EAL smoke.
- `feat(acl/dpdk)`: `DynDpdkLookup` + shape-fuzz property test.
- `test(acl)`: differential property (reference vs DPDK) + Headers/metadata
  projection demos.
- `feat(acl)`: criterion benchmarks + nix bench-builder.
---

**Review stack** (merge bottom -> top):
- (7) #1555 threading rewrite
- (8) #1572 dpdk test EAL harness
- (9) #1573 fixed-size + lookup
- (10) #1574 match-action
- (11) #1575 acl reference backend
- (12) #1576 acl rte_acl backend
- (13) #1567 cascade

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -287,6 +287,16 @@ miri = true # must ALWAYS work
 # NOTE: concurrency should likely be dependency gated as `cfg(unix)` or `cfg(not(target_arch = "wasm32"))`
 wasm = false # hopeless + pointless
 
+[workspace.metadata.package.concurrency-macros]
+package = "dataplane-concurrency-macros"
+# Proc-macro crate: runs at the host toolchain, not the miri target.
+# Excluded from miri to keep the target dep graph clean.  Without this,
+# `cargo miri nextest` builds the host test binary and then tries to run
+# it through the cargo-miri runner, which fails with a misleading
+# "outdated or invalid JSON; try `cargo clean`".
+miri = false # hopeless + pointless
+wasm = false # hopeless + pointless
+
 [workspace.metadata.package.dataplane]
 package = "dataplane"
 miri = false # hopeless + pointless
@@ -299,6 +309,16 @@ wasm = false # hopeless + pointless
 
 [workspace.metadata.package.dpdk-sys]
 package = "dataplane-dpdk-sys"
+miri = false # hopeless + pointless
+wasm = false # hopeless + pointless
+
+[workspace.metadata.package.dpdk-test-macros]
+package = "dataplane-dpdk-test-macros"
+# Proc-macro crate: runs at the host toolchain, not the miri target.
+# Excluded from miri to keep the target dep graph clean.  Without this,
+# `cargo miri nextest` builds the host test binary and then tries to run
+# it through the cargo-miri runner, which fails with a misleading
+# "outdated or invalid JSON; try `cargo clean`".
 miri = false # hopeless + pointless
 wasm = false # hopeless + pointless
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,6 +127,7 @@ chrono = { version = "0.4.45", default-features = false, features = [] }
 clap = { version = "4.6.1", default-features = true, features = [] }
 color-eyre = { version = "0.6.5", default-features = false, features = [] }
 colored = { version = "3.1.1", default-features = false, features = [] }
+criterion = { version = "0.8.2", default-features = false, features = [] }
 crossbeam-utils = { version = "0.8.21", default-features = false, features = [] }
 dashmap = { version = "6.2.1", default-features = false, features = [] }
 derive_builder = { version = "0.20.2", default-features = false, features = [] }
--- a/acl/Cargo.toml
+++ b/acl/Cargo.toml
@@ -25,7 +25,8 @@ net = { workspace = true, features = [] }
 thiserror = { workspace = true }
 
 [dev-dependencies]
-# Property-test draws for the shape-fuzz oracle.
+# Differential oracle harness: random rulesets + packets compared
+# between the reference and DPDK backends.
 bolero = { workspace = true, features = ["std"] }
 # Override the regular `dpdk` dep to enable its `test` feature
 # (exposes `dpdk::test_support::start_eal` and `dpdk::with_eal`).
@@ -33,3 +34,4 @@ dpdk = { workspace = true, features = ["test"] }
 # Override match-action to also enable `bolero` for property-test
 # generators (`FieldHit` / `FieldMiss`).
 match-action = { workspace = true, features = ["derive", "bolero"] }
+net = { workspace = true, features = ["test_buffer", "builder"] }
--- a/acl/Cargo.toml
+++ b/acl/Cargo.toml
@@ -5,9 +5,20 @@ license.workspace = true
 publish.workspace = true
 version.workspace = true
 
+[features]
+default = []
+# Enables the DPDK `rte_acl` backend for `cascade::Lookup`.  Pulls in
+# the `dpdk` crate (and transitively `dpdk-sys`), so off by default;
+# only the production binary / DPDK-bound consumers turn it on.  The
+# reference backend (always built) carries its own `thiserror`-derived
+# error types, so `thiserror` is unconditional rather than feature-gated
+# under `dpdk`.
+dpdk = ["dep:dpdk"]
+
 [dependencies]
 arrayvec = { workspace = true, default-features = true }
 concurrency = { workspace = true, features = [] }
+dpdk = { workspace = true, optional = true }
 lookup = { workspace = true, features = [] }
 match-action = { workspace = true, features = ["derive"] }
 net = { workspace = true, features = [] }
--- a/acl/Cargo.toml
+++ b/acl/Cargo.toml
@@ -7,12 +7,6 @@ version.workspace = true
 
 [features]
 default = []
-# Enables the DPDK `rte_acl` backend for `cascade::Lookup`.  Pulls in
-# the `dpdk` crate (and transitively `dpdk-sys`), so off by default;
-# only the production binary / DPDK-bound consumers turn it on.  The
-# reference backend (always built) carries its own `thiserror`-derived
-# error types, so `thiserror` is unconditional rather than feature-gated
-# under `dpdk`.
 dpdk = ["dep:dpdk"]
 
 [dependencies]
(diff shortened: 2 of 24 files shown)


Cargo.toml
```diff
@@ -125,6 +125,7 @@ bytes = { version = "1.11.1", default-features = false, features = [] }
 caps = { version = "0.5.6", default-features = false, features = [] }
 chrono = { version = "0.4.44", default-features = false, features = [] }
 clap = { version = "4.6.1", default-features = true, features = [] }
+criterion = { version = "0.8.2", default-features = false, features = ["cargo_bench_support"] }
```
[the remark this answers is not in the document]
daniel-noland: fixed

acl/Cargo.toml
```diff
@@ -5,10 +5,34 @@ license.workspace = true
...
 concurrency = { workspace = true, features = [] }
+dpdk = { workspace = true, optional = true }
 lookup = { workspace = true, features = [] }
 match-action = { workspace = true, features = ["derive"] }
 net = { workspace = true, features = [] }
 thiserror = { workspace = true }
+
+[dev-dependencies]
+bolero = { workspace = true, features = ["std"] }
+criterion = { workspace = true }
```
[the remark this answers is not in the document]
daniel-noland: fixed

acl/src/lib.rs
```diff
@@ -9,18 +9,31 @@
...
-//! feature gate.
-//!
-//! [`lookup::Lookup`]: lookup::Lookup
-//! [`match_action::MatchKey`]: match_action::MatchKey
+#![allow(missing_docs)]
+#![allow(clippy::missing_errors_doc, clippy::missing_panics_doc)]
 
+#[cfg(feature = "dpdk")]
+pub mod dpdk;
 pub mod reference;
```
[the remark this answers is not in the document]
daniel-noland: fixed

acl/src/dpdk/lookup.rs
```diff
@@ -0,0 +1,224 @@
...
+            let buf = pack_user_to_dpdk_stack::<K, STRIDE>(key, &self.layout);
+            bufs.try_push(buf).map_err(|_| BatchError::TooLarge {
+                size: keys.len(),
+                limit: MAX_BATCH,
+            })?;
+        }
+        let mut ptrs: ArrayVec<*const u8, MAX_BATCH> = ArrayVec::new();
+        for buf in &bufs {
+            let _ = ptrs.try_push(buf.as_ptr());
+        }
```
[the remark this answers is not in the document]
daniel-noland: fixed

acl/src/dpdk/lookup.rs
```diff
@@ -0,0 +1,224 @@
...
+        let mut results: ArrayVec<u32, MAX_BATCH> = ArrayVec::new();
+        for _ in 0..keys.len() {
+            let _ = results.try_push(0);
+        }
```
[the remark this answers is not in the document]
daniel-noland: fixed

justfile
```diff
@@ -160,6 +160,11 @@ test package="tests.all" *args: (build (if package == "tests.all" { "tests.all"
     declare -r target="{{ if package == "tests.all" { "tests.all" } else { "tests.pkg." + package } }}"
     cargo nextest run --archive-file results/${target}/*.tar.zst --workspace-remap $(pwd) {{ filter }}
 
+[script]
+bench: (build "benches")
+    {{ _just_debuggable_ }}
+    for bench in ./results/benches/bin/*; do $bench --bench; done
```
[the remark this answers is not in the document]
daniel-noland: fixed

acl/src/dpdk/dyn_table.rs
```diff
@@ -0,0 +1,588 @@
...
+
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.actions.is_empty()
+    }
+    #[must_use]
+    pub fn lookup_bytes(&self, key: &[u8]) -> Option<&A> {
+        let expected = self.user_key_size();
+        assert_eq!(key.len(), expected, "key length must equal user_key_size");
+        let mut dpdk_buf = [0u8; crate::dpdk::lookup::MAX_USER_KEY_BYTES];
```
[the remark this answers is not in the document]
mvachhar: This seems like a reasonable comment, thoughts @daniel-noland?

acl/src/dpdk/layout.rs
```diff
@@ -0,0 +1,567 @@
...
+    if specs[0].size != 1 {
+        return Err(LayoutError::FirstFieldNotOneByte {
+            actual_size: specs[0].size,
+        });
+    }
+
+    let mut defs: ArrayVec<FieldDef, MAX_FIELDS> = ArrayVec::new();
+    let mut user_to_dpdk: ArrayVec<usize, MAX_FIELDS> = ArrayVec::new();
+    let mut user_chunk_counts: ArrayVec<usize, MAX_FIELDS> = ArrayVec::new();
+    for _ in 0..specs.len() {
```
mvachhar: Does it ever make sense to automatically decide the rule order here for performance?  Not needed now of course, or do we want to always make the user choose the order explicitly?

acl/src/dpdk/lookup.rs
```diff
@@ -0,0 +1,224 @@
...
+                limit: MAX_BATCH,
+            })?;
+        }
+        let mut ptrs: ArrayVec<*const u8, MAX_BATCH> = ArrayVec::new();
+        for buf in &bufs {
+            let _ = ptrs.try_push(buf.as_ptr());
+        }
+        let mut results: ArrayVec<u32, MAX_BATCH> = ArrayVec::new();
+        for _ in 0..keys.len() {
+            let _ = results.try_push(0);
```
mvachhar: I saw this paradigm before, surely there is a better way to get keys to be 0 while checking that it is large enough to have all of keys?

acl/src/dpdk/layout.rs
```diff
@@ -0,0 +1,567 @@
...
+        user_to_dpdk
```
mvachhar: This is dirty and took me a bit to figure out that this is just 0 fill, please use a different idiom.  This occurs elsewhere too, so fix there as well.

acl/src/dpdk/layout.rs
```diff
@@ -0,0 +1,567 @@
...
+    push_def(
+        &mut defs,
+        FieldDef::new(field_type_of(specs[0].kind), FieldSize::One, 0, 0, 0),
+    )?;
+
+    let mut offset: u32 = 4;
+    let mut input_index: u8 = 1;
+    let mut group_used: u32 = 0;
+
+    for (user_idx, spec) in specs.iter().enumerate().skip(1) {
```
mvachhar, line 130: The `skip(1)` and prior hardcoded dummy field needs a comment about the DPDK ACL limitation.

acl/src/dpdk/layout.rs
```diff
@@ -0,0 +1,567 @@
...
+                    offset,
+                ),
+            )?;
+            offset += chunk_size;
+            group_used += chunk_size;
+        }
+    }
+    pad_group(&mut defs, &mut offset, &mut group_used, input_index)?;
+
+    let stride =
```
mvachhar, line 176: Can you push the simplification to eliminate the `stride` per our discussion, it is really confusing.

acl/src/dpdk/lookup.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+fn action_for<A>(actions: &[A], user_data: u32) -> Option<&A> {
+    if user_data == 0 {
+        return None;
+    }
+    let idx = usize::try_from(user_data).ok()?.checked_sub(1)?;
+    actions.get(idx)
+}
+fn pack_user_to_dpdk_stack<K, const STRIDE: usize>(key: &K, layout: &DpdkLayout) -> [u8; STRIDE]
```
mvachhar: Can you add a comment that explains why we copy the headers yet again (first during parse) and then for ACL lookup.  Essentially that there is no easy way to guarantee the Rust packing satisfies the ACL library packing rules.

acl/src/dpdk/dyn_table.rs
```diff
@@ -0,0 +1,611 @@
...
+        Self {
+            priority,
+            category_mask,
+            user_fields,
+            action,
+        }
+    }
+}
+#[must_use]
+pub fn predicate_to_chunks(pred: &FieldPredicate, size_bytes: usize) -> AclFieldChunks {
```
mvachhar, line 45: This is really ugly, do we want people using this?  Should we gate this on tests only?

mvachhar: Or maybe have a feature flag that gates it so that people don't accidentally use it.  Also a comment near the flag or in the source to explain why it is gated.

mvachhar (changes requested):

acl/src/dpdk/lookup.rs
daniel-noland: this bit got removed anyway.

acl/src/dpdk/dyn_table.rs
daniel-noland: actually I can't.  I could make it private, but one of the other structs requires a Box<dyn> of some of this

acl/src/dpdk/layout.rs
daniel-noland: fixed

acl/src/dpdk/lookup.rs
daniel-noland: fixed

acl/src/dpdk/layout.rs
daniel-noland: fixed

acl/src/dpdk/layout.rs
daniel-noland: done

acl/src/dpdk/layout.rs
daniel-noland: We could make it a rule that the one listed first in the proc macro gets checked first.  That is likely the best I can do

acl/src/dpdk/dyn_table.rs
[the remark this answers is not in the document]
daniel-noland: I actually don't strictly agree.  I think None here is actually confusing, and in the safe-er form of this with the proc macro you should not need this call.  I can look into making it more private tho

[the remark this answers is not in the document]
daniel-noland: hmmm, it actually adds a whole feature gate to make this private because the prop tests can only use the public part.  I can look into moving the prop tests into this module, but I think we should do that in another PR.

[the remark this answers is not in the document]
mvachhar: I agree, returning None here is bad, if we want to return instead of assert it should be Result

mvachhar (approved):

acl/src/dpdk/layout.rs
mvachhar: No need to do it now.

acl/src/dpdk/lookup.rs
mvachhar: Please fix this before merge.

PR #1587, opened 2026-06-10
NAT: ICMP Error checksums validation clean-up

### LINKED ISSUE #1586: stateless NAT: validate ICMP checksums before destination-side inner translation
## Background

In `nat/src/stateless/nf.rs`, `source_nat()` guards ICMP-error inner-packet rewrites with `validate_checksums_icmp()` before calling `translate_icmp_inner_packet_dst_if_any()`. However, `destination_nat()` currently skips that validation step when calling `translate_icmp_inner_packet_src_if_any()`.

The validation cannot simply be added inside `destination_nat()` because `destination_nat()` may run *after* `source_nat()` has already rewritten the packet, which would invalidate the checksum and cause the validation to fail spuriously.

## Proposed fix

Move the `validate_checksums_icmp()` call into the parent `translate()` function, before either NAT direction runs. That way the checksum is validated once against the original packet, and both `source_nat()` and `destination_nat()` can safely perform their respective ICMP inner-packet translations.

## References

- PR: https://github.com/githedgehog/dataplane/pull/1548
- Review comment: https://github.com/githedgehog/dataplane/pull/1548#discussion_r3374773293

Raised by @qmonnet.

qmonnet:
Some clean-ups for checksum validation for ICMP Error messages.

Please refer to individual commits for details.

Fixes: #1586

### DIFF
--- a/nat/src/icmp_handler/icmp_error_msg.rs
+++ b/nat/src/icmp_handler/icmp_error_msg.rs
@@ -7,32 +7,21 @@
 use crate::NatPort;
 use crate::NatTranslationData;
 use net::buffer::PacketBufferMut;
-use net::checksum::{Checksum, ChecksumError};
+use net::checksum::Checksum;
 use net::headers::{
-    EmbeddedTransport, TryEmbeddedHeaders, TryEmbeddedHeadersMut, TryEmbeddedTransportMut,
-    TryIcmpAny, TryInnerIpMut, TryInnerIpv4, TryIp,
+    EmbeddedTransport, TryEmbeddedHeadersMut, TryEmbeddedTransportMut, TryInnerIpMut,
 };
 use net::icmp_any::TruncatedIcmpAny;
-use net::icmp_any::{IcmpAnyChecksumErrorPlaceholder, IcmpAnyChecksumPayload};
-use net::ipv4::Ipv4;
 use net::packet::{DoneReason, Packet};
 use std::net::IpAddr;
 use std::num::NonZero;
 
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 pub enum IcmpErrorMsgError {
-    #[error("failure to get IP header")]
-    NoIpHeader,
-    #[error("not an ICMP packet")]
-    NotIcmp,
     #[error("failure to get embedded headers")]
     NoEmbeddedHeaders,
     #[error("failure to get inner IP header")]
     NoInnerIpHeader,
-    #[error("failed to validate ICMP checksum")]
-    BadChecksumIcmp(ChecksumError<IcmpAnyChecksumErrorPlaceholder>),
-    #[error("failed to validate ICMP inner IP checksum")]
-    BadChecksumInnerIpv4(ChecksumError<Ipv4>),
     #[error("invalid transport-layer port {0}")]
     InvalidPort(u16),
     #[error("invalid IP version")]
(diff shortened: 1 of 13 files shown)


nat/src/static_nat/nf.rs
```diff
@@ -281,11 +265,22 @@ impl StatelessNat {
             return Err(StatelessNatError::MissingTable(src_vni));
         };
 
+        let icmp_err = packet.is_icmp_error_with_embedded_packet();
+
+        // Before NAT-ing ICMP Error packets, we should validate the following checksums:
+        //
+        // - The outer IP header checksum, in the case of IPv4
+        // - The ICMP header checksum
+        // - The inner IPv4 header checksum, if any
```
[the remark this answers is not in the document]
qmonnet: right, fixed

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -37,68 +39,69 @@ pub enum IcmpErrorMsgError {
...
+/// Validate the checksums for an ICMP packet.
+///
+/// From REQ-3 from RFC 5508, "NAT Behavioral Requirements for ICMP".
+/// NAT should silently discard packet if:
+///
+///    - ICMP checksum fails to validate
+///    - ICMP checksum for embedded packet fails to validate
+///
+/// This function is valid for any ICMP packet, whether it embeds a packet fragment or not. However,
+/// we expect ICMP Error messages to always include a fragment of the offending packet.
```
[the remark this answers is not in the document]
qmonnet: fixed as well

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -37,68 +39,70 @@ pub enum IcmpErrorMsgError {
...
             }
         }
     }
 }
 
-// # Return
-//
-// * `Ok(())` if checksums are valid and we can translate the inner packet
-// * An error if we fail to validate relevant checksums and packet should be dropped
+impl From<IcmpErrorMsgError> for DoneReason {
```
Fredi-raspall: Do we need this? I mean couldn't we just use the previous From impl passing a reference?

nat/src/icmp_handler/icmp_error_msg.rs
```diff
@@ -37,68 +39,70 @@ pub enum IcmpErrorMsgError {
...
+///
+/// From REQ-3 from RFC 5508, "NAT Behavioral Requirements for ICMP".
+/// NAT should silently discard packet if:
+///
+///    - ICMP checksum fails to validate
+///    - ICMP checksum for embedded packet fails to validate
+///
+/// This function is valid for any ICMP packet, whether it embeds a packet fragment or not. However,
+/// we expect ICMP Error messages to always include a fragment of the offending packet.
 pub(crate) fn validate_checksums_icmp<Buf: PacketBufferMut>(
```
Fredi-raspall: Here, you're getting a mut ref to a Packet.
I think the code could be much simplified by just annotating the packet meta with a done reason if s/t did not work and we'd avoid the need to translate an error type into a DoneReason which is costly to maintain?

Fredi-raspall (approved):
I just have two minor comments, which are not blocking.

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: True, I'll fix

nat/src/icmp_handler/icmp_error_msg.rs
qmonnet: > Here, you're getting a mut ref to a Packet.

Just a ref, it's not mutable? Or do you mean that I could get a mutable one?

> I think the code could be much simplified by just annotating the packet meta with a done reason if s/t did not work and we'd avoid the need to translate an error type into a DoneReason which is costly to maintain?

Maybe, it's true we'd avoid the conversion and it would look cleaner. But I'd keep the current version, because:

- This function is going away in #1591 
- More importantly, we've said before we should try to limit the size of the `PacketMeta` object, so I'll avoid adding a new field if I can.

Thanks a lot for the review!

PR #1589, opened 2026-06-10
Upgrade npins

qmonnet:
We had the following error for dependency updates:

```console
$ bash scripts/bump.sh
++ dirname scripts/bump.sh
+ pushd scripts/..
/var/home/qmo/dev/dataplane /var/home/qmo/dev/dataplane
+ npins update
Error: Failed to deserialize sources.json

Caused by:
    Version 7 is too old, you need to run upgrade
```

To address it, we simply ran the following command and committed the changes:

```console
$ npins upgrade
```

@daniel-noland Can you please double-check this is OK?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -652,9 +652,9 @@ dependencies = [
 
 [[package]]
 name = "cbindgen"
-version = "0.29.3"
+version = "0.29.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c95537b45400390270fae69ac098d057c8f5399001cde9d04f700c105ddfff2d"
+checksum = "2ecb53484c9c167ba674026b656d8a27d7657a58e6066aa902bfb1a4aa00ae20"
 dependencies = [
  "heck",
  "indexmap",
@@ -1920,7 +1920,7 @@ dependencies = [
 [[package]]
 name = "dplane-rpc"
 version = "1.1.2"
-source = "git+https://github.com/githedgehog/dplane-rpc.git?branch=pr%2Fdaniel-noland%2Fbumps#7bb6b46fd78051a8e532a70b16732ad99188721e"
+source = "git+https://github.com/githedgehog/dplane-rpc.git?branch=pr%2Fdaniel-noland%2Fbumps#6c84b7aff35abb4e94fbb0d09870a0b4a2322913"
 dependencies = [
  "bytes",
  "cbindgen",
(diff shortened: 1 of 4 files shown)


daniel-noland:
@qmonnet 

yup, this is correct.  That said, hold off on the merge for a few so I can bump the cargo packages with it.  If we are going to take cache churn we may as well concentrate it to minimize CI load.  Should only take a few minutes anyway.  I'll just target a bump run against this branch.

daniel-noland (approved):
Yup, LGTM

qmonnet:
Ugh, looks like the cbindgen update breaks things

qmonnet:
cargo update (commit 5d94383a829a) bumped cbindgen 0.29.3 → 0.29.4. The
    0.29.4 release added "support for constant enums and arrays," which changes
    the C bindings it generates — even though write_to_file itself is
    unchanged.

    The key is how cbindgen decides whether to write. Its write_to_file is:

    let mut new_file_contents = Vec::new();
    self.write(&mut new_file_contents);

    let mut old_file_contents = Vec::new();
    {
        let mut old_file = File::open(&path).unwrap();   // file already exists -> OK
        old_file.read_to_end(&mut old_file_contents).unwrap();
    }

    if old_file_contents != new_file_contents {
        let mut new_file = File::create(&path).unwrap();  // <-- line 263: PermissionDenied
        new_file.write_all(&new_file_contents).unwrap();
        ...
    }

    So in dplane-rpc's build.rs, cbindgen generates a header into the crate's
    own source tree (a header that's committed to the dplane-rpc repo). The
    flow:

    - With 0.29.3: generated output matched the committed header byte-for-byte
      → old == new → it skips the write entirely → build succeeds, even though
      the source tree is read-only.
    - With 0.29.4: the new const-enum/array codegen makes the generated output
      differ from the committed header → old != new → it tries
      File::create(&path) to overwrite it → in the nix sandbox the source tree
      is read-only, so you get Os { code: 13, PermissionDenied } → panic at
      bindings.rs:263.

    It's not really a nix problem — it's that dplane-rpc writes a generated
    artifact back into its source dir at build time, which is inherently
    fragile, and any cbindgen output change trips it. The nix read-only sandbox
    just turns the silent overwrite into a hard failure.

    Proper fix (upstream in dplane-rpc): regenerate and commit the header with
    cbindgen 0.29.4 on the pr/daniel-noland/bumps branch so the on-disk file
    matches again — or better, have its build.rs emit into OUT_DIR instead of
    the source tree so output churn never collides with a read-only build.


(Claude)

So basically there's a header we need to re-generate.

qmonnet:
@daniel-noland I opened https://github.com/githedgehog/dplane-rpc/pull/33 in an attempt to solve this. This is probably not the right long-term solution though, but should unblock this particular PR (I think?).

daniel-noland:
> @daniel-noland I opened [githedgehog/dplane-rpc#33](https://github.com/githedgehog/dplane-rpc/pull/33) in an attempt to solve this. This is probably not the right long-term solution though, but should unblock this particular PR (I think?).

Yeah, we need to reorg the way that works because this is a problem waiting to happen.  I'll approve for now but we should either abosorb that repo into dataplane / nix infra or at least stop committing the generated code because that is a pain.

qmonnet:
Rebased without changes to the commits in the PR, I expect CI to pick the new version of the dplane-rpc branch and to pass

qmonnet:
In fact, I needed to re-run `cargo update` to catch the change in the Cargo.lock.

PR #1596, opened 2026-06-15
Set Tracing Rate Limiter config via CRD

sergeymatov:
Depends on https://github.com/githedgehog/fabric/pull/1472

### DIFF
--- a/args/src/lib.rs
+++ b/args/src/lib.rs
@@ -507,7 +507,8 @@ pub struct TracingConfigSection {
     pub show: TracingShowSection,
     /// Tracing configuration string (e.g., "default=info,nat=debug")
     pub config: Option<String>, // TODO: stronger typing on this config?
-    /// Optional rate limit for lower-severity tracing output
+    /// Rate limit for lower-severity tracing output; throttling is always on,
+    /// so an unset CLI flag falls back to the default (50:5)
     pub rate_limit: Option<TracingRateLimit>,
 }
 
@@ -1317,9 +1318,9 @@ E.g. default=error,all=info,nat=debug will set the default target to error, and 
         long,
         value_name = "BURST:REPLENISH_PER_SECOND",
         value_parser=TracingRateLimit::from_str,
-        help = "Optional rate limit for lower-severity tracing output. Syntax: BURST:REPLENISH_PER_SECOND.
+        help = "Rate limit for lower-severity tracing output. Syntax: BURST:REPLENISH_PER_SECOND.
 Example: --tracing-rate-limit 50:5 allows bursts up to 50 repeated messages and replenishes 5 messages per second.
-If omitted, tracing output is not rate-limited."
+Throttling is always on: if omitted, the default rate limit of 50:5 is applied. (DEBUG is exempt from throttling.)"
     )]
     tracing_rate_limit: Option<TracingRateLimit>,
 
--- a/config/src/converters/k8s/config/tracecfg.rs
+++ b/config/src/converters/k8s/config/tracecfg.rs
@@ -39,19 +39,16 @@ impl TryFrom<&GatewayAgentGatewayLogs> for TracingConfig {
                 config.add_tag(tag, level);
             }
         }
+        // Absent rateLimit => disabled (config.rate_limit stays None). A present
+        // block enables it; an empty `{}` / zero field resolves to the default.
         if let Some(rl) = logs.rate_limit.as_ref() {
-            let burst = rl.burst.filter(|&b| b > 0).ok_or_else(|| {
-                FromK8sConversionError::InvalidData("rate-limit burst must be > 0".to_string())
-            })?;
-            let replenish_per_second =
-                rl.replenish_per_second.filter(|&r| r > 0).ok_or_else(|| {
-                    FromK8sConversionError::InvalidData(
-                        "rate-limit replenishPerSecond must be > 0".to_string(),
-                    )
-                })?;
+            let defaults = TracingRateLimit::default();
             config.set_rate_limit(TracingRateLimit {
-                burst,
-                replenish_per_second,
+                burst: rl.burst.filter(|&b| b > 0).unwrap_or(defaults.burst),
+                replenish_per_second: rl
+                    .replenish_per_second
+                    .filter(|&r| r > 0)
+                    .unwrap_or(defaults.replenish_per_second),
             });
         }
         Ok(config)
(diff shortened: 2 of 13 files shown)


sergeymatov:
Adding `don't merge` label because Fabric-side PR is not yet merged and there is Fabric version commit bump is missing.

tracectl/src/control.rs
```diff
@@ -377,6 +377,71 @@ where
...
+/// `Arc<ArcSwap<Option<TracingRateLimitLayer>>>`.
+///
+/// `None` means "no throttling": every event passes. [`reload`](Self::reload)
+/// builds a brand-new layer (with fresh token buckets) from the new config and
+/// swaps it in atomically, or stores `None` to disable throttling entirely.
+/// A `Some(config)` that fails to build falls back to `None` (unthrottled),
+/// mirroring the init-time behavior of [`TracingControl::build_rate_limit_layer`].
+#[derive(Clone)]
+struct AtomicThrottle {
+    inner: Arc<ArcSwap<Option<TracingRateLimitLayer>>>,
```
Fredi-raspall: I think that ArcSwap should be replaced by `Slot` in the concurrency crate.

config/src/converters/k8s/config/tracecfg.rs
```diff
@@ -36,6 +39,21 @@ impl TryFrom<&GatewayAgentGatewayLogs> for TracingConfig {
                 config.add_tag(tag, level);
             }
         }
+        if let Some(rl) = logs.rate_limit.as_ref() {
```
Fredi-raspall, line 44: The TracingConfig object instantiated a few lines above will have a default `TracingRateLimit`.
So, does this mean that we have no way to disable the throttling?

config/src/internal/device/tracecfg.rs
```diff
@@ -19,16 +19,42 @@ use tracing::metadata::LevelFilter;
...
+    }
+}
+
 #[derive(Clone, Debug)]
 pub struct TracingConfig {
     pub default: LevelFilter,
     pub tags: OrderMap<String, LevelFilter>,
+    /// Log rate limiter — always present; defaults to
+    /// [`TracingRateLimit::default`] when a config omits it.
+    pub rate_limit: TracingRateLimit,
```
Fredi-raspall: Related to the prior question, I wonder if this should be Option<TracingRateLimit>?

Fredi-raspall (changes requested):
Looks mostly good to me  @sergeymatov. Thx !

config/src/converters/k8s/config/tracecfg.rs
sergeymatov: It's true, defined by the design

config/src/internal/device/tracecfg.rs
sergeymatov: TracingRateLimit is mandatory, defaults are set via fabric

dataplane/src/runtime.rs
```diff
@@ -48,12 +48,13 @@ fn init_name(args: &CmdArgs) -> Result<String, String> {
     }
 }
 fn init_logging(args: &CmdArgs, gwname: &str) {
-    TracingControl::init_with_rate_limit(args.tracing_rate_limit().map(|rate_limit| {
-        TracingRateLimitConfig {
+    TracingControl::init_with_rate_limit(Some(args.tracing_rate_limit().map_or_else(
+        TracingRateLimitConfig::default,
+        |rate_limit| TracingRateLimitConfig {
             burst: rate_limit.burst,
             replenish_per_second: rate_limit.replenish_per_second,
```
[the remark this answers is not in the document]
sergeymatov: It is designed to act like this

config/src/internal/device/tracecfg.rs
Fredi-raspall: You mean mandatory in the API?

Fredi-raspall: IMO, ideally, even if defaulted by the API if absent, there would be a way to actually disable the throttling.
If for anything, that would allow us to measure the penalty in performance when enabled / disabled.

sergeymatov: We agreed that Dataplane by default must have throttler, maybe it's worth to disable it as well. Current API won't allow you to do this

args/src/lib.rs
```diff
@@ -1317,9 +1318,9 @@ E.g. default=error,all=info,nat=debug will set the default target to error, and
         long,
         value_name = "BURST:REPLENISH_PER_SECOND",
         value_parser=TracingRateLimit::from_str,
-        help = "Optional rate limit for lower-severity tracing output. Syntax: BURST:REPLENISH_PER_SECOND.
+        help = "Rate limit for lower-severity tracing output. Syntax: BURST:REPLENISH_PER_SECOND.
 Example: --tracing-rate-limit 50:5 allows bursts up to 50 repeated messages and replenishes 5 messages per second.
-If omitted, tracing output is not rate-limited."
+Throttling is always on: if omitted, the default rate limit of 50:5 is applied. (DEBUG is exempt from throttling.)"
```
Fredi-raspall, line 1323: I believe we no longer have a default?

Fredi-raspall: Oh, apparently we do.

config/src/internal/device/tracecfg.rs
```diff
@@ -19,16 +19,43 @@ use tracing::metadata::LevelFilter;
...
 const DEFAULT_DEFAULT_LOGLEVEL: LevelFilter = LevelFilter::INFO;
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TracingRateLimit {
+    pub burst: u32,
+    pub replenish_per_second: u32,
+}
+impl Default for TracingRateLimit {
+    fn default() -> Self {
+        #[cfg(unix)]
```
Fredi-raspall, line 29: what is the purpose of this `#[cfg()` ?

mgmt/src/processor/proc.rs
```diff
@@ -557,6 +557,11 @@ fn apply_tracing_config(tracing: &Option<TracingConfig>) -> ConfigResult {
             .iter()
             .map(|(tag, level)| (tag.as_str(), *level)),
     )?;
+    // `None` (no rateLimit in config) disables throttling; `Some` enables it.
+    get_trace_ctl().reload_rate_limit(tracing.rate_limit.map(|rl| TracingRateLimitConfig {
```
Fredi-raspall, line 561: I am confused about this map().
Why do we have a `TracingRateLimit` and a `TracingRateLimitConfig`, both  identical?
Couldn't we just have `TracingRateLimit`?

Fredi-raspall (approved):
I have some questions, but they are non-blocking.
Thanks @sergeymatov !

PR #1597, opened 2026-06-15
flow-filter: Remove useless flow info check in flow-filter

qmonnet:
In flow-filter, when the flow-filter table lookup returns multiple matches for a packet, we look at the flow information, if any, to figure out whether we know of an open connection for this packet, and whether this connection is associated to port forwarding or masquerade.

While this logic makes sense, it is no longer used in practice: if we have a valid flow for the packet, then `bypass_with_flow_info()` should succeed, and we bifurcate early from the main portion of the flow-filter processing. If that function fails, we won't have a valid flow info to look at.

So we can simplify that step in case of multiple matches, and immediately look at L4 protocols and flow directions in `deal_with_multiple_matches()`.

### DIFF
--- a/flow-filter/src/lib.rs
+++ b/flow-filter/src/lib.rs
@@ -143,29 +143,6 @@ impl FlowFilter {
         true
     }
 
-    /// Attempt to determine destination VPC from packet's flow-info.
-    fn check_packet_flow_info<Buf: PacketBufferMut>(
-        &self,
-        packet: &mut Packet<Buf>,
-    ) -> Result<Option<VpcDiscriminant>, DoneReason> {
-        let nfi = &self.name;
-
-        let Some(flow_info) = packet.active_flow_info() else {
-            debug!("Packet does not contain any active flow-info");
-            return Ok(None);
-        };
-
-        let vpcd = flow_info.get_dst_vpcd();
-
-        let Some(dst_vpcd) = vpcd else {
-            debug!("{nfi}: No VPC discriminant found, dropping packet");
-            return Err(DoneReason::Unroutable);
-        };
-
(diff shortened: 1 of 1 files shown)


qmonnet:
(Note: Even though this code is supposed to go away soon, I saw no reason to skip the simplification - it's always good to get a simpler flow-filter, if only to look at what it currently does and compare with any future replacement)

flow-filter/src/lib.rs
```diff
@@ -436,31 +413,9 @@ impl FlowFilter {
...
-                        self.deal_with_multiple_matches(packet, &data_set, &tuple)
-                    }
-                    Err(reason) => {
-                        debug!("Will drop packet. Reason: {reason}");
-                        packet.invalidate_flows();
-                        packet.done(reason);
-                        return;
-                    }
-                }
+                self.deal_with_multiple_matches(packet, &data_set, &tuple)
```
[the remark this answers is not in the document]
qmonnet: If the `genid` is too old, it's just as well to not consider the flow as valid.

flow-filter/src/lib.rs
```diff
@@ -436,31 +413,9 @@ impl FlowFilter {
             }
             Some(VpcdLookupResult::MultipleMatches(data_set)) => {
```
Fredi-raspall, line 414: As discussed, this logic should also go away.

Fredi-raspall (approved):

flow-filter/src/lib.rs
qmonnet: Agreed. I may remove it, or we'll switch to the new implementation, directly.

Thanks for the review! :pray:

qmonnet:
Thanks! I'll rebase and merge after #1598 is in.

PR #1609, opened 2026-06-29
chore: document required nix option

edipascale:
following just the installation notes in the README I was not able to run 'just build' until someone pointed out to me the required option in nix.conf, so let's make sure future developers will not hit the same issue


Fredi-raspall (approved):
LGTM, thanks @edipascale !

qmonnet (approved):

PR #1612, opened 2026-07-01
ci security

daniel-noland:
Spruce up CI security in response to more opengrep findings.

Specifically, opengrep now wants pinned github actions.

I added pinact to manage the pinning and zizmor to address other security issues in the CI while we are at it.


.pinact.yaml
```diff
@@ -0,0 +1,9 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/suzuki-shunsuke/pinact/refs/heads/main/json-schema/pinact.json
+# pinact - https://github.com/suzuki-shunsuke/pinact
+version: 3
+
+rules:
+    - ignore: true
+      conditions:
+          - expr: |
+                ActionName == "githedgehog/fabricator/.github/workflows/run-vlab.yaml"
```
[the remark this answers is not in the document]
daniel-noland: no, we control that action internally and don't need to pin it

.github/workflows/bump.yml
```diff
@@ -21,10 +21,7 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
-  contents: "write"
-  packages: "read"
-  id-token: "write"
-  pull-requests: "write"
+  contents: "read" # actions/checkout on this private repo; git push + PR go through the scoped app token below
```
qmonnet: Not sure I understand the comment, what “private” repo are we talking about? (Same question in dev.yml)

.github/workflows/lint-cargo-fmt.yml
```diff
@@ -1,50 +0,0 @@
...
-concurrency:
-  group: "${{ github.workflow }}:${{ github.event.pull_request.number || github.event.after || github.event.merge_group && github.run_id }}"
-  cancel-in-progress: true
-
-permissions:
-  contents: "read"
-
-jobs:
-  format-check:
-    name: "Check formatting for Rust code"
```
qmonnet, line 22: This one seems to be removed rather than moved, this is because we already run it in the `check` job for dev.yml, right?

.github/workflows/dev.yml
```diff
@@ -225,6 +225,12 @@ jobs:
         with:
           recipe: "check-dependencies"
 
+      - name: "commitlint"
```
qmonnet, line 241: I don't understand the motivation for this new lint. But from the commit description, I understand you plan to discuss it.

qmonnet (commented):
Looks good overall (with minor comments inline below).

My main concern is the pinning of the GitHub Actions versions. I understand why it makes it better reproducible to pin the Actions, although I don't think we've been bitten by sliding tags so far, so is this something you want to do to harden the posture, or is it to make opengrep happy?

The thing that worries me is that, unless I'm mistaken, Dependabot will start opening PRs to update Actions hashes for every single patch release, and it's going to be rather noisy (and annoying). Do we have a way to limit that? If we could have all Dependabot non-urgent updates in a single, weekly PR, for example, it would be nice, but I don't know if it supports this.

justfile
```diff
@@ -460,6 +460,95 @@ depgraph:
...
+
+[script]
+bump-cargo-deps:
+    {{ _just_debuggable_ }}
+    declare BASE
+    BASE="$(git rev-parse HEAD)"
+    declare -r BASE
+
+    # Run "cargo update"
+    echo "::notice::Running cargo update"
```
qmonnet, line 479: Note: we could use a `print_notice()` wrapper to print `::notice::` in GitHub Actions and format differently outside of it, but it's absolutely fine to keep that for a follow-up or even to keep it like this, I don't expect us to run this locally very often anyway.

justfile
```diff
@@ -460,6 +460,95 @@ depgraph:
...
+    # Create Pull Request description
+    declare upgrade_log
+    upgrade_log="$(mktemp)"
+    declare -r upgrade_log
+    if [[ "${nb_upgrades}" -ge 1 ]]; then
+        {
+            echo "### :rocket: Upgrades available";
+            echo ""
+            echo "| name | old | req | compatible | latest |";
+            echo "|------|-----|-----|------------|--------|";
```
qmonnet, line 532: Nit: The columns size won't be automatically adjusted, so I'd pick a larger width for the "name" column; then I think with awk you have `printf` available and that you can pad the fields to keep something nicely aligned.

Doesn't matter much, though.

qmonnet: In fact, why not reuse the same logic as we had previously?

```bash
          # Create Pull Request description
          echo "### :rocket: Upgrades available";
          if [[ "${nb_upgrades}" -ge 1 ]]; then
              {
                  echo ""
                  echo "\`\`\`"
                  cat "${upgrade_output}"
                  echo "\`\`\`"
              } > "${upgrade_log}"
          fi
```

.github/workflows/bump.yml
```diff
@@ -29,13 +29,14 @@ jobs:
...
       USER: "runner"
+      JUST_VARS: ""
     steps:
       # Use a GitHub App token so that the generated PR can trigger CI
       - name: "Generate GitHub App token"
         id: "app-token"
         uses: "actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1" # v3.2.0
         with:
-          app-id: "${{ secrets.DP_APP_ID }}"
+          client-id: "${{ secrets.DP_APP_ID }}"
```
qmonnet, line 39: Just double-checking, is this intentional?

justfile
```diff
@@ -392,6 +460,95 @@ depgraph:
       | sed 's/dataplane-//g' \
       | dot -Grankdir=TD -Gsplines=polyline -Granksep=1.5 -Tsvg > workspace-deps.svg
 
+[script]
+bump-actions:
+    {{ _just_debuggable_ }}
+    pinact run --update
```
qmonnet, line 466: Looks good, but shouldn't we call this from the bump.yml workflow to update GitHub Actions on a periodical basis, before the cooldown for Dependabot ends and it starts opening its own PRs? I don't see `bump-actions` called in CI at the moment.

qmonnet (commented):
Looks good, thank you.

My two main questions:

- Can you double-check (and ideally comment on) the change from `app-id` to `client-id` in bump.yml? I note this makes it inconsistent with dev.yml and version-bump.yml, by the way.

- It seems to me that we want to call `just bump-actions` as part of bump.yml to avoid Dependabot opening too many PRs for GitHub Actions updates?

.github/workflows/dev.yml
qmonnet: Discussed during the meeting - this is to make it easier to run the linter locally from the nix-shell, without pulling the Node.js ecosystem.

Should ideally be mentioned in the commit log.

.github/workflows/bump.yml
daniel-noland: yup.  app-id is deprecated

qmonnet: OK thanks, would you mind fixing the other workflows in this case, please?

daniel-noland:
closed in favor of #1622

justfile
daniel-noland: we can't because that action does not have the authority to edit our actions.

.github/workflows/bump.yml
daniel-noland: done

justfile
daniel-noland: we can go back to that.  I just thought the output should be in markdown.  That output doesn't go in the git log, just the rendered github summary

justfile
daniel-noland: yeah, I considered it but I don't think it is worth the effort

.github/workflows/lint-cargo-fmt.yml
daniel-noland: no, it is just in the same spot as the rest of the cargo oriented commands

.github/workflows/bump.yml
daniel-noland: fixed

.github/workflows/dev.yml
daniel-noland: this is just to let us run the commit lint locally.  Will adjust log

qmonnet (approved):
Looks all good to me, thanks a lot for addressing the feedback.

Let's get this merged, as the crossbeam-epoch bump to 0.9.20 in Cargo.lock, and the memory restriction fixes in the latest commits, are currently required to unblock the CI.

PR #1618, opened 2026-07-06
User ACLs: Add k8s converter from Fabric CRD, and ACLs validation

### LINKED ISSUE #1592: Config: K8s converter for ACLs
Once the CRD for ACLs is supported in the agent, we want to convert it into a configuration object in the dataplane.

### LINKED ISSUE #1593: ACLs validation
Make sure the ACL config we obtain from Kubernetes is valid

Depends on #1592

edipascale:
Translate the newly-added spec.peerings[].acl CRD field into the internal Acl/AclRule/AclPattern model and attach it to VpcPeering. Also fixes two kopium-generated struct literals (GatewayAgentPeerings, GatewayAgentGatewayLogs) in the k8s-intf bolero generators that no longer compiled once the CRD gained the acl and rateLimit fields.

this PR will fail until https://github.com/githedgehog/fabric/pull/1498 is merged; it was built and tested locally by building the CRDs from the feature branch.
it also does not concern itself with validation (past what is strictly required for the conversion) or the broader integration of ACLs within dataplane.

[EDIT: The Fabric changes have been merged in the dedicated repo, and are pulled in this PR via the npin update. Validation for ACL is also present in the branch.]

### DIFF
--- a/config/src/converters/k8s/config/acl.rs
+++ b/config/src/converters/k8s/config/acl.rs
@@ -425,7 +425,7 @@ mod test {
             ),
         )
         .unwrap();
-        assert_eq!(r.action(), AclAction::Allow);
+        assert_eq!(r.action, AclAction::Allow);
 
         // 'to' blank, inferred from 'from'
         assert!(
@@ -501,7 +501,7 @@ mod test {
             ),
         )
         .unwrap();
-        assert!(!r.pattern().src().is_empty());
+        assert!(!r.pattern.src.is_empty());
 
         // Same subnet name doesn't exist under VPC-2: fails when VPC-2 is the "from" side
         assert!(
@@ -581,7 +581,7 @@ mod test {
             ),
         )
         .unwrap();
-        assert!(r.pattern().src().is_empty());
+        assert!(r.pattern.src.is_empty());
     }
 
     // An entry with ports but neither cidr nor vpcSubnet is legal ("any address, these ports") but
@@ -614,7 +614,7 @@ mod test {
             ),
         )
         .unwrap();
-        assert!(r.pattern().src().is_empty());
+        assert!(r.pattern.src.is_empty());
 
         let empty_ports = GatewayAgentPeeringsAclRulesMatch {
             dst: None,
@@ -690,8 +690,9 @@ mod test {
         .validate()
         .unwrap();
 
-        let mut acl = Acl::new(AclAction::Deny, vec![converted_rule]);
-        acl.validate(&left, &right).unwrap();
+        let acl = Acl::new(AclAction::Deny, vec![converted_rule])
+            .validate(&left, &right)
+            .unwrap();
         assert_eq!(
             acl.rules()[0].pattern().src(),
             &PrefixPortsSet::from([PrefixWithOptionalPorts::new(
@@ -751,15 +752,15 @@ mod test {
         // The two rules built via `rule()` default to no explicit scope (CRD `None`), which maps to
         // `AclScope::Flow`; the third rule explicitly requested `Packet` scope, which must survive
         // the conversion.
-        assert_eq!(converted.rules()[0].scope(), AclScope::Flow);
-        assert_eq!(converted.rules()[1].scope(), AclScope::Flow);
-        assert_eq!(converted.rules()[2].scope(), AclScope::Packet);
+        assert_eq!(converted.rules()[0].scope, AclScope::Flow);
+        assert_eq!(converted.rules()[1].scope, AclScope::Flow);
+        assert_eq!(converted.rules()[2].scope, AclScope::Packet);
 
         // Same for `log`: the two rules built via `rule()` default to no explicit `log` (CRD
         // `None`), which maps to `false`; the third rule explicitly requested `log: true`, which
         // must survive the conversion too.
-        assert!(!converted.rules()[0].log());
-        assert!(!converted.rules()[1].log());
-        assert!(converted.rules()[2].log());
+        assert!(!converted.rules()[0].log);
+        assert!(!converted.rules()[1].log);
+        assert!(converted.rules()[2].log);
     }
 }
--- a/config/src/converters/k8s/config/acl.rs
+++ b/config/src/converters/k8s/config/acl.rs
@@ -1,0 +1,765 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+use k8s_intf::gateway_agent_crd::{
+    GatewayAgentPeeringsAcl, GatewayAgentPeeringsAclDefault, GatewayAgentPeeringsAclRules,
+    GatewayAgentPeeringsAclRulesAction, GatewayAgentPeeringsAclRulesMatch,
+    GatewayAgentPeeringsAclRulesMatchDst, GatewayAgentPeeringsAclRulesMatchSrc,
+    GatewayAgentPeeringsAclRulesScope,
+};
+use lpm::prefix::{PortRange, Prefix, PrefixPortsSet, PrefixWithOptionalPorts};
+
+use crate::converters::k8s::FromK8sConversionError;
+use crate::converters::k8s::config::expose::parse_port_ranges;
+use crate::converters::k8s::config::{SubnetMap, VpcSubnetMap};
+use crate::external::overlay::acl::{Acl, AclAction, AclPattern, AclProtoMatch, AclRule, AclScope};
+
+fn resolve_prefix(
+    cidr: Option<&str>,
+    vpc_subnet: Option<&str>,
+    subnets: &SubnetMap,
+) -> Result<Prefix, FromK8sConversionError> {
+    match (cidr, vpc_subnet) {
+        // unreachable as we only call this function if at least one of these fields is not None
+        (None, None) => Err(FromK8sConversionError::MissingData(
+            "ACL match entry must specify either cidr or vpcSubnet".to_string(),
+        )),
+        (Some(_), Some(_)) => Err(FromK8sConversionError::NotAllowed(
+            "ACL match entry must specify either cidr or vpcSubnet, not both".to_string(),
+        )),
+        (Some(cidr), None) => cidr
+            .parse::<Prefix>()
+            .map_err(|e| FromK8sConversionError::InvalidData(format!("CIDR format: {cidr}: {e}"))),
+        (None, Some(subnet_name)) => subnets.get(subnet_name).copied().ok_or_else(|| {
+            FromK8sConversionError::NotAllowed(format!(
+                "ACL match references unknown VPC subnet {subnet_name}"
+            ))
+        }),
+    }
+}
+
+fn parse_proto(proto: Option<&str>) -> Result<AclProtoMatch, FromK8sConversionError> {
+    let Some(proto) = proto else {
+        return Ok(AclProtoMatch::Any);
+    };
+    match proto {
+        "tcp" => Ok(AclProtoMatch::Tcp),
+        "udp" => Ok(AclProtoMatch::Udp),
+        other => other.parse::<u8>().map(AclProtoMatch::Other).map_err(|_| {
+            FromK8sConversionError::InvalidData(format!(
+                "ACL protocol '{other}': expected \"tcp\", \"udp\", or a numeric protocol"
+            ))
+        }),
+    }
+}
+
+/// Common shape of an ACL match entry (`match.src[]` / `match.dst[]`).
+///
+/// The CRD generates distinct `...MatchSrc`/`...MatchDst` types with identical fields; this lets
+/// [`convert_match_side`] handle both sides without duplicating the resolution logic.
+trait AclMatchEntry {
+    fn cidr(&self) -> Option<&str>;
+    fn vpc_subnet(&self) -> Option<&str>;
+    fn ports(&self) -> Option<&[String]>;
+}
+
+impl AclMatchEntry for GatewayAgentPeeringsAclRulesMatchSrc {
+    fn cidr(&self) -> Option<&str> {
+        self.cidr.as_deref()
+    }
+    fn vpc_subnet(&self) -> Option<&str> {
+        self.vpc_subnet.as_deref()
+    }
+    fn ports(&self) -> Option<&[String]> {
+        self.ports.as_deref()
+    }
+}
+
+impl AclMatchEntry for GatewayAgentPeeringsAclRulesMatchDst {
+    fn cidr(&self) -> Option<&str> {
+        self.cidr.as_deref()
+    }
+    fn vpc_subnet(&self) -> Option<&str> {
+        self.vpc_subnet.as_deref()
+    }
+    fn ports(&self) -> Option<&[String]> {
+        self.ports.as_deref()
+    }
+}
+
+/// Parse a `ports` list shared by both concrete and "any address" match entries.
+///
+/// A present-but-empty list is rejected rather than silently treated as "all ports" or "no
+/// ports" — the CRD does not require callers to omit the field instead of sending an empty list,
+/// so guessing which meaning was intended would be unsafe for an access-control rule.
+fn parse_entry_ports(ports: &[String]) -> Result<Vec<PortRange>, FromK8sConversionError> {
+    if ports.is_empty() {
+        return Err(FromK8sConversionError::MissingData(
+            "ACL match entry ports list must not be empty; omit the field to match all ports"
+                .to_string(),
+        ));
+    }
+    ports.iter().try_fold(Vec::new(), |mut ranges, port_str| {
+        ranges.extend(parse_port_ranges(port_str)?);
+        Ok(ranges)
+    })
+}
+
+/// Build a [`PrefixPortsSet`] plus any "any address within the peering" port ranges from one side
+/// (`src` or `dst`) of an ACL rule's `match` block.
+fn convert_match_side<T: AclMatchEntry>(
+    entries: Option<&[T]>,
+    subnets: &SubnetMap,
+) -> Result<(PrefixPortsSet, Vec<PortRange>), FromK8sConversionError> {
+    let Some(entries) = entries else {
+        return Ok((PrefixPortsSet::new(), Vec::new()));
+    };
+    let mut items = Vec::new();
+    let mut any_ports = Vec::new();
+    for entry in entries {
+        match (entry.cidr(), entry.vpc_subnet(), entry.ports()) {
+            // All fields omitted, the rule matches all traffic
+            (None, None, None) => return Ok((PrefixPortsSet::new(), Vec::new())),
+            // Match all addresses, ports returned separately since we need the peering manifest
+            // to resolve them to prefixes
+            (None, None, Some(ports)) => any_ports.extend(parse_entry_ports(ports)?),
+            (cidr, vpc_subnet, ports) => {
+                let prefix = resolve_prefix(cidr, vpc_subnet, subnets)?;
+                match ports {
+                    None => items.push(PrefixWithOptionalPorts::from(prefix)),
+                    Some(ports) => {
+                        for range in parse_entry_ports(ports)? {
+                            items.push(PrefixWithOptionalPorts::new(prefix, Some(range)));
+                        }
+                    }
+                }
+            }
+        }
+    }
+    Ok((items.into_iter().collect(), any_ports))
+}
+
+impl From<GatewayAgentPeeringsAclRulesAction> for AclAction {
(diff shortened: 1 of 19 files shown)


config/src/converters/k8s/config/expose.rs
```diff
@@ -26,6 +26,32 @@ fn parse_port_ranges(ports_str: &str) -> Result<Vec<PortRange>, FromK8sConversio
...
 }
 
+/// Resolve a `Prefix` from an explicit CIDR string or a named VPC subnet.
+///
+/// Exactly one of `cidr`/`vpc_subnet` must be `Some`.
+/// Callers are responsible for enforcing that exclusivity themselves, since the wording for
+/// invalid combinations differs by call site; this only resolves whichever one is present.
+/// `context` is prefixed onto the "unknown VPC subnet" error to identify the caller (e.g.
+/// "Expose", "ACL match").
+pub(crate) fn resolve_cidr_or_subnet(
```
qmonnet: I would get rid of this function. I understand the intent of reusing the code between exposes and ACLs, but in `process_ip_block()` it becomes weird as we already know which field (`cidr` or `vpc_subnet`) is `None` and which is not. In `resolve_prefix()` for the ACL code, we also end up doing a match (in `resolve_cidr_or_subnet()`) within a match, and it doesn't look great.

Plus this helper means we've got to pass some context string, which doesn't help keep things readable.

Plus there's the `unreachable!()`, which isn't really locally proved to be unreachable - it would be easy to change the call site in the ACL file and to crash the dataplane by accident.

I'd just keep the code separate for these, never mind the few lines that are duplicated, I think it would be simpler.

config/src/converters/k8s/config/acl.rs
```diff
@@ -0,0 +1,708 @@
...
+        return Ok(AclProtoMatch::Any);
+    };
+    match proto {
+        "tcp" => Ok(AclProtoMatch::Tcp),
+        "udp" => Ok(AclProtoMatch::Udp),
+        other => other.parse::<u8>().map(AclProtoMatch::Other).map_err(|_| {
+            FromK8sConversionError::InvalidData(format!(
+                "ACL protocol '{other}': expected \"tcp\", \"udp\", or a numeric protocol"
+            ))
+        }),
```
qmonnet, line 52: Does the CRD currently support this? I probably missed it.

I think we should reject the numeric protocol option for now, and come back to this when we add support for ICMP ACLs. For now, I think `udp`/`tcp`/any should be OK.

config/src/converters/k8s/config/acl.rs
```diff
@@ -0,0 +1,708 @@
...
+}
+
+/// Build a [`PrefixPortsSet`] plus any "any address within the peering" port ranges from one side
+/// (`src` or `dst`) of an ACL rule's `match` block.
+///
+/// `None` (the field omitted) matches all traffic in this direction, per the CRD's documented
+/// semantics. An entry with neither `cidr` nor `vpcSubnet` set matches any address within the
+/// peering: if it also has no `ports`, the whole side becomes unrestricted (same as omitting the
+/// field); if it has `ports`, those ports are returned separately since they can't be resolved
+/// into concrete prefixes until the peering's manifest is known.
```
qmonnet: Nit: I'd probably move these comments closer to the code, rather than leaving Claude add them above the function:

```rust
            // All fields omitted, the rule matches all traffic: we can return directly
            // as no other entry can further extend the selection.
            (None, None, None) => return Ok((PrefixPortsSet::new(), Vec::new())),
```

etc.

config/src/external/overlay/acl.rs
```diff
@@ -149,42 +248,16 @@ impl AclRule {
...
-                    self.name,
-                )))
-            }
-            // Reject if one value is non-empty but does not match either manifest
-            _ => Err(ConfigError::InvalidAcl(format!(
-                "Rule '{}' has invalid 'from' or 'to' fields: '{}' -> '{}'",
-                self.name, self.from, self.to,
-            ))),
-        }
+        let (from, to) = complete_from_to(
```
qmonnet: If you complete `from` and `to` fields when parsing the CRDs, we probably don't need to redo the full completion thing at validation time. At validation time we can probably keep something like:

```rust
        // match...
            (from, to) if from == manifest_left.name() && to == manifest_right.name() => Ok(()),
            (from, to) if from == manifest_right.name() && to == manifest_left.name() => Ok(()),
            _ => return Err(...),
```

And move the completion code to the converter?

config/src/converters/k8s/config/acl.rs
```diff
@@ -0,0 +1,708 @@
...
+impl From<GatewayAgentPeeringsAclRulesAction> for AclAction {
+    fn from(action: GatewayAgentPeeringsAclRulesAction) -> Self {
+        match action {
+            GatewayAgentPeeringsAclRulesAction::Deny => AclAction::Deny,
+            GatewayAgentPeeringsAclRulesAction::Allow => AclAction::Allow,
+        }
+    }
+}
+
+impl From<GatewayAgentPeeringsAclRulesScope> for AclScope {
```
qmonnet: Nit: I think you could implement `From<&GatewayAgentPeeringsAclRulesScope>` instead (from a reference) and avoid `clone()`ing the scope when converting

config/src/converters/k8s/config/acl.rs
```diff
@@ -0,0 +1,708 @@
...
+            &VpcSubnetMap,
+            &str,
+            &str,
+            &GatewayAgentPeeringsAcl,
+        ),
+    ) -> Result<Self, Self::Error> {
+        // "deny-unless-exposed" has no dedicated representation in `AclAction` yet: traffic outside
+        // exposed prefixes is already filtered upstream by the expose/manifest logic, so `Allow` is
+        // the correct default at the ACL layer. The CRD documents "deny-unless-exposed" as the
+        // default when the field is empty/unset, so `KopiumEmpty` maps the same way.
```
qmonnet: I'd rephrase to remove the part about “no representation yet”. `deny-unless-exposed` maps internally to `Allow`, period. That's indeed because we already filter by the expose/peering/manifest configuration.

config/src/converters/k8s/config/acl.rs
```diff
@@ -0,0 +1,708 @@
...
+    }
+
+    #[test]
+    fn test_proto_parsing() {
+        assert_eq!(parse_proto(None).unwrap(), AclProtoMatch::Any);
+        assert_eq!(parse_proto(Some("tcp")).unwrap(), AclProtoMatch::Tcp);
+        assert_eq!(parse_proto(Some("udp")).unwrap(), AclProtoMatch::Udp);
+        assert_eq!(parse_proto(Some("47")).unwrap(), AclProtoMatch::Other(47));
+        assert!(parse_proto(Some("bogus")).is_err());
+        // "icmp" is no longer a supported match protocol per the fabric CRD.
```
qmonnet: Change to “not supported yet”

config/src/converters/k8s/config/acl.rs
```diff
@@ -0,0 +1,708 @@
...
+        };
+
+        let mut rule_with_scope = rule(
+            "scoped",
+            "VPC-1",
+            "VPC-2",
+            GatewayAgentPeeringsAclRulesAction::Deny,
+            None,
+        );
+        rule_with_scope.scope = Some(GatewayAgentPeeringsAclRulesScope::Packet);
```
qmonnet, line 743: Similarly to the scope, can you please also change the value for `.log` and ensure that it's preserved, at the end of this test function?

qmonnet (changes requested):
Looking at the converter (Emanuele's commit), it looks great overall, thanks! I've got some minor comments below, none really blocking, but a couple of things could be slightly improve.

I need to take another pass tomorrow for the logic behind the port ranges when no prefixes are specified, and the `expand_any_ports()` that seems somewhat redundant with the processing in the converter, but I'm too tired to figure it out properly tonight.

I'd appreciate if someone could take a look at the other commits, please (_wink_ Fredi)

qmonnet:
I rebased to address the (trivial) conflicts with recently merged #1616.

config/src/converters/k8s/config/expose.rs
edipascale: that make sense, claude was overzealous in his optimization. done

config/src/converters/k8s/config/acl.rs
edipascale: done

config/src/external/overlay/acl.rs
edipascale: done

config/src/converters/k8s/config/acl.rs
edipascale: doh! done

config/src/converters/k8s/config/acl.rs
edipascale: done

config/src/converters/k8s/config/acl.rs
edipascale: done

config/src/converters/k8s/config/acl.rs
edipascale: added

config/src/converters/k8s/config/acl.rs
edipascale: it is supported in the CRD, and personally I think it is pretty important to support it - there's life beyond TCP/UDP! But even if we end up not supporting it in dataplane yet, I would keep the converter for it - you can always reject it as not supported in the validation step later on, right?

qmonnet: OK fair enough

And yes we want to support it anyway, just not sure it will be in 26.03

config/src/external/overlay/vpcpeering.rs
```diff
@@ -825,9 +828,16 @@ impl VpcPeering {
...
             gwgroup: "default".to_string(),
+            acl: None,
         }
     }
 
+    #[must_use]
+    pub fn with_acl(mut self, acl: Acl) -> Self {
+        self.acl = Some(acl);
+        self
+    }
```
qmonnet: Nit, forgot it in yesterday's review: `with_acl()` is only called once, and the `acl` field is public, so I'd just do `vpc_peering.acl = acl;` instead of `vpc_peering = vpc_peering.with_acl(acl);` and get rid of this method.

config/src/external/overlay/acl.rs
```diff
@@ -228,6 +266,8 @@ impl AclRule {
             // manifest's prefixes, we'll use a wildcard match anyway.
         } else {
             let src_set = manifest_from.all_ips();
+            let any = Self::expand_any_ports(&src_set, &self.pattern.src_any_ports);
+            self.pattern.src = self.pattern.src.union_prefixes_and_ports(&any);
```
qmonnet: I don't like this part much. I think the AI was right in its report (not sure why it considers it as addressed now 🤔), here we change `self.pattern.src` to “expand” the port ranges over the existing prefixes in the manifest, but we don't do it in the case of a default expose.

This means that after this function, we've got 1) expanded prefixes-with-ports and `self.src_any_ports` is no longer useful, if the manifest doesn't have any default expose, or 2) `self.src_any_ports` that still contain port ranges to process later, in the case of a default expose, to restrict the application of the rule to the relevant port ranges.

Instead, how about we add a step before `validate_patterns_coverage()` to expand port ranges like this line does, and to deal with the case of manifests with default exposes to expand to the root prefix (`0.0.0.0/0` or `::/0`)? (Given that we use manifests from a `ValidatedPeering` we can know whether it's IPv4 or IPv6, see the first two commits in https://github.com/githedgehog/dataplane/pull/1611/commits if necessary.) Then `validate_patterns_coverage()` no longer has to deal with any special case for default exposes? And when it exits, `src_any_ports`/`dst_any_ports` are no longer necessary?

config/src/external/overlay/acl.rs
```diff
@@ -784,6 +825,59 @@ mod validation_tests {
...
+        p.src_any_ports = vec![PortRange::new(443, 443).unwrap()];
+        let rule = rule("r", "VPC-1", "VPC-2", AclAction::Allow, p);
+        let validated = validate_rule(rule).expect("should validate");
+        assert_eq!(
+            validated.pattern.src,
+            [prefix_with_ports("10.0.0.0/16", 443, 443)].into()
+        );
+        // dst was left fully unrestricted (empty, no any_ports): still "match everything".
+        assert_eq!(validated.pattern.dst, prefixes(&["10.1.0.0/16"]));
+    }
```
qmonnet, line 917: We should add one test similar to this but with a manifest containing a default expose, at the moment I don't think the ports would be passed down to `validated.pattern.src`.

config/src/external/overlay/acl.rs
```diff
@@ -784,6 +825,59 @@ mod validation_tests {
...
+
+    // A protocol that doesn't support ports (e.g. ICMP) rejects an "any address" port-restricted
+    // entry just like it rejects a concrete-prefix one.
+    #[test]
+    fn test_acl_any_address_with_ports_rejected_for_portless_proto() {
+        let mut p = pattern(
+            PrefixPortsSet::new(),
+            PrefixPortsSet::new(),
+            AclProtoMatch::Icmp,
```
qmonnet, line 926: Maybe also validate this with `AclProtoMatch::Any` (in the same test)?

qmonnet:
(And thanks a lot for addressing all of yesterday's feedback! :heart:)

config/src/external/overlay/acl.rs
edipascale: you are right, that was a bug. I cherry-picked the is_v4 commit from the flofi PR and used it here

config/src/external/overlay/vpcpeering.rs
edipascale: absolutely

config/src/external/overlay/acl.rs
edipascale: added a test to cover ports only + default expose in manifest

config/src/external/overlay/acl.rs
edipascale: added

config/src/external/overlay/acl.rs
```diff
@@ -217,28 +242,64 @@ impl AclRule {
...
-            // manifest's prefixes, we'll use a wildcard match anyway.
-        } else {
-            let dst_set = manifest_to.all_public_ips();
-            Self::validate_pattern_coverage(&self.name, &mut self.pattern.dst, &dst_set)?;
-        }
+        let dst_set = Self::manifest_coverage_set(manifest_to, is_v4, true);
+        let any =
+            Self::expand_any_ports(&dst_set, &std::mem::take(&mut self.pattern.dst_any_ports));
+        self.pattern.dst = self.pattern.dst.union_prefixes_and_ports(&any);
+        Self::validate_pattern_coverage(&self.name, &mut self.pattern.dst, &dst_set)?;
```
qmonnet, line 354: My last suggestion would be to maybe add some comments in the body of this method, to explain what we do; but that's not blocking.

qmonnet (approved):
Looks good, thanks!

I've pushed a commit on top to get rid of the `src_any_ports`, `dst_any_ports` fields after validation.

qmonnet:
Added a last commit to propagate the `Acl`/`ValidatedAcl` objects to `Peering`/`ValidatedPeering`.

qmonnet:
(Sorry I clicked the wrong button by accident and caused Copilot to somehow add a commit :facepalm: - Now removed)

qmonnet (approved):

Fredi-raspall:
Adding don't merge until we discuss if we embrace the validation changes.
See https://github.com/githedgehog/dataplane/pull/1631

config/src/external/overlay/acl.rs
```diff
@@ -0,0 +1,324 @@
...
+    Allow,
+    #[default]
+    Deny,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub enum AclProtoMatch {
+    Tcp,
+    Udp,
+    Icmp,
```
Fredi-raspall, line 23: Icmp seems not to be covered by the k8s converter. As the code stands, that variant is not possible I believe.

config/src/external/overlay/acl.rs
```diff
@@ -0,0 +1,1016 @@
...
+    pub dst_any_ports: Vec<PortRange>,
+    pub proto: AclProtoMatch,
+}
+
+impl AclPattern {
+    #[must_use]
+    pub(crate) fn new(
+        src: PrefixPortsSet,
+        dst: PrefixPortsSet,
+        src_any_ports: Vec<PortRange>,
```
Fredi-raspall, line 47: I wonder if these Vec<PortRange> could just be represented by `PrefixPortsSet` and therefore go along src and dst, given that a `PrefixPortsSet` is a collection of `PrefixWithOptionalPorts` and a wildcard on the prefix may be represented by a 0/0?

config/src/external/overlay/acl.rs
```diff
@@ -0,0 +1,1016 @@
...
+        Ok(ValidatedAclPattern {
+            src: self.src,
+            dst: self.dst,
+            proto: self.proto,
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct ValidatedAclPattern {
```
Fredi-raspall, line 96: Surprise! I would strongly encourage not to duplicate any of the types in this module with their validated versions.

Fredi-raspall (changes requested):
I believe there are 3 points to double check:
1) internal protocol variant ICMP (is it part of API?)
2) whether internally the only-match ports can be represented the same way as prefixes with ports
3) the need to add validated types.

config/src/external/overlay/acl.rs
qmonnet: Correct, [we decided to leave ICMP aside for now](https://github.com/githedgehog/fabric/pull/1498#discussion_r3520184710) and to add it later. I need to follow-up with some small validation adjustments about the current limitations for the feature.

config/src/external/overlay/acl.rs
qmonnet: We could probably _store_ them that way, the difficulty is that we don't know what wildcard prefix to use at parsing time: IPv4 `0.0.0.0/0`, or IPv6 `::/0`? We haven't validated the peering yet so we don't know what IP version to use.

config/src/external/overlay/acl.rs
qmonnet: Who would have thought :slightly_smiling_face: 

Given that you've been working on that, is this OK to leave the removal for your follow-up work?

qmonnet:
I rebased to simply address the conflict on the npins update for fabric.

@Fredi-raspall I replied to your comments, is this satisfying to you? (Or considering the validated structs, is this maybe “acceptable” rather than satisfying? :sweat_smile:)

config/src/external/overlay/acl.rs
Fredi-raspall: Does that mean that users won't be able to filter ICMP traffic ?

qmonnet: At the moment, yes. Although I expect it to be easy to add, so we can try to add it before the cut-off date for the release.

config/src/external/overlay/acl.rs
Fredi-raspall: > [ ..] We haven't validated the peering yet so we don't know what IP version to use.

Can't we know if the peering (or exposes) will use ipv4 or ipv6 at this stage?

qmonnet: Not without duplicating the processing that we have at validation time. We'd need to be sure that all expose blocks in both manifests of the peering are consistent with a single IP version. Even if we picked just the first version we find, we'd need to be sure that there's at least one expose block with an IP version. This can be checked, but would rather belong to validation stage. What's the matter with the dedicated fields? We drop them after validation anyway

config/src/external/overlay/acl.rs
Fredi-raspall: Well, if we know we'll not use the validated types, why adding them to remove them later?

qmonnet: I don't know we won't use them, I've not reviewed your PR yet. The current code (in this PR, and the follow-up I'm working on) uses them

Fredi-raspall:
> I rebased to simply address the conflict on the npins update for fabric.
> 
> @Fredi-raspall I replied to your comments, is this satisfying to you? (Or considering the validated structs, is this maybe “acceptable” rather than satisfying? 😅)

It is just half acceptable because of the type duplication :sweat_smile:

Fredi-raspall (approved):

PR #1622, opened 2026-07-07
ci: stop using opengrep

daniel-noland:
We have decided that this doesn't matter.


daniel-noland:
re-closing in favor of #1612

PR #1628, opened 2026-07-08
Simplify Prefix with opt ports and fix bug

Fredi-raspall:
This goes on top of https://github.com/githedgehog/dataplane/pull/1627

### DIFF
--- a/config/src/utils/overlap.rs
+++ b/config/src/utils/overlap.rs
@@ -113,8 +113,7 @@ pub fn merge_contiguous_prefixes(prefixes: &mut PrefixPortsSet) {
                 continue 'next_prefix;
             }
         }
-        // Also "simplify" the prefix before inserting: remove the associated port range if relevant
-        merged_prefixes.insert(prefix_left.simplify());
+        merged_prefixes.insert(prefix_left);
     }
     *prefixes = merged_prefixes;
 
--- a/config/src/utils/overlap.rs
+++ b/config/src/utils/overlap.rs
@@ -3,7 +3,7 @@
 
 use crate::ConfigError;
 use crate::external::overlay::vpcpeering::ValidatedExpose;
-use lpm::prefix::{IpRangeWithPorts, PrefixPortsSet, PrefixWithOptionalPorts};
+use lpm::prefix::{IpRangeWithPorts, PrefixPortsSet};
 
 pub fn check_private_prefixes_dont_overlap(
     expose_left: &ValidatedExpose,
@@ -94,7 +94,7 @@ pub fn merge_contiguous_prefixes(prefixes: &mut PrefixPortsSet) {
     sorted_prefixes.sort_by_key(|p| p.prefix().length());
 
     'next_prefix: while let Some(prefix_left) = sorted_prefixes.pop() {
-        if matches!(prefix_left, PrefixWithOptionalPorts::PrefixPorts(_)) {
+        if prefix_left.ports().is_some() {
             uses_ports = true;
         }
         for (index_right, prefix_right) in sorted_prefixes.iter().enumerate() {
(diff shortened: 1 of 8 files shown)


config/src/utils/overlap.rs
```diff
@@ -133,7 +133,7 @@ pub fn merge_contiguous_prefixes(prefixes: &mut PrefixPortsSet) {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use lpm::prefix::PortRange;
+    use lpm::prefix::{PortRange, PrefixWithOptionalPorts};
```
qmonnet, line 135: Strange, how did we miss the missing include so far? 🤔

lpm/src/prefix/with_ports.rs
```diff
@@ -233,233 +233,95 @@ impl std::ops::DerefMut for PrefixPortsSet {
...
-                false
-            }
-            _ => {
-                // We necessarily cover all prefixes from the other instance, only check prefixes
-                self.prefix().covers(&other.prefix())
-            }
-        }
+        // Since portrange = None means all ports, represent both self and other
+        // as `PrefixWithPorts` and use its `covers` implementation
+        PrefixWithPorts::from(*self).covers(&(*other).into())
```
qmonnet, line 283: smart 👍

lpm/src/prefix/with_ports.rs
```diff
@@ -482,18 +344,13 @@ pub enum PortRangeError {
 
 impl From<PrefixWithPorts> for PrefixWithOptionalPorts {
     fn from(value: PrefixWithPorts) -> Self {
-        PrefixWithOptionalPorts::PrefixPorts(value)
+        Self::new(value.prefix(), Some(value.ports()))
```
qmonnet, line 330: Call `.simplify()` here and get rid of `from_prefix_with_ports()`?

lpm/src/prefix/with_ports.rs
```diff
@@ -653,8 +507,12 @@ impl PortRange {
...
-            // We know we have left.start <= right.end given that left <= right
-            Some(PortRange::new(left.start, right.end).unwrap_or_else(|_| unreachable!()))
+            // left has the smaller start value, but the union ends at the larger of the two ends
+            // since left could contain right. So we have to take the highest of the two.
+            Some(
+                PortRange::new(left.start, left.end.max(right.end))
+                    .unwrap_or_else(|_| unreachable!()),
+            )
         }
     }
```
qmonnet, line 498: The commit log says it fixes `covers()`, but isn't that this function that it fixes instead? Can you please adjust the commit log if this is a mistake?

lpm/src/prefix/with_ports.rs
qmonnet, line 499: I think we could have something better-looking:

```rust
#[must_use]
pub fn merge(&self, other: Self) -> Option<Self> {
    let mergeable = self.end.min(other.end).saturating_add(1) >= self.start.max(other.start);

    mergeable.then(|| Self {
        start: self.start.min(other.start),
        end: self.end.max(other.end),
    })
}
```


[EDIT: In fact I don't think building the port range this way works 🤦 , sorry]

lpm/src/prefix/with_ports.rs
qmonnet: Never mind, I see you get rid of the function in a later commit.

qmonnet (changes requested):
Looks good, thanks! Making `PrefixWithOptionalPorts` does make the code simpler indeed.

I think we can have something nicer for the `merge()` method (see below), but it looks good otherwise.

qmonnet:
I checked and there's no conflict with #1618, by the way :+1:

lpm/src/prefix/with_ports.rs
Fredi-raspall: We have a distinct criteria for "better-looking" :sweat_smile:

qmonnet: Come on, this one doesn't have any `unreachable!()` in it!! :stuck_out_tongue:

Fredi-raspall: That's why it's uglier!

lpm/src/prefix/with_ports.rs
Fredi-raspall: fixed

qmonnet (approved):

qmonnet:
Rebased on `main` to get rid of the three (merged) commits from #1627, which for some reason were still in the base branch for this PR and were causing a conflict.

PR #1629, opened 2026-07-10
clean-ups + vpc route abstraction + extend validation for failover

Fredi-raspall:
* Adds a new representation for vpc peerings in the form of VPC level routes
 * Replaces validation of overlaps allowed based on the VPC routes
 * extends validation for failover gateway groups
 Fixes https://github.com/githedgehog/internal/issues/447

### DIFF
--- a/cli/src/cliproto.rs
+++ b/cli/src/cliproto.rs
@@ -286,6 +286,7 @@ pub enum CliAction {
     // config: vpcs & peerings
     ShowVpc,
     ShowVpcPeerings,
+    ShowVpcRouting,
 
     // router: Eventlog
     RouterEventLog,
--- a/config/src/display.rs
+++ b/config/src/display.rs
@@ -425,6 +425,10 @@ impl ValidatedVpcTable {
     pub fn as_peerings(&self) -> ValidatedVpcTablePeerings<'_> {
         ValidatedVpcTablePeerings(self)
     }
+    #[must_use]
+    pub fn as_route_tables(&self) -> VpcTableRoutingTables<'_> {
+        VpcTableRoutingTables(self)
+    }
 }
 
 impl Vpc {
@@ -498,7 +502,7 @@ impl Display for ValidatedOverlay {
 
 macro_rules! VPC_RT_TBL_FMT {
     () => {
-        "{:<30} {:<10} {:<12} {:<13} {:<16}"
+        " {:<30} {:<10} {:<12} {:<13} {:<16}"
     };
 }
 fn fmt_vpc_rt_table_heading(f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
--- a/config/src/display.rs
+++ b/config/src/display.rs
@@ -16,7 +16,9 @@ use crate::external::overlay::vpcpeering::{
     VpcExposePortForwarding, VpcExposeStaticNat,
 };
 use crate::external::overlay::vpcpeering::{VpcManifest, VpcPeering, VpcPeeringTable};
+use crate::external::overlay::vpcrouting::{ExposeAction, VpcRoute, VpcRouteTable};
 use crate::external::overlay::{Overlay, ValidatedOverlay};
+
 use chrono::{DateTime, Utc};
 use net::vxlan::Vni;
 
@@ -347,6 +349,18 @@ impl Display for ValidatedVpcSummary<'_> {
     }
 }
 
+pub struct VpcTableRoutingTables<'a>(&'a ValidatedVpcTable);
+impl Display for VpcTableRoutingTables<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for vpc in self.0.values() {
+            Heading(vpc.name().to_string()).fmt(f)?;
+            vpc.route_table().fmt(f)?;
+            writeln!(f)?;
+        }
+        Ok(())
+    }
+}
+
 pub struct VpcTableSummary<'a>(&'a VpcTable);
 impl Display for VpcTableSummary<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
@@ -477,6 +491,58 @@ impl Display for Overlay {
 impl Display for ValidatedOverlay {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         self.vpc_table().as_summary().fmt(f)
+    }
+}
+
+/* ===== VPC routing tables =====*/
+
+macro_rules! VPC_RT_TBL_FMT {
+    () => {
+        "{:<30} {:<10} {:<12} {:<13} {:<16}"
+    };
+}
+fn fmt_vpc_rt_table_heading(f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    writeln!(
+        f,
+        VPC_RT_TBL_FMT!(),
+        "destinations", "remote-VPC", "remote-vni", "remote-action", "gw-group"
+    )
+}
+
+impl Display for ExposeAction {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.pad(match self {
+            ExposeAction::Default => "default",
+            ExposeAction::Forward => "forward",
+            ExposeAction::StaticNat => "static-nat",
+            ExposeAction::PortForwarding => "port-forward",
+            ExposeAction::Masquerade => "masqueraded",
+        })
+    }
+}
+
+impl Display for VpcRoute {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let dst = self.destination().to_string();
+        writeln!(
+            f,
+            VPC_RT_TBL_FMT!(),
+            dst,
+            self.dst_vpc(),
+            self.dst_vni(),
+            self.remote_action(),
+            self.gw_group()
+        )
+    }
+}
+
+impl Display for VpcRouteTable {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        fmt_vpc_rt_table_heading(f)?;
+        for route_set in self.iter() {
+            route_set.fmt(f)?;
+        }
+        Ok(())
     }
 }
 
--- a/config/src/display.rs
+++ b/config/src/display.rs
@@ -18,6 +18,7 @@ use crate::external::overlay::vpcpeering::{
 use crate::external::overlay::vpcpeering::{VpcManifest, VpcPeering, VpcPeeringTable};
 use crate::external::overlay::{Overlay, ValidatedOverlay};
 use chrono::{DateTime, Utc};
+use net::vxlan::Vni;
 
 use common::cliprovider::Heading;
 const SEP: &str = "       ";
@@ -134,10 +135,11 @@ fn fmt_remote_manifest(
     f: &mut std::fmt::Formatter<'_>,
     manifest: &VpcManifest,
     remote_id: &VpcId,
+    remote_vni: Vni,
 ) -> std::fmt::Result {
     writeln!(
         f,
-        "     remote VPC is {} (id:{}):",
+        "     remote VPC is {} (id:{}), vni: {remote_vni}:",
         manifest.name, remote_id
     )?;
     for e in &manifest.exposes {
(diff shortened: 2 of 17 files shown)


config/src/external/overlay/vpcrouting.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Build a `VpcRouteTable` from the set of `ValidatedPeering` of a VPC
+    pub fn build(peerings: &Vec<ValidatedPeering>) -> Self {
+        let mut rt = VpcRouteTable::new();
+        for peering in peerings {
+            for expose in peering.remote().valexp() {
+                let destinations = if expose.is_default() {
+                    &PrefixPortsSet::root_v4()
+                } else {
+                    expose.public_ips()
+                };
```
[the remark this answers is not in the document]
Fredi-raspall: It builds correctly.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -839,7 +839,7 @@ impl VpcPeering {
     }
     /// Given a peering fetch the manifests, orderly depending on the provided vpc name
     #[must_use]
-    pub fn get_peering_manifests(&self, vpc: &str) -> (&VpcManifest, &VpcManifest) {
+    pub(crate) fn get_peering_manifests(&self, vpc: &str) -> (&VpcManifest, &VpcManifest) {
```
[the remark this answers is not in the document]
Fredi-raspall: It is intentional

config/src/external/overlay/tests.rs
```diff
@@ -688,14 +689,8 @@ pub mod test {
...
 
-        /* collect ids */
-        let id_map: VpcIdMap = vpc_table
-            .values()
-            .map(|vpc| (vpc.name.clone(), vpc.id.clone()))
-            .collect();
-
         /* collect the peerings for each VPC */
-        vpc_table.collect_peerings(&peering_table, &id_map);
+        vpc_table.collect_peerings(&peering_table);
```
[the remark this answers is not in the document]
Fredi-raspall: Fixed

config/src/external/overlay/vpc.rs
```diff
@@ -302,28 +325,34 @@ impl Vpc {
...
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Debug)]
 pub struct ValidatedVpc {
     name: String,                     /* name of vpc, used as key */
     id: VpcId,                        /* internal Id, unique*/
     vni: Vni,                         /* mandatory */
     interfaces: InterfaceConfigTable, /* user-defined interfaces in this VPC */
     peerings: Vec<ValidatedPeering>,  /* peerings of this VPC - NOT set via gRPC */
+    #[allow(unused)]
+    rt: VpcRouteTable,
```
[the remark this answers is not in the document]
Fredi-raspall: Removed

cli/src/cliproto.rs
```diff
@@ -286,6 +286,7 @@ pub enum CliAction {
     // config: vpcs & peerings
     ShowVpc,
     ShowVpcPeerings,
+    ShowVpcRouting,
```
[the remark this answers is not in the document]
Fredi-raspall: It's fine as it is

config/src/external/overlay/tests.rs
```diff
@@ -688,14 +687,8 @@ pub mod test {
...
 
-        /* collect ids */
-        let id_map: VpcIdMap = vpc_table
-            .values()
-            .map(|vpc| (vpc.name.clone(), vpc.id.clone()))
-            .collect();
-
         /* collect the peerings for each VPC */
-        vpc_table.collect_peerings(&peering_table, &id_map);
+        vpc_table = vpc_table.collect_peerings(&peering_table);
```
qmonnet, line 693: The review bot had a hidden comment about this:

Image

Sounds worth addressing.

config/src/external/overlay/vpcrouting.rs
```diff
@@ -0,0 +1,164 @@
...
+pub struct VpcRoute {
+    dst: PrefixWithOptionalPorts, // destination(s) this route applies to
+    dstvpc: String,               // destination VPC
+    dstvni: Vni,                  // data path discriminant towards destination
+    rem_action: ExposeAction,     // action required by remote VPC
+    gwgroup: String,              // gateway group handling the peering this route corresponds to
+}
+
+/// A type representing a set of routes to the same destination.
+/// This type is currently not public
```
qmonnet, line 48: We could remove this comment, we can tell from the absence of `pub`

config/src/external/overlay/vpcrouting.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Tell if a route is allowed to overlap with some other route.
+    fn can_overlap(&self, other: &VpcRoute) -> bool {
+        if self.dstvpc == other.dstvpc {
+            return true;
+        }
+        if self.is_masquerade() && other.is_masquerade() {
+            return true;
+        }
+        // both can't be default at the same time
+        self.is_default() ^ other.is_default()
```
qmonnet, line 79: Not sure I understand the reasoning for `^` rather than `||`. I agree that we don't support default expose to default expose connections, but if we did add support for it, we'd return false here if trying to compare two default expose blocks??

config/src/external/overlay/vpcrouting.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+    #[must_use]
+    fn is_masquerade(&self) -> bool {
+        self.rem_action == ExposeAction::Masquerade
+    }
+    /// Tell if a route is allowed to overlap with some other route.
+    fn can_overlap(&self, other: &VpcRoute) -> bool {
+        if self.dstvpc == other.dstvpc {
+            return true;
+        }
```
qmonnet, line 74: I'm not sure I understand this condition. Different prefixes in a given peering would have the same `dstvpc` but shouldn't overlap, except if one is port forwarding and the other is masquerade?

config/src/display.rs
```diff
@@ -480,6 +494,58 @@ impl Display for ValidatedOverlay {
...
+}
+
+impl Display for ExposeAction {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.pad(match self {
+            ExposeAction::Default => "default",
+            ExposeAction::Forward => "forward",
+            ExposeAction::StaticNat => "static-nat",
+            ExposeAction::PortForwarding => "port-forward",
+            ExposeAction::Masquerade => "masqueraded",
```
qmonnet, line 523:
```suggestion
            ExposeAction::Masquerade => "masquerade",
```

Would look more consistent with the other variants, using imperative

config/src/external/overlay/vpc.rs
```diff
@@ -347,6 +351,7 @@ pub struct ValidatedVpc {
     vni: Vni,                         /* mandatory */
     interfaces: InterfaceConfigTable, /* user-defined interfaces in this VPC */
     peerings: Vec<ValidatedPeering>,  /* peerings of this VPC - NOT set via gRPC */
+    rt: VpcRouteTable,
```
qmonnet: Nit: `route_table` is not that long and would be more explicit?

config/src/external/overlay/tests.rs
```diff
@@ -1171,7 +1171,7 @@ pub mod test {
 
         let overlay = Overlay::new(vpc_table, peering_table);
         assert!(overlay.validate().is_err_and(
-            |e| e == ConfigError::Forbidden("Multiple 'default' destinations exposed to VPC")
+            |e| e == ConfigError::Forbidden("Multiple default destinations exposed to VPC")
```
qmonnet, line 1174: I'm concerned that the term “default” may be confusing to users when seeing these logs. What's a default destination? That's why we had quotes. I'm not opposed to removing them, just voicing a concern in terms of user experience, I don't know how to make this clearer.

config/src/external/overlay/vpcrouting.rs
qmonnet: After seeing the next commits, I think I understand better. This part just replaces the check between the different peerings for the VPC, that you remove with `check_overlap_and_default()` in 59f2dd0220c80e0fd23111cd778616598a1cee24, but this means we still validate that “inside a peering, prefixes don't overlap” when validating individual peerings, right?

config/src/external/overlay/vpc.rs
qmonnet: More importantly, it would be nice to add some motivation about keeping this table in the structure. When reading `for future use` in the commit log, I thought it would be in a follow-up commit from the PR, but I see this is not the case. Can you explain in the commit log what the intended use case is, please?

config/src/external/overlay/tests.rs
Fredi-raspall: It is addressed already.

config/src/external/overlay/vpcrouting.rs
Fredi-raspall: Yes, I left the comment to avoid anyone adding pub unecessarily.

config/src/external/overlay/vpcrouting.rs
Fredi-raspall: ^ is xor so no default is fine, one default is fine, but no more.

config/src/external/overlay/vpcrouting.rs
Fredi-raspall: The method is used for routes across all peerings. If the route prefixes belong to the same vpc, we have already validated whether they can overlap or not. I can remove the condition.

config/src/display.rs
Fredi-raspall: Initially it was masquerade, but that was very confusing. I think the participle in this case makes more sense. Otherwise, the routing table would give the false impression that you would "route" to some prefix P and masquerade it, while, in reality the meaning is that P is masquerading some other prefix which is "masqueraded".

config/src/external/overlay/vpc.rs
Fredi-raspall: can be changed I guess

config/src/external/overlay/tests.rs
Fredi-raspall: Default is what users set in the CRD. If you quote it, couldn't users believe that it refers to something else?

config/src/external/overlay/vpc.rs
Fredi-raspall: changed

config/src/external/overlay/tests.rs
qmonnet: Right, apologies 🤦

config/src/external/overlay/vpcrouting.rs
qmonnet: I don't follow the reasoning for 2 default exposes... But then given that we don't allow 2 default exposes to connect, I guess it doesn't really matter for now. Ok

config/src/external/overlay/vpcrouting.rs
qmonnet: Ok sounds good then, OK to keep as is. Maybe add a comment

config/src/display.rs
qmonnet: OK, why not. It's only in `Display` anyway so no objection.

config/src/external/overlay/tests.rs
qmonnet: True, again I don't have an ideal solution here. Let's keep your change as is

qmonnet (approved):
Thanks for expanding the variable name, and answering my questions. Looks good to me!

config/src/external/overlay/vpcrouting.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    #[must_use]
+    /// Build a `VpcRouteTable` from the set of `ValidatedPeering` of a VPC
+    pub fn build(peerings: &Vec<ValidatedPeering>) -> Self {
+        let mut rt = VpcRouteTable::new();
+        for peering in peerings {
+            for expose in peering.remote().valexp() {
+                let destinations = if expose.is_default() {
+                    &PrefixPortsSet::root_v4()
+                } else {
```
qmonnet, line 136: Ouch, I just remembered, I spotted that during review and thought I would comment at the end in case it changed in a later commit, but it doesn't and I forgot it 🤦.

@Fredi-raspall This looks broken for IPv6 and needs a fix, I think? 😕

Fredi-raspall: As discussed, IPv6 support is not complete. Will address this in a separate PR

PR #1633, opened 2026-07-13
feat(config,mgmt): avoid prefix duplications

### LINKED ISSUE #1594: FRR advertise prefix-list gets duplicate entries when masquerade and port-forward share a NAT pool
When a GatewayPeering has two exposes for the same VPC that reference the same NAT pool CIDR, the generated FRR config contains two identical permit entries in the advertise prefix-list -- one per expose.

The following GatewayPeering (captured from run 20260613-130630-matrix-master, env-5/mesh/l3vni) triggers the bug:

```yaml
apiVersion: gateway.githedgehog.com/v1alpha1
kind: GatewayPeering
metadata:
  name: vpc-01--default
  namespace: default
spec:
  gatewayGroup: default
  peering:
    ext.default:
      expose:
        - default: true
    vpc-01:
      expose:
        - as:
            cidr: 192.168.85.0/24
          ips:
            - cidr: 10.50.1.0/24
          nat:
            masquerade:
              idleTimeout: 5m0s
        - as:
            cidr: 192.168.85.0/24
          ips:
            - cidr: 10.50.1.0/24
          nat:
            portForward:
              idleTimeout: 5m0s
              ports:
                - as: 15201
                  port: 5201
                  proto: tcp
```

Both exposes have `as.cidr: 192.168.85.0/24`. The generated gen 46 FRR config on gateway-1 contains:

```
ip prefix-list ADV-TO-EXT.DEFAULT-FOR-VPC-01 seq 1 permit 192.168.85.0/24
ip prefix-list ADV-TO-EXT.DEFAULT-FOR-VPC-01 seq 2 permit 192.168.85.0/24
```

Claude Analysis: `mgmt/src/processor/confbuild/internal.rs` lines 194-200 iterates `rmanifest.exposes` and calls `adv_prefixes()` per expose, appending results to the prefix-list without deduplication. `adv_prefixes()` (`config/src/external/overlay/vpcpeering.rs:257`) returns the pool prefix from `nat.as_range` for each NAT-bearing expose. When both exposes reference the same pool, the same prefix is added twice. The same flat_map at line 184 has the same issue for `adv_nets`.

All prior test variants (masquerade-only, port-forward-only) produce a single entry and pass. This is the first test to combine both on the same pool.

Is this the expected behavior?

Fredi-raspall:
On top of https://github.com/githedgehog/dataplane/pull/1629
Fixes https://github.com/githedgehog/dataplane/issues/1594

### DIFF
--- a/config/src/internal/routing/bgp.rs
+++ b/config/src/internal/routing/bgp.rs
@@ -271,7 +271,11 @@ impl AfIpv4Ucast {
         self.networks.push(network);
     }
     pub fn add_networks(&mut self, networks: impl IntoIterator<Item = Prefix>) {
-        self.networks.extend(networks);
+        for net in networks {
+            if !self.networks.contains(&net) {
+                self.add_network(net);
+            }
+        }
     }
 }
 impl AfIpv6Ucast {
@@ -293,7 +297,11 @@ impl AfIpv6Ucast {
         self.networks.push(network);
     }
     pub fn add_networks(&mut self, networks: impl IntoIterator<Item = Prefix>) {
-        self.networks.extend(networks);
+        for net in networks {
+            if !self.networks.contains(&net) {
+                self.add_network(net);
+            }
+        }
     }
 }
 impl AfL2vpnEvpn {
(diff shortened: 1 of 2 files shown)


qmonnet (approved):
Looks OK, thanks.
I'm not sure why this needs to be on top of #1629.

Fredi-raspall:
> Looks OK, thanks. I'm not sure why this needs to be on top of #1629.

Ok. Thx! No need to in reality.

qmonnet:
Needs a rebase now that #1629 has been merged.

PR #1634, opened 2026-07-13
User ACLs logic implementation

### LINKED ISSUE #1621: User ACL logic implementation

qmonnet:
Based on #1618 

It will probably require some changes in the way we use the ACLs (similar to the flow-filter reimplementation from #1611).


acl-filter/src/context.rs
```diff
@@ -0,0 +1,457 @@
...
+        match (template_rule.proto, template_rule.is_ipv4) {
+            (AclProtoMatch::Tcp, true) => self.tcp_v4.push(template_rule),
+            (AclProtoMatch::Udp, true) => self.udp_v4.push(template_rule),
+            (_, true) => self.other_v4.push(template_rule),
+
+            (AclProtoMatch::Tcp, false) => self.tcp_v6.push(template_rule),
+            (AclProtoMatch::Udp, false) => self.udp_v6.push(template_rule),
+            (_, false) => self.other_v4.push(template_rule),
+        }
+    }
```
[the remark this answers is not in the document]
qmonnet: Woopsies. Now fixed

acl-filter/src/tests.rs
```diff
@@ -0,0 +1,766 @@
...
+
+// -------------------------------------------------------------------------------------------------
+// End-to-end flow-scope test
+//
+// This runs a real NAT pipeline so the `related` flow link is established by masquerade (rather
+// than faked). The ACL filter sits right after the flow filter, i.e. where the packet still
+// carries VPC-internal (pre-source-NAT / post-destination-NAT) addresses that ACL rules are
+// written against. A 'flow'-scoped Allow rule authorizes the request; the reply matches no rule
+// directly but must be allowed because it belongs to the established flow.
+
```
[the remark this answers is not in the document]
qmonnet: fixed

acl-filter/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            let reverse_lookup_result = tables.lookup(&reverse_summary);
+            if let Some(result) = reverse_lookup_result
+                && result.action == AclAction::Allow
+                && result.scope == AclScope::Flow
+            {
+                let verdict = result.action;
+                if result.log {
+                    info!("ACL filtering: {summary:?}) -> {verdict:?} (reply from allowed flow)")
+                }
```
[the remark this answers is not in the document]
qmonnet: fixed

acl-filter/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+impl<Buf: PacketBufferMut> TryFrom<&Packet<Buf>> for PacketSummary {
+    type Error = DoneReason;
+
+    fn try_from(packet: &Packet<Buf>) -> Result<Self, Self::Error> {
+        let Some((src_vpcd, dst_vpcd)) = packet.meta().src_vpcd.zip(packet.meta().dst_vpcd) else {
+            debug!("Missing source VPC discriminant, dropping packet");
+            return Err(DoneReason::Unroutable);
+        };
```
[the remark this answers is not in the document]
qmonnet: fixed

acl-filter/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+        let guard = self.tablesr.load();
+        let tables = &guard.as_ref().acls;
+
+        // Look up for an ACL directly matching the packet
+        let lookup_result = tables.lookup(summary);
+        if let Some(result) = lookup_result {
+            let verdict = result.action;
+            if result.log {
+                info!("ACL filtering: {summary:?}) -> {verdict:?}")
+            }
```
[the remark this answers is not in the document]
qmonnet: fixed

acl-filter/src/context.rs
```diff
@@ -0,0 +1,457 @@
...
+            let local_vni = vpc.vni();
+            for peering in vpc.peerings() {
+                let Some(acl) = peering.acl() else {
+                    continue;
+                };
+                let remote_vni = overlay.vpc_table().get_remote_vni(peering);
+                for acl_rule in acl.rules() {
+                    ruleset.insert(
+                        local_vni,
+                        remote_vni,
```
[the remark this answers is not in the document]
qmonnet: fixed

config/src/external/overlay/acl.rs
```diff
@@ -421,6 +421,11 @@ impl Acl {
         manifest_left: &ValidatedManifest,
         manifest_right: &ValidatedManifest,
     ) -> Result<ValidatedAcl, ConfigError> {
+        if self.rules.is_empty() {
```
Fredi-raspall, line 440: Didn't we say that in the API there would be a default ACL action that would be configurable?
How would this check work in such a case?

config/src/external/overlay/acl.rs
```diff
@@ -76,7 +75,7 @@ impl AclPattern {
     fn validate_ports(&self) -> bool {
         match self.proto {
             AclProtoMatch::Tcp | AclProtoMatch::Udp => true,
-            AclProtoMatch::Other(_) | AclProtoMatch::Icmp | AclProtoMatch::Any => {
+            AclProtoMatch::Other(_) | AclProtoMatch::Any => {
```
Fredi-raspall, line 83: I think ICMP may actually deserve special treatment since users may want to filter out/allow certain types of errors? Also, there are icmp packets (e.g. errors) that we may want to allow regardless?

acl-filter/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+        let valid_flow = self.packet_has_valid_flow(packet.meta(), genid);
+
+        if self.lookup(&summary, valid_flow) == AclAction::Deny {
+            debug!("{nfi}: Packet rejected by ACLs, dropping packet");
+            packet.invalidate_flows();
+            packet.done(DoneReason::Filtered);
+        }
+    }
+
+    fn packet_has_valid_flow<'a>(
```
Fredi-raspall, line 67: Don't we have similar utils around that we could use for this?

config/src/external/overlay/acl.rs
qmonnet: The `Acl` object always has a default "rule" (a default action) if no rule in the list matches, but it's not in `self.rules`, it's `self.default`:

```rust
pub struct Acl {
    default: AclAction,    <-- default if no other rule matches
    rules: Vec<AclRule>,
}
```

So we reject the Acl object if there's no other rule than the default one. It's easier for the rest of the processing, and it doesn't really make sense to have an empty list anyway (just remove the ACL, or the peering if you want to block everything).

config/src/external/overlay/acl.rs
qmonnet: No objection with adding something specific for ICMP at some point.

At the `icmp` protocol field is not supported in the CRD right now (although it's trivial to add), and even if we add it, we don't need a specific variant internally.

If/when we want additional match options for ICMP in the future, we need to define the API first, update the CRD, and then we can revisit this part in dataplane - but at the moment I see no reason why to keep this variant.

acl-filter/src/lib.rs
qmonnet: Do you know where? I can't remember seeing one, but I can look.

Note: You may have seen something similar in the PR for the new flow-filter implementation? If we keep both (if we don't already have something for this, I mean), I'll move them to a shared location.

Fredi-raspall: probably masquerade / port-forwarding or flow-filter NFs use it

config/src/external/overlay/acl.rs
Fredi-raspall: Ok. Understood. I would not add that restriction, though. If there is a default, that means that there need not be rules. So a config with Acl = {default: whatever, rule = empty} should be a valid one, even if you would achieve the same behavior with other configuration objects (or the lack of them).

Fredi-raspall:
@qmonnet I'll continue my review once Daniel's changes land.

acl-filter/src/lib.rs
qmonnet: Flow-filter will after the re-implementation, I haven't been able to find one that fits here in masquerade or port-forwarding

qmonnet:
I integrated the code changes from Daniel. The code is now the same as in https://github.com/githedgehog/dataplane/pull/1639 (except for some `cargo update` that made it to that other PR, and which I dropped). It should pass CI (it did on the other PR) and should be ready to go. @mvachhar @daniel-noland @Fredi-raspall please take a look when you can (Daniel, please double-check the last commit).

config/src/external/overlay/acl.rs
qmonnet: https://github.com/githedgehog/dataplane/issues/1637

qmonnet:
All tests succeeded, but VLAB didn't run because I added the labels too late. Closing/reopening to trigger VLAB tests

daniel-noland (approved):
nice work.  LGTM

acl-filter/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+};
+
+pub struct AclFilter {
+    name: String,
+    tablesr: AclFilterContextReader,
+    pipeline_data: Arc<PipelineData>,
+}
+
+impl AclFilter {
+    pub fn new(name: &str, tablesr: AclFilterContextReader) -> Self {
```
Fredi-raspall, line 39: #[must_use]

acl-filter/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+        let genid = self.pipeline_data.genid();
+
+        let summary = match PacketSummary::try_from(&*packet) {
+            Ok(summary) => summary,
+            Err(reason) => {
+                packet.done(reason);
+                return;
+            }
+        };
+        let valid_flow = self.packet_has_valid_flow(packet.meta(), genid);
```
Fredi-raspall, line 58: If the flow has a flow but it is not valid, we should probably drop the packet already?

acl-filter/src/access.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct AclFilterContextReader(Arc<Slot<AclFilterContext>>);
+
+impl AclFilterContextReader {
+    /// Load the current context for read-only access.
+    #[must_use]
+    pub fn load(&self) -> Arc<AclFilterContext> {
+        self.0.load_full()
```
Fredi-raspall, line 93: Why load_full() and not load() ?

Fredi-raspall:
@qmonnet the PR looks good to me overall. I've pushed small enhancements / nits.
The only point to clarify is the use of load_full() (slower than load()) and whether we should drop packets pointing to non-valid flows.

acl-filter/src/lib.rs
qmonnet: Good catch, I'm annoyed this hasn't been surfaced by Clippy locally or in CI :thinking:. Fixed now

acl-filter/src/lib.rs
qmonnet: I don't think so? If the genid is too old, this means the flow was created from a previous config and is no longer valid; but this doesn't mean this packet is invalid. If it successfully goes through flow-filter without relying on the outdated flow, then gets marked for masquerade or port-forwarding, we can create a new updated flow for this packet?

acl-filter/src/access.rs
qmonnet: Because I couldn't get it to compile with `load()` for all of the standard case, loom, and shuttle otherwise. I used `load_full()` like we use in masquerade for the allocator. I can look a bit more into it, do you know what the right solution would be instead?

qmonnet: @Fredi-raspall The alternative I have consists in introducing dedicated features for the crate:

```diff
diff --git i/Cargo.lock w/Cargo.lock
index 66dc60c5e598..d49acb88cfc8 100644
--- i/Cargo.lock
+++ w/Cargo.lock
@@ -1241,6 +1241,7 @@ dependencies = [
 name = "dataplane-acl-filter"
 version = "0.22.0"
 dependencies = [
+ "arc-swap",
  "dataplane-acl",
  "dataplane-common",
  "dataplane-concurrency",
diff --git i/acl-filter/Cargo.toml w/acl-filter/Cargo.toml
index ce11065fc465..e60b6fb3c6eb 100644
--- i/acl-filter/Cargo.toml
+++ w/acl-filter/Cargo.toml
@@ -5,8 +5,14 @@ license.workspace = true
 publish.workspace = true
 version.workspace = true
 
+[features]
+loom = ["concurrency/loom"]
+shuttle = ["concurrency/shuttle"]
+shuttle_dfs = ["concurrency/shuttle_dfs", "shuttle"]
+
 [dependencies]
 acl = { workspace = true }
+arc-swap = { workspace = true }
 common = { workspace = true }
 concurrency = { workspace = true }
 config = { workspace = true }
diff --git i/acl-filter/src/access.rs w/acl-filter/src/access.rs
index 0667dda7961a..5c1eb1fd954d 100644
--- i/acl-filter/src/access.rs
+++ w/acl-filter/src/access.rs
@@ -83,9 +83,17 @@ pub struct AclFilterContextReader(Arc<Slot<AclFilterContext>>);
 
 impl AclFilterContextReader {
     /// Load the current context for read-only access.
+    #[cfg(not(any(feature = "loom", feature = "shuttle")))]
+    #[must_use]
+    pub fn load(&self) -> arc_swap::Guard<Arc<AclFilterContext>> {
+        self.0.load()
+    }
+
+    /// Load the current context for read-only access (loom/shuttle).
+    #[cfg(any(feature = "loom", feature = "shuttle"))]
     #[must_use]
     pub fn load(&self) -> Arc<AclFilterContext> {
-        self.0.load_full()
+        self.0.load()
     }
 }
 
```

If you think it's worth it, I can change.

@daniel-noland or do you know if we have a wrapper for this already? Introducing new crate features just to accommodate for the return type doesn't look optimal, maybe I'm missing something simpler?

qmonnet: Discussed in meeting (without Fredi): Let's keep as it is for now, the perf delta isn't worth troubling ourselves with the new crate features. I added a FIXME: comment in the code to explain that, at some point, we should add a wrapper around the type in the `concurrency` crate, so we can use `.load()`.

qmonnet:
> the PR looks good to me overall. I've pushed small enhancements / nits. The only point to clarify is the use of load_full() (slower than load()) and whether we should drop packets pointing to non-valid flows.

Fredi, I replied to your comments above. If the reply regarding the packet drop does not address your concern, we can discuss again and address as a follow-up. Regarding the `.lookup()`, we discussed it at the stand-up, it shouldn't have any impact on x86 and it's acceptable to keep for now.

I'm toggling auto-merge. Thanks for the reviews!

PR #1636, opened 2026-07-15
flofi reworked

daniel-noland:


acl/Cargo.toml
```diff
@@ -6,8 +6,9 @@ publish.workspace = true
 version.workspace = true
 
 [features]
-default = []
+default = ["dpdk"]
```
qmonnet, line 9: What's the consequence of this change, does it mean the dataplane needs to prepare the DPDK components by default?

acl/src/dpdk/dyn_table.rs
```diff
@@ -648,4 +648,342 @@ mod tests {
             other => panic!("expected EmptySpecs, got {:?}", other.err()),
         }
     }
+
+    // ---------------------------------------------------------------------------------------------
+    // SPIKE: de-risk the flofi two-table redesign field encoding (see flofi/docs/two-table-redesign.md).
```
qmonnet, line 653: We should probably not reference “flofi”, see my comment above - I created a new branch and cleaned these up.

qmonnet:
@daniel-noland I created a new branch: https://github.com/githedgehog/dataplane/tree/pr/qmonnet/flow-filter-v2 with my commits + your commits from this PR, with the following changes:

- Rebased on `main`
- Some minor fixes for `Display`, and to have the dataplane compile with Loom/Shuttle
- More importantly, I renamed `flofi` as `flow-filter` (and the corresponding types `Flofi*` as `FlowFilter*`), and removed all references to `flofi`, including in your commits.
- The rest of your commits should be unchanged

As discussed before, the name `flofi` was temporary, but I didn't expect you'd build that much code on top of my branch. Can you please use the new one going forward?

flofi/README.md
```diff
@@ -0,0 +1,194 @@
...
+SPDX-License-Identifier: Apache-2.0
+Copyright Open Network Fabric Authors
+-->
+
+# flofi (flow-filter)
+
+`flofi` is the pipeline stage that answers one question for every overlay packet:
+
+> Given where this packet came from (source VPC + 5-tuple), **which destination VPC
+> does it belong to, and what NAT does each end require** -- or should it be dropped?
```
qmonnet, line 11: That's actually two questions 🫣

flofi/src/context/display.rs
```diff
@@ -1,190 +1,177 @@
...
-                write!(indented(f).with_str("    "), "{ends}")?;
-            }
-        }
-        writeln!(f, "remote ends:")?;
-        write!(indented(f).with_str("  "), "{}", self.remote_ends)?;
-        match &self.default_remote_vpcd {
-            Some(vpcd) => writeln!(f, "default remote VPC: {vpcd}"),
-            None => writeln!(f, "default remote VPC: -"),
+// -------------------------------------------------------------------------------------------------
+// Production (rte_acl / opaque): a one-line summary per table.
```
qmonnet, line 32: I don't understand that change, is this because we don't have a way to get more information from the ACL at runtime? Because if we have one, I'd prefer to get something more verbose to get a better idea of the table contents

flofi/src/context/tables.rs
```diff
@@ -1,387 +1,569 @@
...
-            has_default: false,
-        }
+// IP protocol numbers. flofi only ever emits rules for TCP, UDP, or "any" (see
+// `L4Protocol`), so a non-TCP/UDP packet only ever needs to match an any-proto
+// (mask 0x00) rule -- any sentinel byte other than TCP/UDP works for it.
+const PROTO_TCP: u8 = 6;
+const PROTO_UDP: u8 = 17;
+const PROTO_OTHER: u8 = 0;
+
+const PORT_WILDCARD: RangeSpec<u16> = RangeSpec::new(0, u16::MAX);
```
qmonnet, line 91: We have an identical `PORT_RANGE_WILDCARD` in lpm/src/prefix/with_ports.rs

flofi/src/context/tables.rs
```diff
@@ -1,387 +1,569 @@
...
-            udp: Vec::new(),
-            other: Vec::new(),
-            has_default: false,
-        }
+// IP protocol numbers. flofi only ever emits rules for TCP, UDP, or "any" (see
+// `L4Protocol`), so a non-TCP/UDP packet only ever needs to match an any-proto
+// (mask 0x00) rule -- any sentinel byte other than TCP/UDP works for it.
+const PROTO_TCP: u8 = 6;
+const PROTO_UDP: u8 = 17;
+const PROTO_OTHER: u8 = 0;
```
qmonnet, line 89: Nit: I have something like this instead for the user ACL PR:

```suggestion
const PROTO_TCP: u8 = NextHeader::TCP.as_u8();
const PROTO_UDP: u8 = NextHeader::UDP.as_u8();
const PROTO_OTHER: u8 = 0;
```

(requires marking `as_u8()` as const)

qmonnet (commented):
Here are some minor comments, but I haven't gone through - let alone assimilated - all the code changes yet, I'm still looking at this.

qmonnet:
Rebased and somewhat reworked in https://github.com/githedgehog/dataplane/pull/1638

PR #1644, opened 2026-07-17
Release/v0.20 fixes

Fredi-raspall:
Minimal log fixes for v0.20.2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,24 +142,24 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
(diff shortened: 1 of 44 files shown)


qmonnet (approved):

qmonnet:
Do we have any sort of established process for backports? I guess not?

qmonnet:
We'll need to figure out something for CI, in the future :confused: This change is simple enough, but generally, not being able to test backports in CI is annoying

Frostman (approved):

Frostman:
> Do we have any sort of established process for backports? I guess not?

we have, it's been used for a while, but it's a big secret 😆

mvachhar:
@Frostman had some concerns about backporting some of these fixes, can we please revisit and have the minimal set of fixes that we are comfortable backporting to the release branch?

mvachhar:
Also, with the impending release, is this worth doing at this point?

Fredi-raspall:
> Also, with the impending release, is this worth doing at this point?

The condition to maintain v0.20 was "if it is not too costly".

In addition to the idea of just changing the bare minimum, I even wonder if we should pin the rust compiler. It appears that several commits were just needed due to stricter lints in a more recent compiler. That requires extra effort to address them and introduces more changes.

PR #1648, opened 2026-07-20
chore: Discard reserved IPs from the config entities

sergeymatov:


config/src/external/overlay/vpcpeering.rs
```diff
@@ -319,6 +356,17 @@ impl VpcExpose {
...
+        // Reject reserved prefixes in the address-bearing ranges. Only the allowed sets are checked;
+        // excluded ranges are subtractive and cannot introduce an endpoint.
+        for prefixes in [&self.ips, self.as_range_or_empty()] {
+            for prefix_with_ports in prefixes {
+                let prefix = prefix_with_ports.prefix();
+                if let Some(class) = special_use_class(&prefix) {
+                    return Err(ConfigError::SpecialUsePrefix(prefix, class));
+                }
+            }
+        }
```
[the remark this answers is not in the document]
sergeymatov: Valid concern. I will fix this

config/src/external/overlay/validation_tests.rs
```diff
@@ -151,20 +150,32 @@ mod test {
...
+        let expose = VpcExpose::empty().ip("169.254.1.2/32".into());
+        assert!(expose.validate().is_err());
+    }
+
+    // IPv6 link-local fe80::/10 (and any host within it) should be rejected
+    #[test]
+    fn test_ipv6_link_local_rejected() {
+        let expose = VpcExpose::empty().ip("fe80::1/128".into());
+        assert!(expose.validate().is_err());
+    }
```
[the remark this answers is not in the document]
sergeymatov: Ipv6Addr::LOCALHOST is ::1; covers(::1/128) is reflexively true  to loopback. ff00::/8 — test uses ff02::1/128. So its covered. No need for redundant tests

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,46 @@ use crate::utils::{
...
 use std::time::Duration;
 use tracing::warn;
 
+/// Reserved prefixes that are never valid tenant endpoints, with the reason they are rejected.
+static SPECIAL_USE_PREFIXES: LazyLock<Vec<(Prefix, &'static str)>> = LazyLock::new(|| {
+    [
+        (IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 8, "unspecified"),
+        (IpAddr::V4(Ipv4Addr::new(127, 0, 0, 0)), 8, "loopback"),
+        (IpAddr::V4(Ipv4Addr::new(169, 254, 0, 0)), 16, "link-local"),
+        (IpAddr::V4(Ipv4Addr::new(224, 0, 0, 0)), 4, "multicast"),
```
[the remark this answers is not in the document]
sergeymatov: Aint this is redundant? I mean, its partially included already. I will add " 240/4" part

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,65 @@ use crate::utils::{
...
+        (
+            IpAddr::V6(Ipv6Addr::new(0xff00, 0, 0, 0, 0, 0, 0, 0)),
+            8,
+            "multicast",
+        ),
+    ]
+    .into_iter()
+    // filter_map avoids unwrap/expect, forbidden outside tests; these entries are all well-formed.
+    .filter_map(|(addr, len, class)| Prefix::try_from((addr, len)).ok().map(|p| (p, class)))
+    .collect()
```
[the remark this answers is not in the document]
sergeymatov: Good call. I will fix this.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,69 @@ use crate::utils::{
...
+/// Reason a prefix is reserved, or `None` if it is usable.
+///
+/// Rejects only a prefix wholly inside a reserved block, so the roots `0.0.0.0/0` and `::/0`
+/// (which contain reserved blocks but are not contained by one) stay valid.
+fn special_use_class(prefix: &Prefix) -> Option<&'static str> {
+    SPECIAL_USE_PREFIXES
+        .iter()
+        .find(|(special, _)| special.covers(prefix))
+        .map(|(_, class)| *class)
+}
```
qmonnet: The doc comment for this function is not accurate, it doesn't _reject_ anything - it only tells whether the prefix uses any of the special-use prefixes, right? Can you please fix?

(I'm not sure we need it as a separate function by the way, I'd probably inline it inside of `reject_special_use()`, but I don't mind much.)

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,69 @@ use crate::utils::{
...
 use std::collections::BTreeMap;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::time::Duration;
 use tracing::warn;
 
+/// Reserved prefixes that are never valid tenant endpoints, with the reason they are rejected.
+static SPECIAL_USE_PREFIXES: LazyLock<Vec<(Prefix, &'static str)>> = LazyLock::new(|| {
+    [
+        (IpAddr::V4(Ipv4Addr::UNSPECIFIED), 8, "unspecified"),
+        (IpAddr::V4(Ipv4Addr::new(127, 0, 0, 0)), 8, "loopback"),
```
qmonnet: Nit: you could use the `IpAddr::V4` wrapping part to the `.map()` at the end of the function, or even use `Prefix::from_str("127.0.0.0/8")` directly instead of `IpAddr::V4(Ipv4Addr::new(127, 0, 0, 0)), 8`, it would be easier to read?

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,69 @@ use crate::utils::{
...
+        (IpAddr::V4(Ipv4Addr::new(169, 254, 0, 0)), 16, "link-local"),
+        (IpAddr::V4(Ipv4Addr::new(224, 0, 0, 0)), 4, "multicast"),
+        // Listed before 240/4 so the more specific match wins for 255.255.255.255.
+        (IpAddr::V4(Ipv4Addr::BROADCAST), 32, "limited broadcast"),
+        (
+            IpAddr::V4(Ipv4Addr::new(240, 0, 0, 0)),
+            4,
+            "reserved (240/4)",
```
qmonnet:
```suggestion
            "reserved (240.0.0.0/4)",
```

Not much longer, and more explicit in debug messages

config/src/external/overlay/vpcpeering.rs
```diff
@@ -340,6 +405,14 @@ impl VpcExpose {
         if let Some(nat) = &mut clone.nat {
             normalize(&mut nat.as_range);
         }
+
+        // Re-check the effective (post-exclusion) sets: collapsing can carve a broad allowed prefix
+        // down to a fragment that lands wholly inside a reserved range.
+        reject_special_use(&clone.ips)?;
+        if let Some(nat) = &clone.nat {
+            reject_special_use(&nat.as_range)?;
+        }
```
qmonnet: I don't see the point in checking before and after the exclusion prefix collapsing. Just keep this one after the collapse. This way:

1. We avoid checking twice
2. If an allowed prefix is covered by a special prefix, _but_ there exists an exclusion prefix removing it, then it's an acceptable configuration because we won't end up using the prefix covered by a special prefix. So the early check is not entirely correct, only this one should be kept.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,69 @@ use crate::utils::{
...
 use lpm::prefix::{IpRangeWithPorts, L4Protocol, Prefix, PrefixPortsSet, PrefixWithOptionalPorts};
 use std::collections::BTreeMap;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::time::Duration;
 use tracing::warn;
 
+/// Reserved prefixes that are never valid tenant endpoints, with the reason they are rejected.
+static SPECIAL_USE_PREFIXES: LazyLock<Vec<(Prefix, &'static str)>> = LazyLock::new(|| {
+    [
+        (IpAddr::V4(Ipv4Addr::UNSPECIFIED), 8, "unspecified"),
```
qmonnet: Can we change the string for this one? In logs it will read like no reason was specified. Could be `INADDR_ANY`, or maybe just `0.0.0.0`?

config/src/external/overlay/vpcpeering.rs
```diff
@@ -11,9 +11,69 @@ use crate::utils::{
...
+});
+
+/// Reason a prefix is reserved, or `None` if it is usable.
+///
+/// Rejects only a prefix wholly inside a reserved block, so the roots `0.0.0.0/0` and `::/0`
+/// (which contain reserved blocks but are not contained by one) stay valid.
+fn special_use_class(prefix: &Prefix) -> Option<&'static str> {
+    SPECIAL_USE_PREFIXES
+        .iter()
+        .find(|(special, _)| special.covers(prefix))
```
qmonnet: Shouldn't this be `collides_with()` instead of `covers()`? If the special prefix does not fully cover the allowed prefix, then:

1. In theory, we could have the non-covered part of the allowed prefix covered by some _other_ special-use prefix, so maybe there's no valid IP left in the allowed prefix anyway; but we don't know, because we check coverage from individual special prefixes only, rather than checking coverage by the union of these prefixes. I don't think we have contiguous prefix in `SPECIAL_USE_PREFIXES` so it's probably not an issue in practice but still something we'd have to keep in mind if we were to expand the list for whatever reason.

2. If we accept prefixes that contains IPs from some special-use ranges, in particular for `as_range` prefixes, we risk using these reserved IPs in practice, for example by using them for masquerading in the allocator.

So instead, I think we should reject if any IP from Prefix is also in one of the special use prefixes? If, for example, the user wants 0.0.0.0/8 without breaking it down into a list of smaller prefixes to avoid including 0.0.0.0, they can always add a `not: 0.0.0.0/32` to comply?

config/src/external/overlay/validation_tests.rs
```diff
@@ -124,45 +124,113 @@ mod test {
...
+            .unwrap()
             .ip("10.0.0.1/32".into())
             .as_range("255.255.255.255/32".into())
-            .unwrap();
-        assert!(expose.validate().is_err());
+            .unwrap()
+            .validate();
+        assert!(
+            matches!(result, Err(ConfigError::SpecialUsePrefix(..))),
+            "{result:?}"
```
qmonnet: Maybe check the exact message for this one, to make sure you don't get the one for `240.0.0.0/4`

qmonnet (changes requested):

config/src/external/overlay/vpcpeering.rs
sergeymatov: Maybe I can use `collides_with()` only for `as` ranges? I think `covers()` for `ips` is fine

config/src/external/overlay/vpcpeering.rs
sergeymatov: Oh, I thought I removed the special use case before the prefix collapsing right after copilot mentioned it in previous review. Thanks for noticing it. Fixed

config/src/external/overlay/vpcpeering.rs
sergeymatov: Done

config/src/external/overlay/vpcpeering.rs
sergeymatov: Done, added more user-friendly note

config/src/external/overlay/vpcpeering.rs
sergeymatov: you are right. I Changed the behavior after the doc comment was added

config/src/external/overlay/validation_tests.rs
sergeymatov: done

config/src/external/overlay/vpcpeering.rs
sergeymatov: So as discussed, I put `collides_with()` and refactored tests

config/src/external/overlay/vpcpeering.rs
sergeymatov: done

config/src/external/overlay/validation_tests.rs
```diff
@@ -124,45 +124,114 @@ mod test {
...
+        let result = VpcExpose::empty().ip("ff02::1/128".into()).validate();
+        assert!(
+            matches!(result, Err(ConfigError::SpecialUsePrefix(..))),
+            "{result:?}"
+        );
+    }
+
+    // A broad prefix that only overlaps reserved space passes the raw check, but excluding the
+    // non-reserved part leaves a wholly-reserved fragment. The post-exclusion re-check must reject
+    // it. Here 224.0.0.0/3 minus 240.0.0.0/4 collapses to multicast 224.0.0.0/4.
```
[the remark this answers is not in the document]
sergeymatov: fixed

config/src/errors.rs
```diff
@@ -68,6 +68,8 @@ pub enum ConfigError {
     OverlappingPrefixes(PrefixWithOptionalPorts, PrefixWithOptionalPorts),
     #[error("Inconsistent IP version in VpcExpose: {0}")]
     InconsistentIpVersion(Box<VpcExpose>),
+    #[error("Prefix {0} is a special-use/reserved range ({1}) and cannot be used in a VpcExpose")]
```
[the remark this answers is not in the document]
sergeymatov: fixed

qmonnet (approved):
Looks good to me, thanks for addressing all the feedback!

PR #1657, opened 2026-07-22
flow filter fixes

daniel-noland:


qmonnet:
CI looks good apart for the known flaky test.

qmonnet:
I've mostly looked at the first commit and I think it's good, I still need to go through it some more. The rest of the commits seems to be mostly clean-ups and additional tests on top of the flow-filter work, so no objection.

qmonnet:
Rebased on current `pr/qmonnet/flow-filter-v2`.

qmonnet (approved):
The first commit looks good, thanks for working on that fix. If anything, maybe a quick pass to simplify some comments could help with readability.

I've only glanced superficially over the rest, given that it's fuzzing and other tests, although I plan to look again at the fuzzing parts out of curiosity. Looks good to merge, I'm happy to either merge into the base PR (bigger PR but fewer merge queue runs) or as a follow-up, as you prefer.

qmonnet:
Moving to “ready for review” given that the previous iteration passed CI, except for a known flake.

PR #1671, opened 2026-07-29
fix(interface-manager): Do not delete interfaces we did not create

daniel-noland:
The reconciler garbage collected every observed interface which was absent from the plan, exempting only interfaces of kinds it does not understand.  The dataplane does not own the network namespace it runs in, so this destroyed other people's devices: flannel's vxlan device parses as an ordinary VTEP and its `cni0` parses as an ordinary bridge, and both were removed on the first reconciliation pass, severing pod networking on the node.

Ownership is now decided by name.  `ManagedInterfaceName` defines the dataplane's naming scheme (`-vrf`, `-bri`, `-vtp`, `-tap`) in one place, and both the code which generates those names and the code which recognizes them go through it, so the two cannot drift apart.  An interface whose name does not fit the scheme is foreign: the reconciler leaves it strictly alone.

The rule is deliberately asymmetric.  Failing to recognize one of our own interfaces leaks it, which is bounded and self correcting. Mistaking somebody else's interface for one of ours destroys their network, which is neither.

Foreign interfaces are also excluded from the observed vtep and vrf property indexes.  Those indexes are keyed uniquely by vni and route table id in order to detect collisions between interfaces we manage, and a colliding foreign interface (flannel defaults to vni 1) evicted a legitimate observation of ours from the map, leaving the reconciler to conclude that its own interface was missing and to try, and fail, to recreate it until the config apply gave up.

The new integration test builds a flannel-like vxlan device and bridge in a private network namespace, gives a VPC the same vni flannel is using, and asserts that flannel's devices survive while a stale interface of ours is still collected.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1680,6 +1680,7 @@ dependencies = [
  "multi_index_map",
  "n-vm",
  "netdev",
+ "nix 0.31.3",
  "pretty_assertions",
  "rtnetlink",
  "serde",
--- a/interface-manager/src/interface/managed.rs
+++ b/interface-manager/src/interface/managed.rs
@@ -1,0 +1,349 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Which network interfaces the dataplane is allowed to manage.
+//!
+//! The dataplane does not own the network namespace it runs in.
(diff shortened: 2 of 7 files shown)


Frostman:
I've tested on VLAB and it seems to be fine and not touching the flannel-managed interfaces

mvachhar (approved):

PR #1679, opened 2026-07-31
flow filter printer

### LINKED ISSUE #1667: Fuzzer for User ACLs

daniel-noland:
Make the production CLI show the rules installed in the flow-filter and ACL-filter tables, rather than only reporting their counts.

DPDK’s `rte_acl` contexts are opaque after construction, so each table now retains the typed rules from which its classifier was built. Rendering from those typed rules lets domain values format themselves—for example, VNIs appear as VNIs and protocols as `TCP`—without decoding erased predicate bytes by width or position.

Rules are displayed in match order as a fixed-width grid containing:

- the rule’s rank;
- its typed match fields;
- a separator between matches and actions;
- the resulting action fields.

The displayed rank is the rule’s match position, not its internal priority encoding.

The classifier layout and matching semantics are intended to remain unchanged. The principal runtime tradeoff is that tables retain a typed copy of their rules and actions for operator inspection.

The final commit contains the user-visible behavior. The preceding commits establish the typed representation and formatting infrastructure it depends on.

### DIFF
--- a/acl-filter/src/context.rs
+++ b/acl-filter/src/context.rs
@@ -18,7 +18,8 @@ use config::external::overlay::acl::{AclAction, AclProtoMatch, AclScope, Validat
 use dpdk::acl::{CategoryMask, Priority};
 use lookup::Lookup;
 use lpm::prefix::{Prefix, PrefixPortsSet, PrefixWithOptionalPorts};
-use match_action::{Erased, ExactSpec, FieldPredicate, FixedSize, MaskSpec, MatchKey, RangeSpec};
+use match_action::{Erased, ExactSpec, FieldPredicate, FixedSize, MatchKey, PrefixSpec, RangeSpec};
+use net::ip::NextHeader;
 use net::vxlan::Vni;
 use std::collections::HashMap;
 use std::fmt;
@@ -167,25 +168,31 @@ impl From<&ValidatedOverlay> for PeeringAclRuleSet {
 #[derive(Debug, MatchKey, Clone, PartialEq, Eq)]
 pub(super) struct AclKey<I> {
     #[mask]
-    proto: u8,
+    proto: NextHeader,
     #[exact]
-    src_vni: u32,
+    #[cli(column_name = "src-vni")]
+    src_vni: Vni,
     #[exact]
-    dst_vni: u32,
+    #[cli(column_name = "dst-vni")]
+    dst_vni: Vni,
     #[prefix]
+    #[cli(column_name = "source")]
     src_ip_range: I,
     #[prefix]
+    #[cli(column_name = "destination")]
     dst_ip_range: I,
     #[range]
+    #[cli(column_name = "src-port")]
     src_port_range: u16,
     #[range]
+    #[cli(column_name = "dst-port")]
     dst_port_range: u16,
 }
 
 impl<I> AclKey<I> {
     #[must_use]
     fn new(
-        proto: u8,
+        proto: NextHeader,
         src_vni: Vni,
         dst_vni: Vni,
         src_ip: I,
@@ -195,8 +202,8 @@ impl<I> AclKey<I> {
     ) -> Self {
         Self {
             proto,
-            src_vni: src_vni.as_u32(),
-            dst_vni: dst_vni.as_u32(),
+            src_vni,
+            dst_vni,
             src_ip_range: src_ip,
             dst_ip_range: dst_ip,
             src_port_range: src_port.unwrap_or(0),
@@ -208,7 +215,7 @@ impl<I> AclKey<I> {
 /// Lower a single rule to backend-neutral field predicates for the concrete IP version of its
 /// prefixes. Returns `None` if the source and destination prefixes disagree on IP version, which
 /// the config validation already rules out for a well-formed peering.
-fn rule_predicates(
+fn rule_predicates<T: IpVersion>(
     proto: AclProtoMatch,
     src_vni: Vni,
     dst_vni: Vni,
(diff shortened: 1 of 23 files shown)


mvachhar (dismissed):
Please fix up some of these comments, plus could you post in the comments some sample output from the Display formatter so we can see what these look like.  Also printing in priority order would be very helpful.

acl/tests/eal_install_classify.rs
```diff
@@ -22,6 +22,12 @@ impl IpProto {
     const TCP: Self = IpProto(6);
 }
 
+impl core::fmt::Display for IpProto {
```
Fredi-raspall, line 25: Why is IpProto re defined also in eal_classify_via_projection.rs, as well as its Display and FixedSize implementations?

config/src/external/overlay/vpc.rs
```diff
@@ -151,8 +151,9 @@ impl ValidatedPeering {
     }
 
     fn validate_nat_combinations(&self) -> ConfigResult {
-        // If stateful NAT is set up on one side of the peering, we don't support NAT (static or
-        // stateful) on the other side.
+        // If stateful NAT is set up on one side of the peering, we don't support stateful NAT on
```
Fredi-raspall: I believe this comment is out-dated and should be fixed.

flow-filter/src/context/tables.rs
```diff
@@ -82,6 +97,12 @@ pub(crate) struct LookupInput {
     pub(crate) dst_ip: IpAddr,
     pub(crate) proto: NextHeader,
     pub(crate) ports: Option<(u16, u16)>,
+    /// The destination VPC recorded on the packet's flow, if it has one.
+    ///
+    /// An input to the lookup, not an output: it is the candidate stage 3 verifies when stage 1
+    /// reports a masquerade destination. Carried here rather than consulted in the NF so the
+    /// verification stays on the batched path with the other two stages.
+    pub(crate) flow_dst_vpcd: Option<VpcDiscriminant>,
```
Fredi-raspall: What is the purpose of this?

flow-filter/src/lib.rs
```diff
@@ -174,6 +179,30 @@ impl FlowFilter {
         let nfi = &self.name;
         let (dst_vpcd, dst_nat_mode, src_nat_mode) = match result {
             LookupResult::Route(route) => route,
+            // A masquerade destination cannot accept a new connection, so only reply traffic on an
+            // established flow may pass -- and only towards the VPC the tables confirmed
+            // masquerades this destination for this source. The tables cannot name that VPC from
+            // the address alone (two peers may masquerade behind one range), so they verified the
+            // candidate the flow supplied; `None` means no candidate, or one the configuration
+            // does not agree with. Either way nothing vouches for the packet.
+            LookupResult::MasqueradeDestination(verified) => {
```
Fredi-raspall: I don't understand the need for MasqueradeDestination and the problem it addresses.
If a packet has a flow and the flow is active, it will bypass the filter and never get here.
If the flow is not anymore active, the packet should be just dropped; the flow entry won't live much (if it does at all). If the flow was outdated (a config was changed), the masquerading module will take care of checking if it has to survive or be cancelled. Am I missing something here?

match-action/tests/derive_roundtrip.rs
```diff
@@ -9,6 +9,12 @@ use dataplane_match_action::{
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 struct IpProto(u8);
```
Fredi-raspall, line 10: I think this is the 3rd definition I see of IpProto ? (and the impl of FixedSize and Display)

Fredi-raspall (changes requested):
@daniel-noland thanks for writing this.
However, it goes beyond the goal of being able to expose the contents of the flow filter  tables. Can you please clarify the need for the changes related to masquerading?

Fredi-raspall:
The current  version shows as:
```
dataplane(✔)# show flow-filter table
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Routing context (flow filter) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
remote v4 (destination -> dst VPC + dst NAT):
  [0] proto=*, src_vni=3000, dst_ip=1.2.3.4/32, dst_port=* -> VNI(4000), NAT: masquerade
  [1] proto=*, src_vni=2000, dst_ip=20.30.90.30/32, dst_port=* -> VNI(4000), NAT: masquerade
  [2] proto=*, src_vni=3000, dst_ip=192.168.128.0/27, dst_port=* -> VNI(2000), NAT: -
  [3] proto=*, src_vni=4000, dst_ip=192.168.128.0/27, dst_port=* -> VNI(2000), NAT: -
  [4] proto=TCP, src_vni=3000, dst_ip=20.10.90.0/24, dst_port=2222 -> VNI(4000), NAT: port-forwarding
  [5] proto=UDP, src_vni=3000, dst_ip=20.10.90.0/24, dst_port=2053 -> VNI(4000), NAT: port-forwarding
  [6] proto=*, src_vni=3000, dst_ip=192.168.80.0/24, dst_port=* -> VNI(4000), NAT: -
  [7] proto=*, src_vni=3000, dst_ip=192.168.100.0/24, dst_port=* -> VNI(2000), NAT: -
  [8] proto=*, src_vni=4000, dst_ip=192.168.50.0/24, dst_port=* -> VNI(3000), NAT: -
  [9] proto=*, src_vni=4000, dst_ip=192.168.60.0/24, dst_port=* -> VNI(3000), NAT: -
  [10] proto=*, src_vni=2000, dst_ip=192.168.50.0/24, dst_port=* -> VNI(3000), NAT: -
local v4 (source -> src NAT):
  [0] proto=*, src_vni=2000, dst_vni=3000, src_ip=192.168.128.0/27, src_port=* -> NAT: -
  [1] proto=*, src_vni=2000, dst_vni=4000, src_ip=192.168.128.0/27, src_port=* -> NAT: -
  [2] proto=*, src_vni=3000, dst_vni=4000, src_ip=192.168.50.0/24, src_port=* -> NAT: -
  [3] proto=*, src_vni=3000, dst_vni=4000, src_ip=192.168.60.0/24, src_port=* -> NAT: -
  [4] proto=*, src_vni=3000, dst_vni=2000, src_ip=192.168.50.0/24, src_port=* -> NAT: -
  [5] proto=*, src_vni=4000, dst_vni=3000, src_ip=192.168.90.0/24, src_port=* -> NAT: masquerade
  [6] proto=*, src_vni=4000, dst_vni=3000, src_ip=192.168.80.0/24, src_port=* -> NAT: -
  [7] proto=*, src_vni=4000, dst_vni=2000, src_ip=192.168.90.0/24, src_port=* -> NAT: masquerade
  [8] proto=*, src_vni=2000, dst_vni=3000, src_ip=192.168.100.0/24, src_port=* -> NAT: -
remote v6 (destination -> dst VPC + dst NAT):
  (no rules)
local v6 (source -> src NAT):
  (no rules)
```
I would opt for something more compact and human-friendly like
```
Remotes:
  rank  srcVpc   destination     proto   dst-port      to         action
  [0]   3000     1.2.3.4/32        *        *       |  VNI(4000)  masquerade
  [1]   2000     20.30.90.30/32    *        *       |  VNI(4000)  masquerade
  [2]   3000     192.168.128.0/27  *        *       |  VNI(2000)  -
  [3]   4000     192.168.128.0/27, *        *       |  VNI(2000)  -
  [4]   3000     20.10.90.0/24     TCP     2222     |  VNI(4000)  port-forwarding
  [5]   3000     20.10.90.0/24     UDP     2053     |  VNI(4000)  port-forwarding

Local:
  rank  srcVpc   dstVpc  source                        action
  [0]   2000     3000    192.168.128.0/27           |  -
  [1]   2000     4000    192.168.128.0/27           |  -     
``` 

* Side note: is the proto match field needed in the local table?
* Also, in the current view, we have  src_vni=2000 (local) but then show it as VNI(2000) in the remote table. We should clarify / unify the vpcdiscriminant / vni at some point.

qmonnet:
I haven't really reviewed the PR, but I saw this commit:

```
fix(acl-filter): file each ACL rule by its own IP version, not the peering's

A manifest may hold exposes of both IP versions. `ValidatedManifest::is_v4`
is `any`, not `all`, so such a manifest reports `is_v4()` and `is_v6()` at
once, and `validate_ip_version` -- which compares only `is_v4()` -- accepts
the peering. [...]
```

A `ValidatedManifest` only has expose blocks of _one_ IP version (unless it is only made of a single “default” expose - see `validate_ip_version()` in `config`, and manifest validation). So there's no need to look at individual rules' IP version, we know that it's always the same for the peering. Based on that, I'm not sure this commit is relevant at all.

qmonnet:
The follow-up commit relies on a similar assumption:

```
fix(flow-filter): a catch-all expose covers every version its peering carries

A default expose names no address at all, so the lowering had to pick the
root prefix's IP version from somewhere -- and it picked
`ValidatedPeering::is_v4`. That is `any`, not `all`: a manifest holding one
expose of each version reports `is_v4()` and `is_v6()` at once, and
`validate_ip_version` accepts the peering.
```

But unless we have a bug, we ensure that validated manifests only contain one IP version so the difference between `any` and `all` of the validated expose blocks is not significant, and we should be able to rely on the IP version for the peering to determine the version to use.

flow-filter/src/lib.rs
qmonnet: +1, I'm also struggling to understand these changes

acl/tests/eal_install_classify.rs
daniel-noland: Three reasons:

1. I was avoiding test dependency on net (can create cycles)
2. because these tests are intended for illustration as much as testing
3. DPDK requires you to have at least one single byte field (which may be an ignored wildcard but still needs to exist)

The idea was to show how to use the library here and IpProto is a really simple example.

net/src/fixed_size.rs
```diff
@@ -93,6 +102,14 @@ mod tests {
         assert_eq!(buf, [10, 0, 1, 2]);
     }
 
+    /// The mask constants are bit patterns, so they are pinned as bytes: `ALL_BITS` must be every
```
daniel-noland: this test is somewhat absurd IMO.  Should remove realistically

flow-filter/src/context/tests.rs
```diff
@@ -757,3 +754,136 @@ fn reference_and_dpdk_backends_agree() {
         assert_eq!(ref_out[i], single, "batched != single at index {i}");
     }
 }
+
+/// The CLI dump is produced from the retained (typed) rules, not from the classifier, so the
+/// opaque production backend must render exactly what the reference oracle renders. If these ever
+/// diverge, the rendering has started depending on the backend again -- which is the state that
```
daniel-noland: ???

acl-filter/src/context.rs
```diff
@@ -308,58 +311,73 @@ pub(super) enum Backend {
...
+        &self.rules
     }
 }
 
 impl<K: MatchKey, A> fmt::Debug for AnyTable<K, A> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let kind = match self {
-            AnyTable::Empty => "empty",
-            AnyTable::Dpdk(_) => "dpdk",
+        let kind = match self.classifier {
```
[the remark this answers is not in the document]
daniel-noland: not true.  This does in fact compile just fine.

daniel-noland:
> I haven't really reviewed the PR, but I saw this commit:
> 
> ```
> fix(acl-filter): file each ACL rule by its own IP version, not the peering's
> 
> A manifest may hold exposes of both IP versions. `ValidatedManifest::is_v4`
> is `any`, not `all`, so such a manifest reports `is_v4()` and `is_v6()` at
> once, and `validate_ip_version` -- which compares only `is_v4()` -- accepts
> the peering. [...]
> ```
> 
> A `ValidatedManifest` only has expose blocks of _one_ IP version (unless it is only made of a single “default” expose - see `validate_ip_version()` in `config`, and manifest validation). So there's no need to look at individual rules' IP version, we know that it's always the same for the peering. Based on that, I'm not sure this commit is relevant at all.

I think you reviewed some draft code which isn't properly part of this PR.  Sorry for confusion

flow-filter/src/context/tables.rs
daniel-noland: I think you were reviewing scratch code from before I split and refactored all of that.  This isn't part of this PR.

flow-filter/src/lib.rs
daniel-noland: I think you were reviewing scratch code from before I split and refactored all of that.  This isn't part of this PR

daniel-noland:
Example flow filter CLI print

```
dataplane# show flow-filter table
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Routing context (flow filter) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Remote v4 (destination -> dst VPC + dst NAT):
  rank  proto  src-vni  destination  dst-port  |  to        NAT
  [0]   TCP    100      80.0.0.5/32  2222      |  VNI(200)  port-forwarding
  [1]   *      100      90.0.0.0/24  *         |  VNI(200)  -
  [2]   *      100      70.0.0.0/24  *         |  VNI(200)  masquerade
  [3]   *      200      10.0.0.0/24  *         |  VNI(100)  -
Local v4 (source -> src NAT):
  rank  proto  src-vni  dst-vni  source           src-port  |  NAT
  [0]   *      100      200      10.0.0.0/24      *         |  -
  [1]   *      200      100      90.0.0.0/24      *         |  -
  [2]   *      200      100      192.168.70.0/24  *         |  masquerade
Remote v6 (destination -> dst VPC + dst NAT):
  (no rules)
Local v6 (source -> src NAT):
  (no rules)
```

>[!NOTE]
> I didn't use the exact column names suggested by @Fredi-raspall because I think conflating vpc with vni is quite confusing

qmonnet (approved):
The last commit is large and would be easier to review as smaller chunks (although it's probably not worth splitting, now).

Looks good to me, thank you! It will be super helpful to be able to dump these tables.

Fredi-raspall (approved):

PR #1680, opened 2026-07-31
test(flow-filter): Add non-regression test

qmonnet:
As described in the test's comment, we recently discovered a bug in the previous flow-filter implementation. The new version is not affected, but let's introduce a non-regression test to be sure of it.

### DIFF
--- a/flow-filter/src/context/tests.rs
+++ b/flow-filter/src/context/tests.rs
@@ -498,6 +498,87 @@ fn ipv6_lookup() {
 }
 
 // -------------------------------------------------------------------------------------------------
+// Non-regression test: a prior implementation of the flow-filter stage used to have a bug. For a
+// setup with:
+//
+// - vpc2 and vpc3 exposing overlapping prefixes to vpc1, and
+// - vpc3's exposed prefixes being contained within vpc2's exposed prefixes to vpc1, and
+// - vpc2's exposed prefixes containing contiguous prefixes that could be merged into a single
+//   parent prefix,
+//
+// then we would merge entries in the temporary list of overlapping prefixes used to compute table
+// entries, resulting in a discrepency between (merged) prefixes for vpc2's context and (split)
+// prefixes for vpc3's context. In our example here, the context table would contain, for source
+// vpc1, among other entries, one entry for 10.0.2.0/31 (dst: vpc2), and additional entries for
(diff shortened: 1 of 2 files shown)


flow-filter/src/context/tests.rs
```diff
@@ -497,6 +497,87 @@ fn ipv6_lookup() {
...
+    for rule in rules {
+        let dst_prefix = rule.fields()[2].as_prefix().unwrap();
+        assert_ne!(dst_prefix.1, 32);
+    }
+    // Check there are no /32 prefixes in the local-side rules
+    let rules = ctx.local_v4.reference_rules().unwrap();
+    for rule in rules {
+        let src_prefix = rule.fields()[3].as_prefix().unwrap();
+        assert_ne!(src_prefix.1, 32);
+    }
```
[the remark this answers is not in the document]
qmonnet: I could, but I'm not sure it's worth it. It's hard to tell what checks should be done to detect the regression _exactly_. I could check the whole table contents to make sure that it corresponds to today's expectations, too, although it would break easily on table structure changes.

flow-filter/src/test_utils.rs
```diff
@@ -58,6 +58,15 @@ pub(crate) fn expose(ip: &str) -> VpcExpose {
     VpcExpose::empty().ip(ip.into())
 }
 
+/// Plain expose (no NAT): the listed prefixes are both private and public.
+pub(crate) fn expose_multi(ips: &[&str]) -> VpcExpose {
```
Fredi-raspall, line 62: #[cfg(test)] ?

Fredi-raspall (approved):
I can't say I like the style of AI-assisted written tests in this crate. I believe they are very verbose and add tons of layers that are hard to read and maintain; but that falls outside the scope of this PR.

flow-filter/src/test_utils.rs
qmonnet: No need, the whole module is already under `#![cfg(test)]`

qmonnet:
> I can't say I like the style of AI-assisted written tests in this crate. I believe they are very verbose and add tons of layers that are hard to read and maintain; but that falls outside the scope of this PR.

I found them less verbose than the tests for the previous version of the crate, but I'm open to suggestions to improve them in the future.

Thanks for the review!

PR #1687, opened 2026-08-04
test(flow-filter): carry the config oracle through to the NF, and widen the generators

daniel-noland:
Fuzz tests for flow filter.

I don't love everything about these tests (I wish the typing in particular were cleaner), but this is the version which doesn't require refactors outside of the tests themselves.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,6 +1243,7 @@ dependencies = [
 name = "dataplane-acl-filter"
 version = "0.24.0"
 dependencies = [
+ "bolero",
  "dataplane-acl",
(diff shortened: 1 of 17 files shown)


qmonnet:
Note: It may be worth waiting for the flow-filter fixes to make it first before merging this PR

mvachhar (approved):

qmonnet:
Removing from merge queue, I think it's worth waiting for the flow-filter fixes first (but open to discussion)

qmonnet:
We agreed to merge the fuzz tests first

PR #1688, opened 2026-08-04
fix(config,acl-filter): a manifest may not mix IP versions; refuse rules that cannot be lowered

daniel-noland:
Similar to #1687, this is more testing logic and some fixes for the acl-filter.

Again, I don't love the structure of these fuzzers, but this is the version which doesn't require extra refactor of the production logic.  The types should be reworked later so that we compose better and to promote reuse.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,6 +1243,7 @@ dependencies = [
 name = "dataplane-acl-filter"
 version = "0.24.0"
 dependencies = [
+ "bolero",
  "dataplane-acl",
  "dataplane-common",
  "dataplane-concurrency",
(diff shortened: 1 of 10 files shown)


qmonnet (approved):
Good catch for the IP version fix, thank you!

Not sure whether the 3rd commit is necessary, but I've got no objection.

qmonnet:
Does it need to be on top of the #1687? It seems the two PRs are entirely distinct? I'd send to merge queue if this one was based on `main`

mvachhar (commented):
This LGTM, but I want to confirm that we decided the fix for the IPv4 vs. IPv6 ACL any match issue was to reject (for now) manifests that have both IPv4 and IPv6 in the same peering?

qmonnet:
> the fix for the IPv4 vs. IPv6 ACL any match issue

@mvachhar Sorry, what issue is that, exactly?

PR #1689, opened 2026-08-04
fix(flow-filter): verify a masquerade destination, do not resolve it

daniel-noland:
ai generated, undergoing extra review by human coauthor, don't review yet


mvachhar:
Superceded by #1691

PR #1696, opened 2026-08-05
(1)-masquerade: allocate each public pair once

daniel-noland:
Masquerade hands out a public `(address, port)` pair, and return traffic is matched on a
reverse flow key built from that pair, the remote endpoint and the peer VPC. Nothing in that
key says which VPC the traffic came from, so two flows that collide on it are
indistinguishable and one tenant's return traffic can be delivered into another's VPC.

This branch fixes three ways the allocator could produce such a collision, and builds the
test machinery to show it does not.

- **Pools identified by their public range.** Exposes sharing a public range each got their
  own allocator, so each believed it owned the whole range.
- **Pools keyed by source VPC.** A private address only means something inside its own VPC.
  Two VPCs using the same private space produced one key, and the second silently replaced
  the first.
- **Allocation from disjoint regions.** Public ranges may overlap only partially, where
  neither contains the other. The public space is now cut at every point where the set of
  exposes covering it changes, giving maximal intervals with one allocator each.

It also carries, as its first commit, the `set_bitmap_value` fix that used to sit in #1699:
freeing a port was an OR with zero and never freed, and the guard against reserving a port already
taken compared the wrong two values, so a pair another flow held could be handed out again. That is
a fourth collision vector, and it belongs here because everything above depends on it -- in
particular the concurrent model checker, which cannot assert that a held pair is refused until the
guard works. Its pool-level test needs machinery introduced later and stays in #1699.

Fixing the bitmap turned a dormant hazard into a live one, so that is fixed here too: an
`AllocatedPort` frees its port on drop and was also `Clone`, so every copy freed the same port.
Dropping a copy released a pair its original still held, and the allocator would hand that pair to
a second flow. Dormant only because a drop that does nothing is harmless to repeat. `Clone` comes
off the allocation, and off `Allocation` and `MasqueradeState` with it -- the data plane never
wanted it, and the one caller was a test helper cloning live state to read two fields.

Then the testing those rest on: a bolero x model-checker suite over a config change, `just`
recipes for real libfuzzer campaigns, and two follow-on fixes found while writing them (a
non-exhaustion error being masked by region fallback, and a panic on an IPv6 address a pool
cannot index).

Every commit passes `cargo nextest run` on its own.

## The allocator can deadlock against its own pool lock

Moved here from the top of the stack. It predates the stack and depends on nothing in it, and at
the top it left every PR below it running with the hang -- which is not hypothetical: it wedged a
thread-sanitizer CI job for six hours before it was understood.

The pool holds weak references to the addresses in use; the strong ones belong to the blocks handed
out from each. Upgrading one under the pool guard and then letting it go can run
`AllocatedIp::drop` on the same thread, and that takes the same lock for writing. Another thread
ending the last flow on an address at that moment is all it takes. Four sites did it:
`reuse_allocated_ip`, `cleanup`, `reserve_from_pool`, and -- found by review of this PR, after the
first three were fixed -- `IpAllocator::fmt`, which reaches it from the management side, since
`NatAllocator` is a `CliSource` and the table is formatted on its own thread.

Each now keeps every upgrade alive until the guard is gone. Two Shuttle regressions come with it;
both deadlock in a single execution without the fix, and Shuttle names it directly: "tried to
acquire a `RwLock` it already holds".

### DIFF
--- a/nat/src/masquerade/apalloc/alloc.rs
+++ b/nat/src/masquerade/apalloc/alloc.rs
@@ -137,10 +137,17 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
         // FIXME: Should we clean up every time??
         self.cleanup_used_ips();
 
-        if let Ok(port) = self.reuse_allocated_ip(allow_null) {
-            return Ok(port);
+        // Drawing a fresh address is what to do when the addresses already in hand have no room,
+        // and only then. `reuse_allocated_ip` distinguishes the two: it walks past an address that
+        // has run out and reports anything else as it found it. Taking only `Ok` here would put
+        // that back, burying an error about the allocator under whatever the fresh address
+        // returns -- the same mistake `PoolSet::allocate` avoids one level up, where trying the
+        // next region on any error would bury it under a success.
+        match self.reuse_allocated_ip(allow_null) {
+            Ok(port) => Ok(port),
+            Err(e) if e.is_exhaustion() => self.allocate_from_new_ip(allow_null),
+            Err(e) => Err(e),
         }
-        self.allocate_from_new_ip(allow_null)
     }
 
     fn get_allocated_ip(&self, ip: I) -> Result<Arc<AllocatedIp<I>>, AllocatorError> {
@@ -227,21 +234,26 @@ impl<I: NatIpWithBitmap> PoolSet<I> {
 
     /// Allocate from the first region with room. Regions are ordered so that those the expose has
     /// to itself are tried first, leaving shared space for exposes that have nowhere else to go.
+    ///
+    /// Only a region being full moves on to the next one. Any other error is about the allocator
+    /// rather than about how full that region is, and trying the next region would either bury it
+    /// under a success or replace it with a later region's `NoFreeIp`.
     pub(crate) fn allocate(
         &self,
         allow_null: bool,
     ) -> Result<port_alloc::AllocatedPort<I>, AllocatorError> {
-        let mut last_error = None;
+        let mut exhausted = None;
         for region in &self.regions {
             match region.allocator.allocate(allow_null) {
                 Ok(port) => return Ok(port),
-                Err(e) => {
-                    debug!("Region {:?} could not allocate: {e}", region.range);
-                    last_error = Some(e);
+                Err(e) if e.is_exhaustion() => {
+                    debug!("Region {:?} is out of space: {e}", region.range);
+                    exhausted = Some(e);
                 }
+                Err(e) => return Err(e),
             }
         }
-        Err(last_error.unwrap_or(AllocatorError::NoFreeIp))
+        Err(exhausted.unwrap_or(AllocatorError::NoFreeIp))
     }
 
     /// Reserve a specific address and port, which has to come from the region owning that address.
--- a/nat/src/masquerade/apalloc/alloc.rs
+++ b/nat/src/masquerade/apalloc/alloc.rs
@@ -22,7 +22,7 @@ use roaring::RoaringBitmap;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::net::{IpAddr, Ipv6Addr};
 use std::time::Duration;
-use tracing::debug;
+use tracing::{debug, error};
 
 ///////////////////////////////////////////////////////////////////////////////
 // IpAllocator
@@ -467,8 +467,15 @@ impl<I: NatIpWithBitmap> NatPool<I> {
 
     fn deallocate_from_pool(&mut self, ip: I) {
         debug!("Address {ip} was deallocated");
-        let offset = I::try_to_offset(ip, &self.reverse_bitmap_mapping).unwrap();
-        self.bitmap.set_ip_free(offset);
+        // The address was handed out by this pool, so it maps back into it. This runs while an
+        // allocation is being dropped and has nowhere to report a failure, so say so and leave the
+        // address marked in use rather than panicking on the drop path.
+        match I::try_to_offset(ip, &self.reverse_bitmap_mapping) {
+            Ok(offset) => {
+                self.bitmap.set_ip_free(offset);
+            }
+            Err(e) => error!("Address {ip} does not map back into the pool it came from: {e}"),
+        }
     }
 
     /// `keep_alive` collects every address upgraded here, for the caller to release once the pool
@@ -626,11 +633,25 @@ pub(crate) fn map_offset(
 }
 
 // Reverse operation from map_offset()
-pub(crate) fn map_address(address: Ipv6Addr, bitmap_mapping: &BTreeMap<u128, u32>) -> u32 {
+//
+// Not every address of a region has an offset. A region may hold more addresses than a u32 can
+// index, in which case the bitmap covers only the first 2^32 of them (see `NatPool::for_range`),
+// and an address beyond that is simply not one this pool can serve. That is reachable from
+// configuration rather than a bug: a flow carried across a config change presents the address it
+// already holds, and the region it falls in may have grown downwards underneath it. Report it as
+// the pool not serving the address, so the flow is dropped like any other that cannot be carried
+// over, rather than panicking in the middle of applying a config.
+pub(crate) fn map_address(
+    address: Ipv6Addr,
+    bitmap_mapping: &BTreeMap<u128, u32>,
+) -> Result<u32, AllocatorError> {
     let (prefix_start_bits, prefix_offset) = bitmap_mapping
         .range(..=address.to_bits())
         .next_back()
-        .expect("This should never fail");
+        .ok_or(AllocatorError::NoPoolFound)?;
 
-    prefix_offset + u32::try_from(address.to_bits() - prefix_start_bits).unwrap()
+    u32::try_from(address.to_bits() - prefix_start_bits)
+        .ok()
+        .and_then(|offset| prefix_offset.checked_add(offset))
+        .ok_or(AllocatorError::NoPoolFound)
 }
--- a/nat/src/masquerade/apalloc/alloc.rs
+++ b/nat/src/masquerade/apalloc/alloc.rs
@@ -9,14 +9,15 @@
 //!
 //! See also the architecture diagram at the top of mod.rs.
 
+use super::region::AddrInterval;
 use super::{NatIpWithBitmap, port_alloc};
 use crate::masquerade::allocation::AllocatorError;
 use crate::masquerade::natip::NatIp;
 use crate::port::NatPort;
 use crate::ranges::IpRange;
 use concurrency::sync::{Arc, RwLock, RwLockReadGuard, Weak};
+use lpm::prefix::PortRange;
 use lpm::prefix::range_map::DisjointRangesBTreeMap;
-use lpm::prefix::{IpPrefix, PortRange, Prefix};
 use roaring::RoaringBitmap;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use std::net::{IpAddr, Ipv6Addr};
@@ -47,10 +48,6 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
 
     pub(crate) fn read(&self) -> RwLockReadGuard<'_, NatPool<I>> {
         self.pool.read()
-    }
-
-    pub(crate) fn idle_timeout(&self) -> Duration {
-        self.pool.read().idle_timeout()
     }
 
     fn deallocate_ip(&self, ip: I) {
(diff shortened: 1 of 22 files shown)


nat/src/masquerade/apalloc/setup.rs
```diff
@@ -42,12 +45,153 @@ impl NatAllocator {
...
+                allocator: allocator.clone(),
+                policy,
+            },
+        );
+        allocator
+    }
+
+    // Pools are shared only when their public ranges match exactly. Ranges that merely overlap
+    // still end up with one allocator each, neither aware of what the other hands out, so report
+    // them: that is a configuration we cannot serve correctly.
```
daniel-noland, line 162: fixed?

nat/src/masquerade/apalloc/port_alloc.rs
```diff
@@ -845,27 +845,36 @@ impl Bitmap256 {
...
-            if self.first_half & (1 << port_in_block) == value {
-                return Err(());
-            }
-            self.first_half |= value << port_in_block;
+    /// Mark a port used or free, failing if it is already in that state.
+    ///
+    /// The error is load-bearing in both directions: it is how reserving a port that has already
+    /// been handed out is refused, rather than the pair going to two flows at once, and how giving
+    /// back a port that was never taken is reported as the bookkeeping mistake it is.
+    fn set_bitmap_value(&mut self, port_in_block: u8, used: bool) -> Result<(), ()> {
```
Fredi-raspall, line 853: This probably explains the exhaustion of ports observed. The code is correct now.

mvachhar: Agreed.  It is at least one explanation.

nat/src/masquerade/apalloc/port_alloc.rs
```diff
@@ -845,27 +845,36 @@ impl Bitmap256 {
...
-        if port_in_block < 128 {
-            if self.first_half & (1 << port_in_block) == value {
-                return Err(());
-            }
-            self.first_half |= value << port_in_block;
+    /// Mark a port used or free, failing if it is already in that state.
+    ///
+    /// The error is load-bearing in both directions: it is how reserving a port that has already
+    /// been handed out is refused, rather than the pair going to two flows at once, and how giving
+    /// back a port that was never taken is reported as the bookkeeping mistake it is.
```
mvachhar, line 852: The lower part of this comment should be removed before merge.

nat/src/masquerade/apalloc/test_alloc.rs
```diff
@@ -133,6 +139,67 @@ mod context {
         let config = MasqueradeConfig::new(&vpc_table, 1);
         NatAllocator::new(config)
     }
+
+    // Two *different* VPCs, each peering with the same destination VPC, and each masquerading
```
mvachhar, line 143: We should look at these comments and fix them, they makes it seem like two VPCs sharing a pool is an error, but it is not, it is normal behavior and should be framed as such.

nat/src/masquerade/apalloc/mod.rs
```diff
@@ -144,18 +159,19 @@ impl<I: NatIpWithBitmap, J: NatIpWithBitmap> PoolTable<I, J> {
...
+        let key = PoolTableKey::new(protocol, src_vpcd, dst_vpcd, addr, max_range::<I>());
         self.get(&key)
     }
 
     fn add_entry(&mut self, key: PoolTableKey<I>, allocator: alloc::IpAllocator<J>) {
         if self.0.contains_key(&key) {
             warn!(
-                "Overwriting NAT pool entry {key:?}: the same private prefix is masqueraded by \
-                 more than one expose towards the same peer VPC"
+                "Overwriting NAT pool entry {key:?}: within one VPC, the same private prefix is \
```
mvachhar, line 173: Again, config validation should prevent this from ever happening right?

nat/src/masquerade/apalloc/test_alloc.rs
```diff
@@ -503,6 +573,46 @@ mod std_tests {
...
+        let from_vpc1 = allocator
+            .allocate_v4(vpcd1(), vpcd3(), addr_v4("192.168.0.1"), NextHeader::TCP)
+            .unwrap();
+        let from_vpc2 = allocator
+            .allocate_v4(vpcd2(), vpcd3(), addr_v4("192.168.0.1"), NextHeader::TCP)
+            .unwrap();
+
+        assert_eq!(
+            from_vpc1.allocation.ip(),
+            addr_v4("10.1.0.0"),
```
mvachhar, line 593: Based on the comment, should both translation be using the address here for the best test?  Or at least from the same prefix, I don't think the manifest has 10.0.0.0/8 as the IP range does it?

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -140,6 +137,94 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
...
+    pub(crate) fn idle_timeout(&self) -> Duration {
+        self.idle_timeout
+    }
+
+    pub(crate) fn regions(&self) -> impl Iterator<Item = &PoolRegion<I>> {
+        self.regions.iter()
+    }
+
+    /// Allocate from the first region with room. Regions are ordered so that those the expose has
+    /// to itself are tried first, leaving shared space for exposes that have nowhere else to go.
```
mvachhar, line 194: Eventually this logic will have to be augmented to try and reuse the last used public IP for a given private IP.  Perhaps we should add a comment to this effect?

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -235,26 +320,43 @@ pub(crate) struct NatPool<I: NatIpWithBitmap> {
...
+        // Index the region from its own start. IPv4 indexes its bitmap by the address bits and
+        // ignores the mapping; IPv6 cannot fit its space in a u32, so indices count from the start
+        // of the region and the mapping carries them back to real addresses. Going through
+        // try_to_offset gets both right without naming either version here.
+        let bitmap_mapping = BTreeMap::from([(0u32, range.start)]);
+        let reverse_bitmap_mapping = BTreeMap::from([(range.start, 0u32)]);
+
+        // A region holding more addresses than the u32 bitmap can index is truncated. We would run
+        // out of memory long before allocating four billion addresses.
+        let span = range.len().saturating_sub(1).min(u128::from(u32::MAX));
```
mvachhar, line 345: "640k should be enough for everyone" - not Bill Gates circa 1981

nat/src/masquerade/apalloc/mod.rs
```diff
@@ -54,6 +62,15 @@
 //! Returned object
 //! ```
 //!
+//! The layer worth reading twice is [`PoolSet`](alloc::PoolSet) and
```
mvachhar, line 65: This comment is a little confusing since it talks about distinguishing return traffic but doesn't explain that that is why you need to guarantee a unique allocation for the public address, that way the flow keys are never the same.  Please reword.

nat/src/masquerade/apalloc/setup.rs
```diff
@@ -1,287 +1,284 @@
...
-    }
-
-    // Pools are shared only when their public ranges match exactly. Ranges that merely overlap
-    // still end up with one allocator each, neither aware of what the other hands out, so report
-    // them: that is a configuration we cannot serve correctly.
-    fn report_partial_overlaps(&self, identity: &PoolIdentity) {
-        for existing in self.pools.keys() {
-            if existing.protocol != identity.protocol || existing.dst_vpc_id != identity.dst_vpc_id
-            {
+// Collect the masquerade exposes of every peering, grouped by the VPC they masquerade towards.
```
mvachhar, line 84: We have an interesting case here if two remote peers are externals that route the same address space.  Not sure how we would handle that at a system level so we probably don't need to address it yet, but it is worth mentioning.  It might be the case that we should treat all addresses to/from externals to be the same address space for NAT purposes.  It would cause less surprises.  If other reviewers agree, we should file an issue and put it in 26.04 for triaging.

nat/src/masquerade/apalloc/setup.rs
```diff
@@ -1,287 +1,284 @@
...
-    v4: PoolRegistry<Ipv4Addr>,
-    v6: PoolRegistry<Ipv6Addr>,
+// Convert a set of public prefixes into raw address intervals, dropping any that are not of the
+// version being built.
+fn public_intervals<J: NatIp>(ranges: &PrefixPortsSet) -> Vec<AddrInterval> {
+    ranges
+        .iter()
+        .filter_map(|prefix| {
+            // FIXME: Account for port ranges. A public range may be restricted to a port range,
+            // which the pools do not model, so the whole port space of the address is used.
```
mvachhar, line 138: Are we going to fix this in this PR?

nat/src/masquerade/apalloc/mod.rs
mvachhar, line 62: Please rewrite the commit comment for "refactor(masquerade): Find a pool without relying on disjoint prefixes", it is really hard to understand.

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -58,25 +58,46 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
...
-            if !ip.has_free_ports() {
-                continue;
-            }
-            match ip.allocate_port_for_ip(allow_null) {
-                Ok(port) => {
-                    debug!("Allocated port {port}");
-                    return Ok(port);
+        // An address upgraded out of the in-use list has to outlive the guard below.
+        //
+        // The list holds weak references; the strong ones belong to the blocks handed out from
```
mvachhar, line 63: Ok, this explains the deadlock entirely.  And it really sucks, it shows a structural problem that we have to fix.  This is fine for now, but ideally we'd have a structure that someone elides the reader writer lock or uses it in a way that is less tricky than this.  Good catch.

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -115,9 +144,15 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
...
-            .write()
-            .reserve_from_pool(ip, self.clone(), self.randomize)
+        // The third place that upgrades an in-use entry under the pool lock, and so the third that
+        // must not let the upgrade go while holding it. See `cleanup_used_ips`.
+        let mut examined = Vec::new();
+        let outcome =
+            self.pool
+                .write()
+                .reserve_from_pool(ip, self.clone(), self.randomize, &mut examined);
+        drop(examined);
```
mvachhar, line 154: This really makes me sad, it destroys the elegance of the automatic drop, but here we are.

mvachhar (changes requested):
We need to fix a few of the comments, I also have a few questions about the tests and some code.

Also, do we need to fix the port range overlap in this PR?  Do we allow specification of port ranges for expose pools?  I don't think we do, but cannot recall.

nat/src/masquerade/apalloc/mod.rs
```diff
@@ -144,18 +159,19 @@ impl<I: NatIpWithBitmap, J: NatIpWithBitmap> PoolTable<I, J> {
...
         self.get(&key)
     }
 
     fn add_entry(&mut self, key: PoolTableKey<I>, allocator: alloc::IpAllocator<J>) {
         if self.0.contains_key(&key) {
             warn!(
-                "Overwriting NAT pool entry {key:?}: the same private prefix is masqueraded by \
-                 more than one expose towards the same peer VPC"
+                "Overwriting NAT pool entry {key:?}: within one VPC, the same private prefix is \
+                 masqueraded by more than one expose towards the same peer VPC"
```
qmonnet, line 174: This comment doesn't look right; within one VPC, towards a given peer VPC, you should never have masquerade overlap? If we had several `expose` blocks with masquerade and with the same private prefix, both towards the same destination VPC, we wouldn't be able to tell what `expose` rule to pick and what pool to use I think. Maybe add a `debug_assert` to make sure we catch this if it ever happens?

qmonnet (approved):
I've been through the commits and most of the code changes, with more or less depth, though. Even knowing the original code I can't provide a confident opinion about the full set, but overall, the introduction of pools of public prefixes, connected to the internal pools keyed on source prefixes, and the split of overlapping prefixes (masquerade) into separate regions, sounds good. I also saw multiple bug fixes that we certainly do want to merge - thanks for these!! The changes come with extended test coverage and I could find no blocker, so I'm good with the PR.

qmonnet:
@daniel-noland Please note the merge conflict to address

daniel-noland:
> @daniel-noland Please note the merge conflict to address

on it

Fredi-raspall:
@daniel-noland , nice work, it fixes 2 critical bugs at least. There are some changes that I'm not sure I understand yet.
Sorry for the conflict, I think a recent fix from me caused it. Let me know if you need help on the rebase. Should not be difficult, though.

mvachhar:
Superceded by #1707

PR #1697, opened 2026-08-06
(4)-masquerade: find a pool without relying on disjoint prefixes

daniel-noland:
> [!WARNING]
> **AI assisted, not yet ready for external review by other humans.**
> Please do not spend review time on this yet. It is pushed to run CI and to keep
> the stack visible, not to attract review. The `dont-merge` label stays on until
> that changes.

**This fixes no live bug, and is labelled `refactor` for that reason.**

Looking a pool up walked back to the entry nearest below the private address and took it if
its prefix reached far enough. That is only right when the prefixes under one protocol and
pair of VPCs are disjoint. A prefix nested inside another starts nearer to an address while
covering less of it, so for an address of the wider prefix *above* the nested one the walk
stopped on the nested prefix, found it too short, and reported no pool at all. Addresses below
the nested prefix were served correctly, since the walk starts at the address and never reaches
a prefix beginning above it.

Three separate things rule that out for a validated configuration: `VpcExpose::validate`
normalizes a prefix set into disjoint prefixes, `validate_expose_collisions` rejects any
overlap between two masquerade exposes of a manifest, and `check_peering_count` refuses a
second peering between one pair of VPCs.

It is still worth not depending on them. Those guarantees live in another crate and nothing
near the lookup says the lookup rests on them, which is the distant reasoning
`development/code/avoid-global-reasoning.md` asks us to design out. Within this crate the
invariant is not enforced at all — the fuzz and unit harnesses build a `PoolTable` directly.
And the way it failed was misleading: it dropped the packet and logged that the allocator had
a bug, when the configuration was the unusual part.

Covered by a property test against a brute-force oracle over deliberately nesting-heavy inputs.
The oracle is an *interval* one -- nearest start, then narrowest -- rather than longest-prefix:
the generator produces intervals of any offset and length, most of them not CIDR-aligned, and
longest-prefix match is only defined on prefixes. The rule asserted agrees with longest-prefix on
everything the configuration layer can produce and is defined on what it cannot.

Also here, after review:

- **The lookup's doc corrected**, folded down from #1700 so this PR does not make the two claims
  it fixes. A nested prefix hid addresses *above* it, not either side; and overlaps are not
  "reported when the table is built" -- `add_entry` warns only on two entries with identical
  bounds, so a nesting or a partial overlap passes without a word.
- **The group-isolation test now crosses a boundary.** Keys order by protocol, then source VPC,
  then destination VPC, before the address, and the walk only looks back -- so the foreign group
  it used sorted *after* the queried one and was excluded by the range bound before the guard ran.
  The guard could be deleted outright and all five tests still passed. The test now puts a group
  below the queried one on each of the three components in turn, and deleting the guard fails it.

## Stack

Merge bottom to top. Each PR is based on the one above it in this list.

- #1696 — (1)-masquerade: allocator collision fixes
- #1698 — (2)-masquerade: port forwarding reservation fixes
- #1699 — (3)-masquerade: make the concurrent oracle honest
- #1697 — (4)-masquerade: find a pool without relying on disjoint prefixes **← you are here**
- #1700 — (5)-masquerade: allocator coverage, and the bugs it found

Every commit in the stack passes `cargo nextest run` on its own.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 **.profraw
 **/__fuzz__/**
+# libfuzzer writes one of these per worker into the working directory when
+# `just fuzz` is given -j; the corpus itself lives under __fuzz__.
+fuzz-*.log
 # qemu-user core dumps from SIGABRT under emulated tests.
 **/qemu_*.core
 result*
--- a/development/code/running-tests.md
+++ b/development/code/running-tests.md
@@ -62,8 +62,75 @@ change this.
 The major downside is that these processes are very computationally intensive and can take a long time to run.
 In fact, the [afl] fuzzer runs until you terminate it.
 
+## Running a real fuzzing campaign
+
+To run a target under [libfuzzer], which is coverage guided and explores far deeper than the random
+driver the test suite uses, list the targets and pick one:
+
+```shell
+just fuzz-list -p dataplane-nat
+just fuzz 'masquerade::apalloc::region::bolero_tests::decompose_properties' 10min -p dataplane-nat
+```
+
+The duration defaults to `60s`; anything after it is forwarded to `cargo bolero test`. As a sense of
+the difference, a property that manages a few thousand cases per second under `just test` reaches
+several hundred thousand per minute here, because libfuzzer mutates towards inputs that reach new
+code rather than sampling blindly.
+
+Findings are written to a `__fuzz__` directory beside the test. That directory is gitignored: the
+corpus is a local artifact that seeds later runs on the same machine, not something to commit.
+
+Pass `-j` to spread the campaign over more cores, which is the cheapest way to reach deeper:
+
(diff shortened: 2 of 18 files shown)


daniel-noland:
Folded into #1696 and closed as part of a review-friendliness pass over the stack.

This was a one-commit PR whose content — the pool lookup that no longer relies on private prefixes being disjoint — belongs next to the commits that build the pool table in the first place (`Identify NAT pools by their public range`, `Key NAT pools by source VPC as well`, `Allocate from disjoint regions of the public space`). It now sits directly after them in #1696 as `refactor(masquerade): Find a pool without relying on disjoint prefixes`, unchanged apart from the review fixes it already carried.

The stack is now four PRs: #1696 → #1698 → #1699 → #1700.

PR #1698, opened 2026-08-06
(2)-masquerade: port forwarding claims

daniel-noland:
Two defects in the ports masquerade must keep off because port forwarding has claimed them.
Both were reachable. The first was found by the property tests added in the previous PR, which
is why that PR carries two tests committed `#[ignore]`d. The second was found while writing
them rather than by them: the pools are handed claims already expressed in public space, so no
pool-level property can see a mistake in translating the configuration into those claims. It
took the config-level test added here.

- **Only one claim per address was honoured.** The limit was baked in three times over: a map
  keyed by address range where inserting twice overwrote, a pool resolving one range per
  address out of it, and a port allocator storing one range. Each had to change together.
- **Claims were computed in the wrong address space.** They came from intersecting *private*
  prefixes, while the pools are asked about *public* addresses, so effectively nothing was
  ever reserved. Claims are now gathered per peer VPC rather than per manifest, since the
  public space towards a peer is shared between every VPC masquerading onto it.

Verified by mutation: restoring the private prefixes makes the new end-to-end test hand out
the first forwarded port. Two test oracles were also strengthened here rather than higher in
the stack: the reserved-ports property now unions every claim as the implementation does,
instead of skipping claims whose own expose did not declare the address (the gated version
survived ten thousand inputs against a pool built from only each expose's own claims), and the
masquerade-over-forwarded-ports fixture is now the same-manifest configuration production
validation actually accepts, rather than a cross-VPC one it rejects.

## The lifecycle of a claimed block

Claiming ports gives the allocator three states it never had before -- a block claimed in full,
an address whose every block is claimed, and a run of such addresses -- and it mishandled all
three. These were found by review of this PR and the fixes have been moved down from #1700 to
sit with the change that creates the states, so this PR does not ship them:

- **A fully claimed first address disabled the whole region.** Allocation draws the lowest free
  address; if port forwarding had claimed every port on it, the attempt failed and the address
  went straight back into the pool for the next packet to find. Masquerade over `10.1.0.0/31`
  with forwarding over `10.1.0.0/32:1024-65535` could allocate nothing at all, though
  `10.1.0.1` had 64k ports free. Such addresses are now kept out of the pool as it is built.
- **A fully claimed block left the free-block count overstated.** The count was taken before
  claims were applied and never decremented when a block was skipped, so an address reported
  room it did not have, reuse failed on it, and every later flow drew a fresh address. With four
  addresses and one claim, the 260th allocation failed with `NoFreeIp` while 193,533 usable
  pairs sat untouched.
- **An ordinary policy conflict was reported as allocator corruption.** A skipped claimed block
  is not free and has no allocated block, so a reservation aimed at it returned `InternalIssue`
  rather than `PortReservationFailed`. The concurrency model treats `InternalIssue` as fatal,
  and a late carry-over can aim at exactly such a block.

Each was reproduced against this PR's previous head and passes now.

### DIFF
--- a/nat/src/masquerade/apalloc/alloc.rs
+++ b/nat/src/masquerade/apalloc/alloc.rs
@@ -30,7 +30,7 @@ use tracing::{debug, error};
 /// ordinary case, where the first address serves, and an address that cannot serve is taken out of
 /// the pool as it is found, so a run of them is worked through over successive packets rather than
 /// being walked again each time.
-const MAX_ADDRESSES_PER_ALLOCATION: usize = 8;
+pub(super) const MAX_ADDRESSES_PER_ALLOCATION: usize = 8;
 
 ///////////////////////////////////////////////////////////////////////////////
 // IpAllocator
@@ -410,6 +410,16 @@ impl<I: NatIpWithBitmap> Drop for AllocatedIp<I> {
 #[derive(Debug)]
 pub(crate) struct NatPool<I: NatIpWithBitmap> {
     bitmap: PoolBitmap,
+    /// Offsets that may never be handed out, however often they are given back.
+    ///
+    /// Distinct from being absent from `bitmap`, which only says an address is in use at the
+    /// moment. An address lands here either because port forwarding has claimed every port
+    /// masquerade could draw on it, which is known when the pool is built, or because it was found
+    /// to have nothing to give while serving. Either way it must not come back:
+    /// `deallocate_from_pool` runs on the drop path and would otherwise return it to the pool the
+    /// first time a flow holding it ends, which for a carried-over flow is moments after the pool
+    /// was built.
+    unusable: RoaringBitmap,
     bitmap_mapping: BTreeMap<u32, u128>,
     reverse_bitmap_mapping: BTreeMap<u128, u32>,
     in_use: VecDeque<Weak<AllocatedIp<I>>>,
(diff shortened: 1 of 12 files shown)


nat/src/masquerade/test.rs
```diff
@@ -329,6 +329,69 @@ fn build_overlay_2vpcs_modified() -> Overlay {
...
+    let mut peering_table = VpcPeeringTable::new();
+    peering_table.add(peering13).expect("Failed to add peering");
+    peering_table.add(peering23).expect("Failed to add peering");
+
+    Overlay::new(vpc_table, peering_table)
+}
+
+// A TCP packet towards VPC-3, from a given source VPC and private address. A SYN opens a flow;
+// anything else is only translated if one already exists.
+fn tcp_from(src_vni_id: u32, src_ip: &str, syn: bool) -> Packet<TestBuffer> {
```
daniel-noland, line 376: typing

nat/src/masquerade/nf.rs
```diff
@@ -69,10 +69,26 @@ pub struct Masquerade {
...
+    // A flow expires against the wall clock, while a test refreshes it by doing work: sending a
+    // packet. Under miri or qemu-user that work takes something like two orders of magnitude
+    // longer, so the gap between one packet and the next stops being a fraction of a flow's life
+    // and becomes several times it. A flow a native run keeps comfortably alive is one an emulated
+    // run finds long dead, and the test fails on a clock rather than on anything it meant to
+    // check. Stretching the timeouts keeps them measuring the same amount of work either way.
+    //
+    // `emulated` is set only by the miri and qemu-user paths, so a real data plane always gets the
+    // native values.
+    const TIMEOUT_SCALE: u64 = cfg_select! {
```
mvachhar, line 83: I hate this fix but in the interest of time we'll keep it.

nat/src/masquerade/apalloc/setup.rs
```diff
@@ -87,20 +110,28 @@ fn gather_exposes<'a, J, F, FIter, P, PIter>(
...
-            port_forwarding_exposes_filter(manifest).collect();
+        let group = groups.entry(nat_peering.dst_vpcd).or_default();
+
+        // Port forwarding claims public space towards this peer whichever VPC declared it, because
+        // the public space towards a peer is shared: return traffic carries nothing that says
+        // which VPC it belongs to. Claims are therefore collected per peer VPC rather than per
+        // manifest, so a claim made by one VPC is honoured by another VPC's pools, and a peering
+        // that port-forwards without masquerading still has its claims respected.
+        for pf_expose in port_forwarding_exposes_filter(manifest) {
+            group.claimed.add(pf_expose);
```
mvachhar, line 133: If a port-forward is destroyed, who removes the claims?

nat/src/masquerade/apalloc/concurrent_fuzz.rs
```diff
@@ -184,7 +184,8 @@ impl Published {
         generation: u64,
```
mvachhar, line 184: I am not sure that the claiming is needed for port forwards at all.  Consider a port forward from public address `1.1.1.1:80` to private address `10.10.10.1:8000`.  Now suppose that `10.10.10.1` initiates a connection to some 2.3.4.5:1234 and it gets masqueraded to `1.1.1.1:80`  The TCP stack for 2.3.4.5 has knowledge of the flow `1.1.1.1:80 <-> 2.3.4.5:1234`.  So, it will not choose source port `1234` to get to `1.1.1.1:80`, so reservation is not needed for this case.  The same applies vice versa.  If 2.3.4.5:1234 initiates a flow via the port forward to 1.1.1.1:80, this gets translated to 2.3.4.5:1234 <-> 10.10.10.1:8000.  `10.10.10.1` will not generate a new flow toward 2.3.4.5:1234 with source port 8000.

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -94,8 +94,11 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
                         outcome = Ok(port);
                         break;
                     }
-                    // If there is no free port left, loop again to try another IP address
-                    Err(AllocatorError::NoFreePort(_)) => {}
+                    // This address has nothing left, whether it ran out of ports in a block or of
```
mvachhar, line 97: Is this still needed if there is no reservation for port forwards?

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -437,16 +447,41 @@ impl<I: NatIpWithBitmap> NatPool<I> {
...
+        // and which addresses those are is known now rather than discovered a packet at a time.
+        // Left in the pool, such an address is drawn because it is the lowest one free, found
+        // useless, and taken out -- but only eight of them may be worked through before the
+        // allocation gives up, so a run of them costs a dropped packet for every eight, on flows
+        // the region had ample room for. And it costs them again after every config change, since
+        // a new allocator starts with a fresh pool. Keeping them out from the start costs one
+        // sweep over the claims when the pool is built.
+        let mut unusable = RoaringBitmap::new();
+        for interval in reserved_ports.unusable_within::<I>(indexable, exclude_wellknown_ports) {
+            let (first, last) = (to_offset(interval.start), to_offset(interval.end));
```
mvachhar, line 469: The allocator should have a method to reserve a particular port, not just the next port and when it gets resrved, a handle is returned and this handle should be held in the flow_info so that the port is freed when that flow dies, not just remove things from the bitmap, that is broken.

mvachhar (changes requested):
Please remove the unneeded port forward reservations and make sure we reserve ports for active port forward flows in the standard way by holding a handle in the `flow_info`.  But this will likely need a method on the allocator to allocate a particular port.

mvachhar:
Superceded by #1707

PR #1699, opened 2026-08-06
(3)-masquerade: the block-handover races, and honest concurrent oracles

daniel-noland:
The `set_bitmap_value` fix this PR used to carry has moved to the bottom of the stack, into
#1696. It had to: the concurrent model checker cannot assert that a pair already held is refused a
second time until that fix is in, and a suite that accepts such a success certifies the defect
instead of catching it. With it at the bottom, no PR in the stack ever ships a live
duplicate-reservation bug.

What is left here is the testing that fix makes possible.

## The concurrent suite now guards this fix

The model checker's reservation oracle moved here from further up the stack, because here is where
it can hold. It asserts that a second reservation of a pair a carried-over flow still holds is
refused -- which is exactly what the duplicate-reservation guard above makes true, and what was
false before it. Under the previous arrangement the suite accepted such a success and dropped it,
so it passed while certifying the very defect this PR fixes.

Mutation-verified against this PR's own change: revert the `set_bitmap_value` guard and
`stress_test_config_change` fails.

`Published::build` also stopped accepting carry-over failures. Every generation is built from the
same specs with a fresh allocator and distinct survivor pairs, so nothing may refuse one; tolerating
it let the model publish generations that carried nothing, and every property about carried pairs
passed vacuously over them.

## A race the model could not reach, found by review

The suite's own documentation claimed the reservation/release window in `find_block_for_port`
needed generations whose specs differ to reach. That was wrong, and the claim was made by this
PR. One pool and one generation reach it: allocate a pair, then race a reservation of that pair
against dropping its only holder. Shuttle finds the schedule in a single execution.

A block is given back in two steps a reader can fall between -- the free flag is stored `true`,
and the weak entry in the list of allocated blocks expires with the last `Arc`. A reservation
reads the flag and then searches the list, so it could see the block as neither, and returned
`InternalIssue`: the allocator declaring its own bookkeeping broken for a moment in which nothing
is wrong. The lookup now starts over, bounded, and answers a failed reservation if every attempt
lands in the window.

Production reaches this through the late flow `nf.rs` already handles, where a packet that
allocated from the previous allocator installs its flow after a new one is published. That pair
is not among the writer's pinned survivors, so the block behind it can be emptying as it
re-reserves. The cost was a spuriously invalidated flow during a config change.

The defect predates this stack -- the code carried a `FIXME` wondering whether the window was
reachable and noting it had not been seen under shuttle. It is fixed here rather than lower
because this is the PR that makes the concurrent oracle honest, and the claim it corrects is one
this PR made. The regression test fails on the first schedule without the fix.

What the suite still cannot reach on its own is unchanged and now stated correctly: its
reservations target survivors, whose blocks are pinned for the generation. Widening
`ReserveExisting` to unpinned pairs is the extension that would close it.

## A third instance of the same shape

Review of the whole stack found the same two-step handover a third time, and it is fixed here
alongside the other two. Tidying a dead entry out of the list of allocated blocks looks the entry
up and drops it under separate locks, so another thread can claim the freed block and list it at
that index in between; the drop then deletes an entry for a block that is in use.

Nothing is handed out twice, so this is availability rather than isolation — the orphaned block
stays claimed by its holder while the allocator no longer knows of it, so reservations into it are
refused and its free ports stop counting towards the address having room. Shuttle reproduces it and
the symptom is exactly that: `PortReservationFailed` on a block with 255 ports free.

The answer here is to re-check under the write lock rather than to retry, since the caller has a
lock to take anyway. `search_for_block` had the same double-upgrade shape and now keeps its first
one.

### DIFF
--- a/nat/src/masquerade/apalloc/concurrent_fuzz.rs
+++ b/nat/src/masquerade/apalloc/concurrent_fuzz.rs
@@ -34,13 +34,15 @@
 //! * Neither allocation nor reservation ever reports [`AllocatorError::InternalIssue`]. That is
 //!   the allocator saying its own bookkeeping is inconsistent.
 //!
-//!   A limit worth being honest about: `find_block_for_port` carries a standing `FIXME` wondering
-//!   whether the block it just found non-free can be released before it is looked up, and this
-//!   suite does not reach that interleaving. Reservations here target survivors, and a survivor's
-//!   block is pinned for the whole generation by the reservation [`Published`] holds, so it cannot
-//!   disappear mid-lookup. Reaching it would take generations whose specs differ, so that a pair
-//!   stops being carried and its block can empty while another thread reserves it; that is the
-//!   suite's next extension, not something it does today.
+//!   A limit worth being honest about: this suite does not reach the interleaving where the block
+//!   a reservation just found non-free is released before it is looked up. Reservations here
+//!   target survivors, and a survivor's block is pinned for the whole generation by the
+//!   reservation [`Published`] holds, so it cannot empty mid-lookup. That is a property of how
+//!   this suite reserves, not of what it takes to get there: neither a config change nor
+//!   generations whose specs differ is required, and
+//!   [`a_reservation_racing_the_release_of_its_block_is_not_called_a_bug`] reaches it with one
+//!   pool and one generation. Widening `ReserveExisting` to unpinned pairs would let this suite
+//!   reach it too, and is the extension worth making.
 //!
 //! # No loom
 //!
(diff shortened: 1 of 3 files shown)


nat/src/masquerade/apalloc/concurrent_fuzz.rs
```diff
@@ -198,11 +204,14 @@ impl Published {
                     carried.insert((ip, port.as_u16()));
                     reservations.push(reservation);
                 }
-                Err(AllocatorError::InternalIssue(message)) => {
-                    panic!("re-reserving {ip}:{port} for generation {generation}: {message}")
+                // Nothing may refuse a survivor. Every generation is built from the same specs, so
```
mvachhar, line 207: This seems wrong, the fixed code has the same issue as the original with `Err(_) => {}`

nat/src/masquerade/apalloc/concurrent_fuzz.rs
```diff
@@ -513,3 +515,130 @@ fn printing_the_pool_does_not_wedge_it_against_a_flow_ending() {
         releaser.join().expect("the releasing thread panicked");
     });
 }
+
+/// Reserving a port races the release of the block that holds it.
+///
+/// A block is given back in two steps: its free flag is stored `true`, and the weak entry in the
```
mvachhar, line 521: So all three commits here seem reasonable modulo the other comments, but the PR description has nothing to do with the commits that it contains, please fix that too.

mvachhar (changes requested):

mvachhar:
Superceded by #1707

PR #1700, opened 2026-08-06
(4)-masquerade: coverage, and the bugs it found

daniel-noland:
A coverage pass over the masquerade allocator, and what it turned up: three live bugs, and a
path no test had ever entered. Coverage
was measured with `just coverage`-style `cargo llvm-cov` runs with the bolero properties given
10s each, then read for cold paths.

**What it turned up**

- **Three defects in how claimed blocks and addresses are handled** — a fully claimed address
  taking a whole region out of service, a drifting count of usable port blocks, and an ordinary
  policy conflict reported as allocator corruption. All three are caused by the claims #1698
  introduces, so **the fixes now live in #1698** rather than here; review found they were merely
  sequenced late, not dependency-forced. The tests that pin them stay here, along with the
  coverage work that found them.
- **The flow carry-over path had never run.** A hole in the suite rather than a bug in the
  allocator, and the one that let the other two hide. Every reconfiguration test applied an
  *identical* configuration, which the writer answers by keeping the allocator it has. So
  `re_reserve_ip_and_port` was entered by no test at all, including for the source-VPC change
  earlier in this stack. Instrumenting it counted zero entries before, two after.

**Coverage added**

IPv6 end to end (the offset mapping IPv4 skips entirely), a region allocated dry for real
(~129k allocations, about a second, which pins the well-known exclusion and the ordering
between addresses), and the paths that drop a flow when its configuration goes away. Dead
conversions that nothing called were deleted rather than tested.

Mutation testing earned its keep twice here: two tests passed on first write and were still
wrong, and the notes on them say so.

`nat` crate coverage went from 92.0% to 92.4% region measured with a clean profile (the working
figure during development read 93.3%, inflated by stale profile data from earlier runs -- see the
`just coverage` fix below), with `flows.rs` 48% to 92% and `natip.rs` 49% to 95%. The later test
commits on this branch take masquerade production code to 88.6% region and 66.5% branch.

## Added while reviewing the stack

**Fixes**

- **The three claimed-block fixes moved to #1698** — telling a block claimed in full apart from
  broken bookkeeping, keeping used-up addresses out of the pool (sixteen dropped packets for a
  claimed `/25`, measured; now zero), and keeping the usable-block count honest. They sit with the
  change that creates those states so no PR in the stack ships them live; #1698 describes them.
- **Flow timeouts stretch under an emulator** (moved to #1698, where the test that trips it lives)
  and **the exhaustion tests narrow their port space** there too. Between them these take
  `check/miri/powerpc64` from hanging past twenty-five minutes to green.
- **`just coverage` produces a report again** — it was broken in three places and no workflow runs
  it, which is how it stayed broken.
- **`just miri::test` keeps the rustflags `.cargo/config.toml` sets.** Setting `RUSTFLAGS` in the
  environment replaces `build.rustflags` rather than adding to it, so miri was checking a different
  build of tokio from every other job, and every `cfg(emulated)` site warned.

**Tests**

Claims a carried-over flow meets; an address brought in by a reservation keeping them; a port block
coming back while its address stays; an address retired for good; the two config changes that end a
flow (its port claimed, its peering gone); the error classifications pinned against each other; an
IPv6 question for the claims; and a connection walked through an ordinary close in both directions.

Three existing tests could not fail and were mended: one property compared 24 with 24 on every
input, another discarded every error but the one it named, and the concurrent suite dropped a
successful reservation as a temporary, releasing the port it was meant to hold.

The concurrent suite's two oracle fixes have moved down to #1699, where the runtime fix they
assert against lives: they cannot hold before it, and there they double as a regression test for
that PR's own change.

Every fix here is mutation-verified. Mutation testing also caught three claims of mine that were
wrong before they landed, including one in a commit message.


nat/src/masquerade/natip.rs
```diff
@@ -22,12 +21,6 @@ pub trait NatIp:
     // Convert to `IpAddr` object
     fn to_ip_addr(&self) -> IpAddr;
 
-    // Extract the source IP address from a Net object as a `NatIp`
-    fn from_src_addr(net: &Net) -> Option<Self>;
```
mvachhar, line 26: I disagree with this, we should evaluate if it is an appropriate part of the trait.  If it is we should add tests.  if it is not, then deletion is fine.

nat/src/masquerade/apalloc/pool_fuzz.rs
```diff
@@ -594,6 +594,70 @@ fn an_address_with_room_is_reused_before_a_fresh_one_is_drawn() {
     );
```
mvachhar, line 594: The commit comment for "test(masquerade): Allocate a region dry for real" doesn't make sense.  It starts with:
```
The property tests reach exhaustion by claiming ports away, which is a
model of a full pool rather than a full pool.
```

nat/src/masquerade/apalloc/pool_fuzz.rs
```diff
@@ -594,6 +594,70 @@ fn an_address_with_room_is_reused_before_a_fresh_one_is_drawn() {
...
+/// Everything else here claims ports away to reach exhaustion cheaply, which is a model of it
+/// rather than the thing itself. This does it the long way: it takes about 64k allocations to move
+/// off one address, which is exactly why a handful of allocations never leaves the first, and why
+/// the paths that move between addresses and finally give up were untested.
+///
+/// A second per run buys the whole ladder: every port of an address, the move to the next address,
+/// the last port of the last address, and the refusal after that.
+#[test]
+fn a_region_can_be_allocated_dry() {
+    // 65536 ports, less the IANA well-known range that masquerade keeps off for TCP.
```
mvachhar, line 608: Since AI wrote this and we are in a time crunch I'll let this slide, but we should have a test method for the allocator to simply allocate all but one port and then just allocate that one port to reduce the computational load of this test.

nat/src/masquerade/test.rs
```diff
@@ -19,6 +19,7 @@ use flow_entry::flow_table::{FlowLookup, FlowTable};
 // #[dpdk::with_eal] below. The miri-eligible tests use the local TestFlowFilter mock instead.
 #[cfg(not(miri))]
```
mvachhar, line 20: Commit message for `test(masquerade): Cover every way a config change ends a flow` should be made sensible.

nat/src/masquerade/apalloc/pool_fuzz.rs
```diff
@@ -171,6 +171,17 @@ fn allocations_are_unique_and_within_the_declared_ranges() {
         });
```
mvachhar, line 171: Please shorten the commit message for `test(masquerade): Cover every way a config change ends a flow` to get to the meat of the issue.  There are more details than make sense within the context of the commit.

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -758,3 +758,62 @@ pub(crate) fn map_address(
         .and_then(|offset| prefix_offset.checked_add(offset))
         .ok_or(AllocatorError::NoPoolFound)
 }
+
+#[cfg(test)]
```
mvachhar, line 762: Please shorten the commit message for `test(masquerade): Cover a port block coming back while its address stays` to just describe the tests it adds instead of the life story of how we got here.

justfile
```diff
@@ -459,24 +459,27 @@ doctest *args:
     {{ _just_debuggable_ }}
     cargo test --doc {{ _cargo_feature_flags }} {{ _cargo_profile_flag }} {{ args }}
```
mvachhar, line 460: Again the commit message in `build(coverage): Make `just coverage` produce a report again` does not need the life story of how exactly the prior problem occurred.  A short description of the issue like, -Cinstrument-coverage was missing so add it back, or somethign like that would be much better than this AI slop.

nat/src/masquerade/allocation.rs
```diff
@@ -96,3 +96,121 @@ impl<T: Debug + Display> Display for AllocationResult<T> {
         )
     }
```
mvachhar, line 97: Again, no life story in `test(masquerade): Pin the error classifications against each other`. Not sure why the AI couldn't just say AllocatorError does not have coverage for case XXX, so this commit adds a test to cover this case.  The message doesn't even fully make sense in the first paragraph.

nat/src/masquerade/allocation.rs
```diff
@@ -96,3 +96,121 @@ impl<T: Debug + Display> Display for AllocationResult<T> {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
mvachhar, line 102: Are these tests really adding any value?

nat/src/masquerade/test.rs
mvachhar, line 23: Please reword the commit message of `test(masquerade): Walk a connection through an ordinary close` to not use short sentence AI speak and sound more human.

nat/src/masquerade/apalloc/reserved.rs
```diff
@@ -262,8 +262,12 @@ mod bolero_tests {
         fn reserved(&self) -> ReservedPorts {
             let mut reserved = ReservedPorts::default();
             for &(offset, length, shape, port_lo, port_span) in &self.claims {
-                let start = BASE + u32::from(offset) % WINDOW;
```
mvachhar, line 265: I like the test coverage added by `test(masquerade): Let claims overhang the window the sweep is asked about` but the way it is described in the commit message is really confusing, please fix.

nat/src/masquerade/flows.rs
```diff
@@ -61,9 +61,16 @@ fn re_reserve_ip_and_port(
...
-    // have one, so the flow key's source really is the VPC the masqueraded traffic originates in.
-    let src_vpcd = flow_key.src_vpcd().unwrap_or_else(|| unreachable!());
-    let dst_vpcd = flow_info.get_dst_vpcd().unwrap_or_else(|| unreachable!());
+    // have one, so both discriminants should be there. Should: this runs over every flow in the
+    // table while a configuration is applied, and a flow that cannot say which VPCs it belongs to
+    // is one this cannot carry over. Refusing it drops that flow; asserting would take the whole
+    // data plane down in the middle of a config change.
+    let (Some(src_vpcd), Some(dst_vpcd)) = (flow_key.src_vpcd(), flow_info.get_dst_vpcd()) else {
+        error!(
+            "Flow {flow_key} has no VPC discriminant, so it cannot be carried over. This is a bug"
```
mvachhar, line 70: Should we tag this as an internal error or are all `error!` logs really internal errors?

nat/src/masquerade/apalloc/pool_fuzz.rs
mvachhar, line 174: Commit message should not refer to a PR the way it does here, please fix: `test(masquerade): Pin that a pair in use is never handed out twice`

mvachhar (changes requested):

mvachhar:
Superceded by #1707

PR #1704, opened 2026-08-07
build(nix): Wrap vtysh so it loads the gateway's commands

### LINKED ISSUE #1674: image FRR: add alias for vtysh to load extensions
* We augment the vtysh with a couple of custom commands
* Those extra commands can be loaded by vtysh with -X (due to our patches) plus the name of the object we build.
* Currently, the artifact is placed in `/lib/libvtysh_hedgehog.so`

**Goal**
Augment the FRR image with an alias. E.g. s/t like
`alias vtysh='vtysh -X /lib/libvtysh_hedgehog.so'`

daniel-noland:
Addresses the goal of issue #1674, by a different route than the one the issue proposes. 

Our FRR fork teaches vtysh a `-X <path>` flag that dlopen()s an extension, and the commands the gateway adds live in one such extension, `libvtysh_hedgehog.so`. Nobody passes the flag today, so those commands are absent unless whoever is typing knows to ask.

## Why a wrapper rather than an alias

An alias only reaches callers that read a shell rc, and the two that matter most do not: `kubectl exec` and `docker exec` start no login shell. The shortcut would be missing in exactly the session someone opened to work out what a running gateway is doing.

### DIFF
--- a/nix/overlays/frr.nix
+++ b/nix/overlays/frr.nix
@@ -26,13 +26,16 @@ let
       '';
     });
   frr-build =
-    frrSrc:
+    {
+      frrSrc,
+      vtysh-extensions ? false,
+    }:
     dep (
       (final.callPackage ../pkgs/frr (
         final.fancy
         // {
           stdenv = final.stdenv';
-          inherit frrSrc;
+          inherit frrSrc vtysh-extensions;
         }
       )).overrideAttrs
         (orig: {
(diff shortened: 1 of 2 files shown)


mvachhar (approved):
Do we care about the AI nitpick about the install vs. post-install in the comment?  If we care fix and then merge, otherwise just merge.

PR #1709, opened 2026-08-10
CLI: add output filtering and highlighting

Fredi-raspall:
Adds a simple syntax to filter and/or highlight the output of the cli. Filtering/highlighting must come at the end of the line and start with a pipe | symbol. The syntax is:
  - CMD | only FOO: include only lines containing FOO
  - CMD | not BAR:  exclude lines containing BAR
  - CMD | high BAZ: highlight BAZ

All of the single filters can include multiple words or tokens (space separated). E.g.:

 CMD | high BAZ BAR FOO would highlight all of the tokens present
 CMD | not  BAZ BAR would not show the line if it contains any
 CMD | only BAZ BAR FOO would show lines containing any of the
 tokens.

  
 And an arbitrary number of filters can be specified in a command.
 E.g. CMD | not FOO BAR | Only BAZ Other | high BAZ
 would:
    - not show lines including either FOO or BAR
    - only show lines that included BAZ or Other
    - of the remaining lines, highlight BAZ word

### DIFF
--- a/cli/bin/filters.rs
+++ b/cli/bin/filters.rs
@@ -1,0 +1,209 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Open Network Fabric Authors
+
+//! Filter setup for filtering outputs
+
+use colored::Colorize;
+use regex::Regex;
+use std::str::SplitWhitespace;
+
+#[derive(Debug, PartialEq)]
+pub enum FilterCode {
+    Only,
+    Not,
+    HighLight,
+}
+impl TryFrom<&str> for FilterCode {
+    type Error = &'static str;
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        match value {
+            "only" => Ok(FilterCode::Only),
+            "not" => Ok(FilterCode::Not),
+            "high" => Ok(FilterCode::HighLight),
+            _ => Err("Unknown filter"),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Filter {
+    code: FilterCode,
+    tokens: Vec<String>,
+}
+impl Filter {
+    fn from_code(code: FilterCode) -> Self {
+        Self {
+            code,
+            tokens: Vec::new(),
+        }
+    }
+    fn add_token(&mut self, token: &str) {
+        self.tokens.push(token.to_owned());
+    }
+}
+
+pub const PIPE: &str = "|";
+pub const HEADING_CHAR: char = '━';
+
+fn highlight(line: &str, re: &Regex) -> String {
+    let mut result = String::new();
+    let mut last = 0;
+
+    for m in re.find_iter(line) {
+        result.push_str(&line[last..m.start()]);
+        result.push_str(&m.as_str().red().bold().to_string());
+        last = m.end();
+    }
+
+    result.push_str(&line[last..]);
+    result
+}
+
+pub fn parse_filters(
+    mut split: SplitWhitespace<'_>,
+    filters: &mut Vec<Filter>,
+) -> Result<(), &'static str> {
+    loop {
+        let Some(keyword) = split.next() else {
+            return Err("Missing filter");
+        };
+        // build filter from code
+        let mut filter = Filter::from_code(FilterCode::try_from(keyword)?);
+        let mut more_filters = false;
+        for word in split.by_ref() {
+            if word == PIPE {
+                // new filter begins. Store the current one and parse the next
+                more_filters = true;
+                break;
+            }
+            filter.add_token(word);
+        }
+        if filter.tokens.is_empty() {
+            return Err("Missing filter token");
+        }
+        filters.push(filter);
+        if !more_filters {
+            return Ok(());
+        }
+    }
+}
+
+// Filter some output based on the given filters
+pub fn filter_output(output: &str, filters: &Vec<Filter>) -> String {
+    if filters.is_empty() {
+        return output.to_string();
+    }
+
+    // Highlighting does not affect what is output and can be executed last,
+    // collecting all of the patterns to highlight into a single regex.
+    let patterns: Vec<String> = filters
+        .iter()
+        .filter(|filter| matches!(filter.code, FilterCode::HighLight))
+        .flat_map(|filter| filter.tokens.iter().map(String::as_str).map(regex::escape))
+        .collect();
+
+    let high_regex = if patterns.is_empty() {
+        None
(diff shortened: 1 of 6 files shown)


cli/bin/filters.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+impl TryFrom<&str> for FilterCode {
+    type Error = &'static str;
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        if value.starts_with("only") {
+            Ok(FilterCode::Only)
+        } else if value.starts_with("not") {
+            Ok(FilterCode::Not)
+        } else if value.starts_with("high") {
+            Ok(FilterCode::HighLight)
```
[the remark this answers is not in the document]
Fredi-raspall: fixed

cli/bin/filters.rs
```diff
@@ -0,0 +1,192 @@
...
+                    FilterCode::Not => {
+                        if include {
+                            include = !filter.tokens.iter().any(|w| line.contains(w));
+                        }
+                    }
+                    FilterCode::HighLight => {
+                        let all: Vec<&str> = filter.tokens.iter().map(String::as_str).collect();
+                        let re = all.join("|");
+                        let re = Regex::new(re.as_str()).unwrap();
+                        lineout = highlight(line, &re);
```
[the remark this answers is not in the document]
Fredi-raspall: fixed

cli/bin/filters.rs
```diff
@@ -0,0 +1,194 @@
...
+                        let all: Vec<&str> = filter.tokens.iter().map(String::as_str).collect();
+                        let re = all.join("|");
+                        if let Ok(re) = Regex::new(re.as_str()) {
+                            lineout = highlight(line, &re);
+                        } else {
+                            // if we fail to build the regex for highlighting, don't highlight
+                            // and include the line verbatim
+                            lineout = line.to_string();
+                        }
+                    }
```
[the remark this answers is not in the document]
Fredi-raspall: No, evaluation should be as specified by the user. If that's wrong, so be it.

qmonnet (commented):
```
Finally, note that order matters. Filters are evaluated in the
order they are specified. So:

CMD | only FOO | not FOO -> will display nothing
CMD | not FOO | only FOO -> will display lines containing FOO
```

I'd expect the second line to print nothing, too; but I suppose this is because I'm used to the bash pipeline syntax, with the text being filtered sequentially before being passed to the next filter command. So this doesn't work the same, right? It's more like parsing conflicting options, where the last takes precedence, such as in `gcc -Wno-error -Werror`?

Fredi-raspall:
> ```
> Finally, note that order matters. Filters are evaluated in the
> order they are specified. So:
> 
> CMD | only FOO | not FOO -> will display nothing
> CMD | not FOO | only FOO -> will display lines containing FOO
> ```
> 
> I'd expect the second line to print nothing, too; but I suppose this is because I'm used to the bash pipeline syntax, with the text being filtered sequentially before being passed to the next filter command. So this doesn't work the same, right? It's more like parsing conflicting options, where the last takes precedence, such as in `gcc -Wno-error -Werror`?

The comment is actually wrong. In both cases, nothing would be printed. Will amend the commit.

cli/bin/filters.rs
```diff
@@ -0,0 +1,194 @@
...
+        Ok(code) => code,
+        Err(e) => {
+            print_err!("{e}");
+            return;
+        }
+    };
+    // build filter from code
+    let mut filter = Filter::from_code(code);
+    while let Some(word) = split.next() {
+        if word.contains(PIPE) {
```
qmonnet: This means you cannot filter on a token containing `|` in it, right? There's no `\|` escape or quoting mechanism? (Probably doesn't matter much, I'm just checking)

cli/bin/filters.rs
```diff
@@ -0,0 +1,194 @@
...
+            // line is each of the lines we are about to process. For each line, we
+            // need to tell if it must be output or not and how. `Include` below is
+            // per line and must be true in the last filter for the line to be shown.
+            let mut include = true;
+            let mut lineout = line.to_string();
+            if line.contains(HEADING_CHAR) {
+                return Some(lineout);
+            }
+            // apply the filters in order for each of the lines
+            for filter in filters {
```
qmonnet, line 123: I suppose that running the filters in the order they're specified on the command line means you risk highlighting tokens that will later get filtered out; but it probably doesn't matter much, the alternative (re-ordering filters to have the highlights at the end) is probably more complexity for too little gain, and not worth the trouble.

cli/bin/filters.rs
```diff
@@ -0,0 +1,192 @@
...
+            return;
+        }
+    };
+    // build filter from code
+    let mut filter = Filter::from_code(code);
+    while let Some(word) = split.next() {
+        if word.contains(PIPE) {
+            // new filter begins. Store current and recurse
+            filters.push(filter);
+            parse_filters(split, filters);
```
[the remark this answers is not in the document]
qmonnet: Ah that's related to my own comment below on supporting `|` as or in a token

cli/bin/filters.rs
Fredi-raspall: Yeah, if you highlight and then filter, well, that's your choice. A user may be investigating something and then decide to filter stuff. Highlighting should not decide if output is shown. True, we allow highlighting and then filtering out the highlighted text. I can put it in the end, but I won't change the output.

qmonnet: Right, I didn't mean changing the output at all, just mentioning we could avoid some useless highlight transformations (by moving highlight filters to the end of the filters chain). But I agree it's probably not worth it, this was only a comment, not a call for change, sorry if that was confusing.

qmonnet (approved):
Approved, but please go through the AI reports - [the one about rejecting filters when the user provides no filter](https://github.com/githedgehog/dataplane/pull/1709#discussion_r3749080424), in particular, sounds like something that should be fixed.

cli/bin/filters.rs
Fredi-raspall: I've fixed the code, but yeah, I don't think anyone would filter with |

Fredi-raspall:
> Approved, but please go through the AI reports - [the one about rejecting filters when the user provides no filter](https://github.com/githedgehog/dataplane/pull/1709#discussion_r3749080424), in particular, sounds like something that should be fixed.

I think I fixed all of the issues reported by AI.

PR #1715, opened 2026-08-11
decouple port forwarding and masquerading

Fredi-raspall:


nat/src/masquerade/flows.rs
```diff
@@ -16,17 +15,18 @@ use net::flows::FlowInfo;
...
 
-/// Detach flows from an allocator before it is removed.
-pub(crate) fn remove_allocator_from_flows(flow_table: &FlowTable) {
+/// Invalidate all masquerading flows
+pub(crate) fn invalidate_masquerade_flows(flow_table: &FlowTable) {
+    debug!("INVALIDATING all masquerading flows...");
     flow_table.for_each_flow(|_key, flow_info| {
         if flow_info.locked.read().nat_state.as_ref().is_some() {
             flow_info.invalidate_pair();
         }
```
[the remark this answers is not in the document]
Fredi-raspall: nat_state is only set for masqueraded flows.

nat/src/masquerade/apalloc/reserved.rs
```diff
@@ -77,6 +97,14 @@ impl ReservedForAddr {
             .iter()
             .any(|range| range.start() <= port && port <= range.end())
     }
+
+    /// Tell the number of ports reserved. By construction, we can sum the ports
+    /// of each range since, by construction, they are disjoint.
```
qmonnet: Nit: repetition for “by construction”

qmonnet (approved):
Looks good to me, thank you.

There's one thing I'd like to better understand regarding the 2nd commit (`fix(masquerade): fix lookup of masquerading peering`): does it fix an issue that is present in `main`, or does it fix the previous commit in the PR? If it's present in `main`, what was the impact?

Fredi-raspall:
> Looks good to me, thank you.
> 
> There's one thing I'd like to better understand regarding the 2nd commit (`fix(masquerade): fix lookup of masquerading peering`): does it fix an issue that is present in `main`, or does it fix the previous commit in the PR? If it's present in `main`, what was the impact?

It fixes main. Hard to tell what the impact would have been. Probably minor, but it is a correctness fix. If the masquerade config includes port-forwarding peerings, when checking masquerading flows, we could have wrongly hit a port-forwarding peering instead of a masquerading one and tried to validate the flow against that one.

nat/src/masquerade/apalloc/reserved.rs
Fredi-raspall: Fixed!

PR #1716, opened 2026-08-13
Pr/fredi/masq allocation logs

Fredi-raspall:


daniel-noland (approved):
I don't love increasing the global reasoning (we REALLY need to stop doing that).  That said, we don't have a lot of latitude to say no in this case, so I'll approve.

PR #1719, opened 2026-08-14
bump: FRR back to stable/10.5 (just in case)

daniel-noland:
FRR 10.6 regressed on us, so walk the three pins that moved in ddae59e2f ("bump: FRR on stable/10.6") back to their 10.5 branches:

  frr            stable/10.6    -> stable/10.5
  frr-dp         hh-master-10.6 -> hh-master-10.5
  dplane-plugin  hh-master-10.6 -> hh-master-10.5

`frr-dp` and `dplane-plugin` land exactly where they sat before the 10.6 bump -- neither 10.5 branch has moved since. Upstream `frr` picks up the current tip of stable/10.5 (5ece2d67) rather than the older e3681378 we were on, which is the intended behavior for these pins: they float with their branch on pin bump.

`dplane-plugin` goes to `hh-master-10.5` rather than back to `master` as it was named pre-bump; `master` has since advanced onto the 10.6 work, and `hh-master-10.5` is the branch holding the revision we actually ran.

`nix/pkgs/frr` needs no change: `version` is derived from the pin's branch, and both patches apply cleanly to the 10.5 trees.

### DIFF
--- a/npins/sources.json
+++ b/npins/sources.json
@@ -52,11 +52,11 @@
         "owner": "githedgehog",
         "repo": "dplane-plugin"
       },
-      "branch": "hh-master-10.6",
+      "branch": "hh-master-10.5",
       "submodules": false,
-      "revision": "19d17cdba14bfca79fa6797e8459f217232f906c",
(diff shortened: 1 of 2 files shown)


daniel-noland:
closed in response to #1718

PR #1722, opened 2026-08-16
perf(ci): make the nix dependency build reusable across revisions

daniel-noland:
Stacked on #1714.  Read that one first; this branch is based on it, and the
diff shown against `main` includes its commits until it merges.

#1714 decides *what* CI runs and when.  This decides how much it costs.

## What was wrong

Every crane derivation opted out of crane's deps-only split, so each one
compiled the whole dependency graph itself -- 476 of 516 crates in a full
build are third-party, and `-Zbuild-std` rebuilds the standard library on top
of that.  Sharing one dependency build per flag-set fixes that, but only if
the shared artifact actually survives a source edit, and three separate
channels were invalidating it:

- `--remap-path-prefix==${src}` put the filtered source store path into
  RUSTFLAGS, which is part of both the derivation and cargo's per-unit
  fingerprint.  Any Rust edit gave a new compilation identity to all 516
  crates, including the ~475 that contain none of our source.
- `VERSION` came from `git describe`, so it changed per commit.
- `results/` was inside `src`.

`check` was untouched by any of this, because `check-clippy` and
`check-doctest` invoked bare `cargo` against an empty `target/` on an
ephemeral runner.  Routing them through nix exposed that the `docs`
derivation had rotted: thirty-one broken intra-doc links across six crates,
and a missing `--check-cfg=cfg(emulated)` in RUSTDOCFLAGS that would have
failed the build even with every link correct.

## Results

Like-for-like, 32 jobs, measured on the lab runners:

| scenario | before | after |
| --- | --- | --- |
| pull request that edits Rust | 213.5 runner-min | **169.4** (-21%) |
| re-run, docs-only change, or merge-queue recheck | 213.5 runner-min | **103.0** (-52%) |

`clippy` went from forty-one per-package derivations (~12 min) to one (32.4s),
and now runs `--all-targets`: test code is as load bearing as the rest and
gets the same static analysis.  `check/debug` holds at ~233s while also
linting test targets and building docs, neither of which it did before.

Much of what remains was traded from compute to network, which is the cheaper
resource on these runners.

## Guard

`just check-deps-reuse` builds the dependency derivation before and after a
synthetic source edit and fails if the hash moves.  It catches both
regressions above, and runs in `lint` and `check`.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -13,7 +13,7 @@ CARGO_LLVM_COV_BUILD_DIR = { value = "target/llvm-cov/target", relative = true, 
 
 [build]
 # Register `emulated` so cfg_attr sites don't trip unexpected_cfgs natively.
-rustflags = ["--cfg=tokio_unstable", "--check-cfg=cfg(emulated)"]
+rustflags = ["--cfg=tokio_unstable", "--check-cfg=cfg(emulated)", "--check-cfg=cfg(instrumented)"]
 
 [target.wasm32-wasip1]
 # Trailing `--` separates wasmtime's CLI from the module + module args
--- a/.github/actions/nix-shell/action.yml
+++ b/.github/actions/nix-shell/action.yml
@@ -32,6 +32,8 @@ runs:
         signingKey: '${{ inputs.cachix_signing_key }}'
         # prettier-ignore
         authToken: '${{ inputs.cachix_auth_token }}'
+        # Exclude source-volatile builds and image paths that can reintroduce them through a closure.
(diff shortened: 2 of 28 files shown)


daniel-noland:
Superseded by #1725, which is the same branch re-parented onto #1724.

The debug images were split out of #1714 into #1724 on review, and this PR's nix work depends on them (`perf(nix): remap sources to a fixed prefix` patches their `extraCommands` and adds `source-tree`), so it has to sit above them in the stack. GitHub declines to change a stacked PR's base through either the GraphQL or REST API, so this one is closed rather than re-parented.

PR #1733, opened 2026-08-17
fix(net): store the checksum an incremental update computes

daniel-noland:
First of five, stacked on #1724. Extracts the finished fuzzing campaign from
`pr/daniel-noland/icmp-flow-lock`, which was never opened as a PR and had
drifted well behind `main`.

This one is a production bug, on its own so it can be taken without the rest.

`Checksum::increment_update_checksum` works out the new checksum and hands it
back; it does not store it, despite taking `&mut self`.
`EmbeddedTransport::update_checksum` called it and dropped the answer, under a
comment about ignoring errors on a truncated header — but there is no error to
ignore, the return value *is* the checksum. So no quoted transport header has
ever had its checksum updated.

NAT rewrites the ports of the packet quoted inside an ICMP error and calls this
to keep the checksum in step. Receivers that validate the quoted checksum will
reject the error; that is the path PMTUD depends on. Plenty of stacks do not
validate it, which is likely why nothing has complained.

`update_checksum_for_address` is new, and folds a change of one of the quoted
packet's addresses in — TCP, UDP and ICMPv6 checksum over a pseudo-header built
from the addresses, so rewriting one leaves the quoted checksum describing an
address that is no longer there. ICMPv4 has no pseudo-header and is left alone.

Verified locally rather than in CI, given the outage: `dataplane-net` 473/473,
`fmt --check` and `clippy -D warnings` clean. Reverting the fix fails exactly
`a_v4_address_change_matches_a_fresh_tcp_checksum` and its v6/UDP twin.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -103,8 +103,8 @@ If those queue failures stop being rare, the phasing is worth revisiting.
 - Checks: `debug` by default; `release` and `fuzz` on deep runs
 - Coverage: `debug` by default; `fuzz` on deep runs
 - Miri: required on deep runs; opt-in on pull requests with `ci:+miri`
-- Containers: debug/release for dataplane, its debugger, and FRR; release for
-  validator
+- Containers: debug/release for dataplane, its two debug images, and FRR;
+  release for validator
 - VLAB configurations: spine-leaf fabric mode, L2VNI/L3VNI VPC modes,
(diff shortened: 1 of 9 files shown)


daniel-noland:
Its single commit now opens #1734; a one-commit PR with a ten-commit child was a review artifact, not a chapter.

PR #1734, opened 2026-08-17
test(net): property-test the headers view, matcher, and builder

daniel-noland:
Second of five, stacked on #1733.

Property tests for `net::headers` — the `HeadersView` unsafe boundary, the
pattern matchers, the builder's `Within::conform`, and the embedded (ICMP-quoted)
header views. Coverage of the module went from roughly a third to the high
nineties; the lines still uncovered are the `unreachable_unchecked` arm of each
arity, which must stay uncovered because reaching one is the undefined behaviour
the whole `HeadersView` invariant exists to prevent.

The recurring finding, across five separate instances, was code that no
generator could reach rather than code that was wrong: `embedded_ip` was never
set by any generator, the absent-layer arm of every `opt_*` method was
unconstructible, no shape ever named `Vxlan`, the first-step-refusal branch was
unreachable at every arity, and `conform` was only reachable through a builder
no test used. In each case "we have tests for that" was false and the tests
looked fine.

Two things the tests pin rather than fix, both for the owner to decide:

- `when`/`inspect`/`otherwise` on the embedded matchers read only the inner
  accumulator, while `done()` requires both. A packet whose outer chain fails
  but whose quoted packet matches runs `inspect`, skips `otherwise`, and then
  returns `None` — 4,974 packets in 25,000. Documented behaviour, but
  `otherwise` is the error hook and stays silent for a whole class of failure.
- All twelve `Within<Icmp4|Icmp6> for <subtype>` `conform` bodies are dead.
  `Install` runs afterwards from `build_headers` and overwrites whatever they
  wrote, and nothing can be stacked on a subtype, so there is no arrangement in
  which `conform` gets the last word.

The last commit is new, not extracted: the embedded vacuity guard only asserts
once a run is long enough for a miss to mean something, and its threshold was
set at the case count a default run managed when it was written. Coverage
instrumentation costs about a fifth of the throughput here — measured at 17 hits
in 20,206 cases against 21 in 25,857 — which left the threshold sitting a couple
of hundred cases below what CI actually draws, so a busier runner would drop
under it and the check would disappear silently. Lowered to ten thousand.

Worth recording since #1714 adds `--cfg=instrumented` for this purpose: these
properties are bound by the generator rather than by a counter loop, so unlike
the fib test that motivated the cfg, **no iteration counts need cutting here**.

Verified locally: `dataplane-net` 473/473, `fmt --check` and
`clippy -D warnings` clean.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -103,8 +103,8 @@ If those queue failures stop being rare, the phasing is worth revisiting.
 - Checks: `debug` by default; `release` and `fuzz` on deep runs
 - Coverage: `debug` by default; `fuzz` on deep runs
 - Miri: required on deep runs; opt-in on pull requests with `ci:+miri`
-- Containers: debug/release for dataplane, its debugger, and FRR; release for
-  validator
+- Containers: debug/release for dataplane, its two debug images, and FRR;
+  release for validator
 - VLAB configurations: spine-leaf fabric mode, L2VNI/L3VNI VPC modes,
   with gateway enabled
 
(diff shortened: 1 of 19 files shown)


daniel-noland:

PR #1735, opened 2026-08-17
test(routing): property-test the rib, fib, and control plane

daniel-noland:
Third of five, stacked on #1734. The largest slice, and the one carrying the
most production fixes.

Six fixes, each with the property test that found it:

- **Survive a resolution loop in the next-hop walks.** A cycle in next-hop
  resolution ran away instead of terminating.
- **Install a route with no next-hops as a drop, in the rib.** It was previously
  installed as something other than a drop.
- **Refuse to remove the default vrf.**
- **Drop a fib's vni alias when the fib goes**, rather than leaving the alias
  pointing at nothing.
- **Keep the interface name out of the next-hop key** — two next-hops differing
  only by interface name compared unequal when they should not.
- **frrmi: frame a message by what arrived, not by what was asked for**, which
  is a framing bug on the FRR management interface.

Alongside them, property tests and model-checks for the rib-to-fib conversion,
the fib against its own change log, vrf deletion, resolution across vrfs, the
control-plane operations, the FRR config renderers, the interface table and its
attachments, and the adjacency table's publish discipline.

`routing/` had **zero** commits on `main` since this work was written, so it
cherry-picked without a single conflict and needed no adaptation.

Verified locally rather than in CI, given the outage: `dataplane-routing`
115/115, `fmt --check` and `clippy -D warnings` clean.

### DIFF
--- a/routing/src/atable/atablerw.rs
+++ b/routing/src/atable/atablerw.rs
@@ -82,3 +82,298 @@ impl AtableReaderFactory {
         AtableReader(self.0.handle())
     }
 }
+
+/// Model-based properties over the adjacency table and its left-right wrapper.
+///
+/// The table is a map. What is not map-like is the *publishing*: `add_adjacency`, `del_adjacency`
+/// and `clear` all take a `publish` flag, and `AtResolver::refresh_atable_from_proc` relies on it --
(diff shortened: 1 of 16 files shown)


daniel-noland:

PR #1736, opened 2026-08-17
fix(flow-entry): add an insertion that will not displace a live flow

daniel-noland:
Fourth of five, stacked on #1735. One commit.

Adds an insertion that will not displace a live flow, and the test for it.
Displacing a flow takes the other half of its pair with it; the two halves have
different keys, so nothing displaces the partner in its own right, and it is
left live mapping a translation whose allocation has returned to the pool.

Small, but it is the commit that most needed adapting to a `main` that has moved
on, and it is worth knowing how it failed. It cherry-picked with no conflict,
then failed to compile because `FlowInfo::related_pair` became fallible, then
compiled and *failed* with `InvalidPair("One of the flows must be the
initiator")` — an invariant `main` gained after this was written, which the test
was violating by flagging neither half. The forward half is now marked
`INITIATOR`, matching the sibling test in `concurrent_fuzz.rs`.

Three distinct failure modes behind one clean cherry-pick. With CI unavailable
this only surfaced because everything was run locally, which is the argument for
not trusting a clean apply on any of these branches.

Verified locally: `dataplane-flow-entry` 14/14, `fmt --check` clean.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -103,8 +103,8 @@ If those queue failures stop being rare, the phasing is worth revisiting.
 - Checks: `debug` by default; `release` and `fuzz` on deep runs
 - Coverage: `debug` by default; `fuzz` on deep runs
 - Miri: required on deep runs; opt-in on pull requests with `ci:+miri`
-- Containers: debug/release for dataplane, its debugger, and FRR; release for
-  validator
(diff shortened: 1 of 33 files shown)


daniel-noland:
Its single commit now opens #1737, for the same reason as #1733.

PR #1737, opened 2026-08-17
test(config,mgmt,k8s-intf): generate valid configurations and property-test the validator

daniel-noland:
Last of five, stacked on #1736. Twenty commits spanning `config`, `k8s-intf` and
`mgmt`.

These three cannot be split by package — most commits touch two or three of them,
so any split leaves broken intermediate states. They are one body of work:
generating configurations that are **valid by construction** rather than
generated and filtered, then property-testing the validator and the dataplane
tables a validated configuration implies.

The generators build configurations from an algebra of valid operations, which is
what makes them useful: a `TypeGenerator` over the config types produces
syntactically valid, semantically impossible configurations that the validator
refuses, so a coverage-guided fuzzer spends its budget on rejection paths. See
`development/code/config-algebra-testing.md` for the reasoning.

Three fixes, each found by the properties:

- **Refuse a port-forwarding expose the dataplane cannot build**, so the
  validator stops blessing configurations that fail at enactment. This is the
  constraint the whole campaign is aimed at — everything the validator accepts
  has to be enactable, because there is no channel back to the operator.
- **Give every vpc its own slots**, and assert the control validates.
- **Put the generated gateway in its own gateway groups.**

Plus a typed error for port-forwarding mismatches instead of a string.

Two notes for review:

- The three `feat(config): Generate … exposes` commits also touched `nat/`. The
  nat half is superseded — `main` has since gained `reserved.rs` and the
  static-claims rework, so `ReserveSets` moved out of `setup.rs` and a test
  module targeting the old shape was dropped. One nat change was **kept**:
  `overlay_offering` moves out of `nat/src/portfw/portfwtable/setup.rs` into the
  config generator module, which is that commit's intent rather than collateral.
- The final commit follows `genid` out of `MasqueradeConfig::new` into
  `update_nat_allocator`, where `main` moved it. Three commits introduced the
  affected call sites, so it is one commit rather than three amendments — which
  means those three, and the five between them, cannot compile `dataplane-mgmt`'s
  tests on their own. `--autosquash` fixes that if bisectable history inside the
  branch is worth more than the smaller diff.

Verified locally rather than in CI, given the outage: `config` + `mgmt` +
`k8s-intf` 209/209, `fmt --check` and `clippy -D warnings` clean.

With this merged, `pr/daniel-noland/icmp-flow-lock` can be retired — everything
on it is either here, already landed in reworked form, or deliberately dropped.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -103,8 +103,8 @@ If those queue failures stop being rare, the phasing is worth revisiting.
 - Checks: `debug` by default; `release` and `fuzz` on deep runs
 - Coverage: `debug` by default; `fuzz` on deep runs
 - Miri: required on deep runs; opt-in on pull requests with `ci:+miri`
-- Containers: debug/release for dataplane, its debugger, and FRR; release for
-  validator
+- Containers: debug/release for dataplane, its two debug images, and FRR;
+  release for validator
 - VLAB configurations: spine-leaf fabric mode, L2VNI/L3VNI VPC modes,
   with gateway enabled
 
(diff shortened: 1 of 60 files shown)


daniel-noland:

PR #1738, opened 2026-08-17
test: Drive static NAT, masquerade and ACL with configuration-relative packets

daniel-noland:
Stacked on #1737. **Base: `pr/daniel-noland/fuzz-config-generators`** — review the [Commits tab](https://github.com/githedgehog/dataplane/pull/1738/commits) or [compare against the base](https://github.com/githedgehog/dataplane/compare/pr/daniel-noland/fuzz-config-generators...pr/daniel-noland/fuzz-nf-probes), since Files Changed will show everything below it until the parents merge.

The fuzzing so far has been small, focused and intrusive: reach into a structure, exercise it directly, assert something about that structure. This is the other kind — **configure a network function, put generated packets through it, and assert properties that would hold of any NAT or any ACL** rather than of this one.

Three network functions, in increasing order of difficulty: static NAT (a pure function of its configuration), masquerade (stateful), and ACL (where a strong oracle already existed but stopped short of the packet).

## Packets have to be drawn relative to the configuration

`Packet` has a `TypeGenerator`, and pointing it at a generated NAT configuration is useless: every packet misses every table, the fuzzer explores the miss path, and the run is vacuous while looking enormous.

This is the same failure the design note rejects for configuration values, one level down, so it takes the same answer. **The configuration is a parameter to resolution, not a predicate to filter against.** A `ProbeSpec` is drawn with no reference to any configuration — it is a handful of indices — and `resolve` interprets it against the built `Fabric`. Resolution is total, so no draw is discarded and no rejection loop skews the distribution.

## The arrival state is the stage's precondition — and it is not always one stamp

Each stage assumes its predecessors annotated the packet, and nothing in the type system says so. `Arrival` writes that down once per stage, which is what the design note asks for when it puts contracts on network functions rather than on the pipeline.

Masquerade sharpened the point. `FlowLookup` attaches a flow entry only to a packet whose `dst_vpcd` is **absent**; the flow filter that sets `dst_vpcd` runs *after* it; `Masquerade` then requires it to be present. So the annotation has to arrive *between* two stages. The first version of the harness stamped everything up front, the way the static NAT one does — nothing crashed, but no packet was ever given flow state, every packet took the allocation path, and flows appeared to be re-allocated on every packet. Two properties failed and were right to: the harness was describing a pipeline that does not exist.

## Seventeen properties, none with an oracle

Nothing here predicts what an address translates to. Each property is a metamorphic relation or an invariant, so nothing is a second copy of `RangeBuilder` or of the allocator.

**Static NAT** (11 — one suite for address-only exposes, one for the port-range path): round trip, injectivity, frame, permission, attribution, marking.

**Masquerade** (6): reversibility, stability (a flow keeps its first translation), exclusivity, containment (the translation lands inside a declared range), permission, attribution.

**ACL** (4): the stage's verdict is the configuration's verdict, the five-tuple survives the round trip through a packet, a packet with no discriminants is refused, underlay traffic is left alone.

The three stages fail differently, and the properties say so. Static NAT's round trip crosses two independently built tables. Masquerade's crosses no table at all — the reverse translation exists only because the forward packet recorded it, so it is a claim about state. ACL's has a real oracle already and the new claim is about the packet path in front of it.

## Everything is break tested

A property that cannot fail is not evidence.

| break | fires |
| --- | --- |
| static NAT mapping shifted by one | round trip |
| every address mapped to its range start | round trip, injectivity |
| `set_checksum_refresh` dropped | marking |
| `src_natted` mark dropped | marking |
| silent pass on a missing source vni | attribution |
| static NAT permission gates ignored | permission |
| a port rewritten the mapping never named | frame |
| `PortAddrTranslationValue::get_entry` broken | **only** the two port properties |
| `AddrTranslationValue::get_entry` broken | **only** the four address properties |
| masquerade flow hot path disabled | stability, reversibility |
| `requires_masquerade` ignored | masquerade permission |
| allocation reported outside the declared range | containment, exclusivity |
| allocation failure passed silently | masquerade attribution |
| ACL `done(AclDropped)` dropped | stage verdict |
| ACL source address misread | stage verdict, summary round trip |
| ACL `is_overlay` gate removed | underlay gate |

Two of those are worth reading twice.

**The two static NAT suites are mutually exclusive.** Breaking the PAT path fails only the port properties; breaking the address path fails only the address properties. That is the evidence the extra suite reaches code the first does not, rather than re-running the same paths under new names.

**The ACL re-point closes a measurable gap.** Misread one field — take the source address from where the destination lives — and *all five pre-existing `fuzz::` properties pass*, because they never build the packet that is misread. Both new properties fail. That is the whole argument for the change, demonstrated rather than asserted.

## Generator gaps found along the way

Repeating a single-expose generator to get several does not work, for two reasons that are both facts about what a manifest accepts and so belong in the generator: exposes laid out from the same bases overlap, and independent draws mix address families, which a peering refuses.

Measured on the static NAT properties:

| | configurations that validated |
| --- | --- |
| repeating `StaticNatExpose` | 33% |
| one block per expose | 59% |
| one block and one shared family | 100% |

Two thirds of the budget was going into configurations that were thrown away — and worse than the waste, multi-expose configurations were nearly unreachable, so the case with more than one rule in a table was the one being skipped. `MasqueradeExposes` got the same treatment for the same reason. Both single-expose generators are unchanged, so existing callers in `nat` and `mgmt` are untouched.

## Vacuity

Every property counts what it reached and asserts a floor, because the failure that matters is an assertion that stops running rather than one that is wrong. All seventeen build 100% of the configurations they draw. The ACL guard also counts *denials*, since a run that only ever saw permits would pass while the drop path — the only path where the stage does anything — went entirely unexercised.

`Packet::enforce` removes a dropped packet from the output iterator, so probes carry `keep`: without it a drop and a pass-through are the same event from outside, and attribution could not be stated at all.

## The design note comes along

`docs(testing): Record the config-algebra testing strategy` was stranded on an abandoned branch and never reached this stack, and these commits cite it throughout, so it is carried here. A follow-up commit corrects its status line — the per-packet half of its decomposition now has three worked examples, while the operation algebra itself is still unbuilt and the enactment path refactor it implies is still deferred.

## Verification

CI cannot be trusted during the outage, so this was verified locally: `acl-filter` 28, `nat` 186, `config` 183, `mgmt` 19, all passing, with `cargo fmt --check` and `clippy -D warnings` clean on every package touched. Worth re-running once GitHub recovers, since local runs do not exercise the phasing, the cross-qemu path, or miri.

## Not in scope

- **Flow expiry.** Every masquerade property completes inside one flow lifetime, so none depends on expiry either happening or not. Expiry needs the explicitly driven clock the design note asks for, not a wall clock a property happens to outrun.
- **ICMP and arbitrary next headers through the ACL stage.** No builder produces a packet matching those summaries, so they stay with the summary-level properties. Same for the `CrossVersion` stray, which asks for a v4 source with a v6 destination.
- **Port forwarding, FIB, and the remaining stages.**
- The operation algebra itself — sequences, undo, commutation from read and write sets. Still a design note.


daniel-noland:

PR #1741, opened 2026-08-20
(scratch) test: track RFC 4787, and state its requirements as executable contracts

daniel-noland:
**Scratch PR.** Parking the specification-compliance line of work so it is pushed and reviewable;
not proposed for merge in this shape. Stacked on #1738.

## What is here

Nine commits, continuing the duvet work already in the stack (RFC 4884 and RFC 5382 landed earlier).

**Tracking RFC 4787** — the UDP counterpart of RFC 5382, 14 numbered requirements landing on the
same masquerade code that already carries RFC 5382 citations. Cited as the individual RFC, never as
BCP 127: that composite concatenates RFC 4787, RFC 6888 and RFC 7857, whose section numbers collide,
and duvet silently keeps only the last — 42 requirements where the three separately yield 129.

**Findings recorded, not fixed.** REQ-1 (mapping is address-and-port-dependent), REQ-8 (filtering is
stricter than either branch the RFC offers) and REQ-9 (no hairpinning) are one decision, not three:
this gateway is deliberately not UNSAF-traversal-friendly, or it is not deliberate and is a much
larger piece of work. REQ-13 is the same shape — this dataplane originates no ICMP error at all, and
has no MTU anywhere on the datapath to originate one from. All `todo` rather than `exception`,
because nobody has ruled.

**Executable contracts.** `nat/src/masquerade/contract.rs` states RFC 4787 REQ-12 as a
`Requirement` the implementation calls from a debug-only assertion and the test cited `type=test`
calls directly, so the two citations are provably about the same predicate. `SPEC` and `ID` are
`const` and checked against duvet's extracted requirements at compile time: a citation naming a
requirement its specification does not state now fails the build with `E0080`.

**Errata.** The corpus's `inline-errata/` was audited. RFC 4787, 5382, 5508, 6888 and 7857 have no
errata of any status; RFC 4884 has one, and it does not touch us.

## Open, and deliberately not done here

- The `.duvet/` snapshot is **not regenerated** — duvet is not on PATH outside the dev shell, and
  hand-editing the regression gate would defeat it. RFC 4787 REQ-12 should move to
  `[!MUST,implementation,test]` on the next `duvet report`.
- There is no `just duvet` recipe yet.
- `development/code/spec-compliance.md` carries a live open-questions list, expected to grow.
- The RFC corpus itself is moving to a separate repository; nothing here depends on it at build
  time, only `.duvet/` which is committed.

## Checks

`cargo test -p dataplane-nat` 210 pass, 1 ignored. clippy and fmt clean on `dataplane-nat` and
`dataplane`.


daniel-noland:

PR #1743, opened 2026-08-21
fix(masquerade): Close the window where an address belongs to neither list

daniel-noland:
## Issue

`NatPool` answers "is this public address available?" from two structures — a bitmap of free addresses and a list of weak references to the addresses currently in use — and an address is meant to be in exactly one of them. Releasing one passes through a state where it is in neither: `Weak::upgrade` stops resolving the instant the last `AllocatedPort` drops the strong count to zero, but the address only returns to the bitmap later, inside `AllocatedIp::drop`, which must first acquire the pool's write lock. An allocation arriving in that gap finds nothing on either path and is told the pool is empty, so it fails with `NoFreeIp` while the pool is in fact holding an address that belongs to no one. This is not a rare interleaving — `Arc` guarantees the ordering, so every release passes through the gap; only its duration varies, bounded by how long the releasing thread waits for a write lock that every allocation also takes, on every call, via `cleanup_used_ips`. The same reasoning covers a second latent hazard the fix closes: `reserve_from_pool` could create a duplicate `AllocatedIp` for an address whose owner had not yet finished releasing it, after which the first object's `Drop` would free an address the second still held.

## Impact

The packet that lands in the gap is dropped with `DoneReason::NatOutOfResources` and its flow is refused, logged as `masquerade: Ip/port allocation failed ... no free IP available`. Two properties of real deployments turn a narrow race into a recurring one. Where an expose is configured with a single public address per peering — the shape observed in the field — the unavailable address is the entire pool, so there is nothing to fall back on. And DNS drives the window open constantly: a reply from source port 53 moves the flow to `Closed` and invalidates the pair immediately as a port-conservation measure, so a pool carrying mostly DNS destroys and rebuilds its address roughly once per query, opening one window per query. The user-visible effect is mild and easily misread — resolvers retry, so lookups still succeed and the failures surface only as intermittent warnings and a raised drop count, at a rate that does not correlate with load in any obvious way.

## Fix

The root problem is that liveness was inferred from `Arc` refcounts, which change on whatever thread happens to drop the last reference, while availability is a property of the pool, which changes only under its lock — so the two could disagree. The fix gives every lease of an address a `Tenancy`, recorded with the address's bitmap offset alongside the weak reference, and keeps a map of current tenancies that is updated in the same critical section as the bitmap itself. That makes the pool's own state the source of truth and lets the hand-back be completed from either side: an allocation that holds the lock and finds a dead weak reference finishes the hand-back itself rather than waiting for the releasing thread to win the lock, and `use_new_ip` reclaims ended tenancies before it reports an empty bitmap — in the same critical section as the retry, so the answer cannot go stale between the two. A late `AllocatedIp::drop` whose tenancy has since been retired does nothing, which is what makes that takeover safe and, incidentally, what closes the duplicate-`AllocatedIp` hazard in `reserve_from_pool`.

Two smaller corrections ride along, both about being able to see what happened: the reuse scan no longer abandons the search at the first address that has run out of port blocks, stranding every address behind it in the list, and `NoFreeIp` no longer buries the more specific reason — which, on a single-address pool, it always did.

## Testing

- New property test in the existing shuttle harness, `an_allocation_racing_the_last_release_is_still_served`: a one-address pool with one thread releasing the only allocation while another allocates. Against the unfixed allocator it reproduces the field error verbatim (`NoFreeIp`); against this change it passes.
- `cargo test -p dataplane-nat --lib` — 165 pass.
- `cargo test -p dataplane-nat --features shuttle --lib shuttle` — 8 pass.
- `cargo clippy -p dataplane-nat --all-targets` — clean.
- Confirmed on a live gateway by deliberately widening the window: the unpatched allocator refused 10 of 20 queries with the drop counter and the log agreeing exactly, while this change served all 20 under the identical forcing function.

## Notes

- Not addressed here: ports are handed out in 256-blocks owned by a single thread and freed only when the last port in the block dies. Measured density on the affected fabric was ~3.35 ports/block for UDP, so a pool walls well below its nominal capacity. That is a separate redesign.
- A backport of this change onto `v0.25.2` is prepared on `backport/v0.25.2/masquerade-address-handback`; it is semantically identical (the two differ only by one clarifying comment).

### DIFF
--- a/concurrency/src/sync/loom_backend.rs
+++ b/concurrency/src/sync/loom_backend.rs
@@ -28,6 +28,9 @@
 //!     returns `Some` -- the upgrade-fails-after-last-strong-drop race
 //!     real `Weak` exposes is not modelled.
 //!   * `Arc::strong_count` reflects live `Arc`s **and** live `Weak`s.
+//!   * `Weak::strong_count` likewise never reaches `0` while the `Weak`
+//!     itself is alive, so a caller using it to detect "the last strong
+//!     reference is gone" never observes that condition under loom.
 //!   * `Arc::weak_count` panics: the shim has no separate weak count
 //!     to report, and returning `0` silently would make assertions
 //!     pass for the wrong reason on every backend. See the per-method
@@ -545,6 +548,24 @@ impl<T: ?Sized> Weak<T> {
     #[must_use]
     pub fn upgrade(&self) -> Option<Arc<T>> {
         self.inner.as_ref().map(|a| Arc(a.clone()))
+    }
+
+    /// See `std::sync::Weak::strong_count`.
+    ///
+    /// `0` for a `Weak` that has never been associated with an `Arc`, matching std. Otherwise the
+    /// same caveat as `Arc::strong_count` applies and then some: this shim's `Weak` holds a strong
+    /// clone, so the count includes live `Weak`s and cannot reach `0` while `self` is alive. A
+    /// caller asking "has the last strong reference gone?" is therefore always told "no" under
+    /// loom -- inert rather than wrong, but it means such a path is not model-checked here. That is
+    /// the same limitation the module header records for `upgrade`; see the note there about tests
+    /// that need the strong/weak race.
+    ///
+    /// Offered anyway because the alternative is worse: callers that must answer that question
+    /// while holding a lock the value's `Drop` also takes cannot use `upgrade`, since the
+    /// temporary `Arc` it mints may be the last one and would run `Drop` re-entrantly.
+    #[must_use]
+    pub fn strong_count(&self) -> usize {
+        self.inner.as_ref().map_or(0, inner::Arc::strong_count)
     }
 }
 
--- a/nat/src/masquerade/apalloc/alloc.rs
+++ b/nat/src/masquerade/apalloc/alloc.rs
@@ -19,6 +19,33 @@ use std::time::Duration;
 use tracing::{debug, error};
 
 ///////////////////////////////////////////////////////////////////////////////
+// Tenancy
+///////////////////////////////////////////////////////////////////////////////
+
+/// Identifies one lease of one public address, so that a lease which has ended cannot be confused
+/// with the lease that replaced it.
+///
+/// A [`NatPool`] hands an address out at most once per tenancy. The address returns to the pool
+/// when its tenancy ends, and ending a tenancy that is no longer the current one does nothing. That
+/// is what makes the hand-back safe to perform from either side: an [`AllocatedIp`] releasing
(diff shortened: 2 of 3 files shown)


daniel-noland:
Thanks — both blockers confirmed and fixed, plus the three non-blocking notes. Force-pushed as `00963f4f4`.

**`ci::loom` build break.** Reproduced: `cargo check -p dataplane-nat --features loom` failed on `Weak::strong_count`. Worth noting it surfaced as *two* errors, not one — the `E0308 mismatched types` at the same line is a cascade of the failed method resolution, not an independent problem.

I took the "extend the shim" option, for the reason given in the review: `upgrade()` here can genuinely deadlock. `reclaim_ended_tenancies` runs with the pool write lock held, and the temporary `Arc` that `upgrade` mints may be the last one — dropping it re-enters `deallocate_ip` → `pool.write()`. That hazard is why `cleanup` defers through `keep_alive` instead of dropping inline, and reading a count avoids taking a reference at all. The shim is now its own commit (`14808eeb1`), ordered first so each commit builds independently, with the divergence recorded at both the method and module level — including the point that a caller asking "has the last strong reference gone?" is always told "no" under loom, so nobody reads a passing loom run as evidence that path was exercised.

**`ci::check-fmt`.** Fixed; the stray blank line came in with the rebase onto `main`, as diagnosed.

**DCO.** Both commits now carry `Signed-off-by`.

**Non-blocking, all three taken:**
- The "more specific of the two" comment overclaimed — the residual arm doesn't compare specificity. Reworded to say what the code does, and to note every branch stays inside the exhaustion class so `is_exhaustion` callers are unaffected.
- Documented why the broad `is_exhaustion()` arm in `reuse_allocated_ip` can't swallow `NoFreeIp`: allocating a port for an address the caller already holds has no address to run out of, so the port allocator never returns it. That fact isn't local to the function, which is exactly why it needed writing down.
- Left `reclaim_ended_tenancies` and `cleanup` separate rather than factoring out a shared helper — the divergence is the point, so I documented it instead. Sharing them would invite someone to unify on `upgrade` and reintroduce the deadlock.

Verified on `00963f4f4`: `--features loom` compiles, `cargo fmt --all --check` clean, 165 lib tests, 8 shuttle tests, clippy clean across `dataplane-nat` and `dataplane-concurrency`.

One correction to the record: the review noting "no workspace files were changed" is no longer true — the loom shim touches `concurrency/`. That's unavoidable given the fix needs a liveness check that is safe to call under the lock.

daniel-noland:
Superseded by #1744, opened from an `origin` branch so CI will actually run. Same head commit (`00963f4f4`); the review outcomes from this thread are summarised there.

PR #1744, opened 2026-08-21
fix(masquerade): Close the window where an address belongs to neither list

daniel-noland:
## Issue

`NatPool` answers "is this public address available?" from two structures — a bitmap of free addresses and a list of weak references to the addresses currently in use — and an address is meant to be in exactly one of them. Releasing one passes through a state where it is in neither: `Weak::upgrade` stops resolving the instant the last `AllocatedPort` drops the strong count to zero, but the address only returns to the bitmap later, inside `AllocatedIp::drop`, which must first acquire the pool's write lock. An allocation arriving in that gap finds nothing on either path and is told the pool is empty, so it fails with `NoFreeIp` while the pool is in fact holding an address that belongs to no one. This is not a rare interleaving — `Arc` guarantees the ordering, so every release passes through the gap; only its duration varies, bounded by how long the releasing thread waits for a write lock that every allocation also takes, on every call, via `cleanup_used_ips`. The same reasoning covers a second latent hazard the fix closes: `reserve_from_pool` could create a duplicate `AllocatedIp` for an address whose owner had not yet finished releasing it, after which the first object's `Drop` would free an address the second still held.

## Impact

The packet that lands in the gap is dropped with `DoneReason::NatOutOfResources` and its flow is refused, logged as `masquerade: Ip/port allocation failed ... no free IP available`. Two properties of real deployments turn a narrow race into a recurring one. Where an expose is configured with a single public address per peering — the shape observed in the field — the unavailable address is the entire pool, so there is nothing to fall back on. And DNS drives the window open constantly: a reply from source port 53 moves the flow to `Closed` and invalidates the pair immediately as a port-conservation measure, so a pool carrying mostly DNS destroys and rebuilds its address roughly once per query, opening one window per query. The user-visible effect is mild and easily misread — resolvers retry, so lookups still succeed and the failures surface only as intermittent warnings and a raised drop count, at a rate that does not correlate with load in any obvious way.

## Fix

The root problem is that liveness was inferred from `Arc` refcounts, which change on whatever thread happens to drop the last reference, while availability is a property of the pool, which changes only under its lock — so the two could disagree. The fix gives every lease of an address a `Tenancy`, recorded with the address's bitmap offset alongside the weak reference, and keeps a map of current tenancies that is updated in the same critical section as the bitmap itself. That makes the pool's own state the source of truth and lets the hand-back be completed from either side: an allocation that holds the lock and finds a dead weak reference finishes the hand-back itself rather than waiting for the releasing thread to win the lock, and `use_new_ip` reclaims ended tenancies before it reports an empty bitmap — in the same critical section as the retry, so the answer cannot go stale between the two. A late `AllocatedIp::drop` whose tenancy has since been retired does nothing, which is what makes that takeover safe and, incidentally, what closes the duplicate-`AllocatedIp` hazard in `reserve_from_pool`.

Two smaller corrections ride along, both about being able to see what happened: the reuse scan no longer abandons the search at the first address that has run out of port blocks, stranding every address behind it in the list, and `NoFreeIp` no longer buries the more specific reason — which, on a single-address pool, it always did.

## Testing

- New property test in the existing shuttle harness, `an_allocation_racing_the_last_release_is_still_served`: a one-address pool with one thread releasing the only allocation while another allocates. Against the unfixed allocator it reproduces the field error verbatim (`NoFreeIp`); against this change it passes.
- `cargo test -p dataplane-nat --lib` — 165 pass.
- `cargo test -p dataplane-nat --features shuttle --lib shuttle` — 8 pass.
- `cargo clippy -p dataplane-nat --all-targets` — clean.
- Confirmed on a live gateway by deliberately widening the window: the unpatched allocator refused 10 of 20 queries with the drop counter and the log agreeing exactly, while this change served all 20 under the identical forcing function.

## Notes

- Not addressed here: ports are handed out in 256-blocks owned by a single thread and freed only when the last port in the block dies. Measured density on the affected fabric was ~3.35 ports/block for UDP, so a pool walls well below its nominal capacity. That is a separate redesign.
- A backport of this change onto `v0.25.2` is prepared on `backport/v0.25.2/masquerade-address-handback`; it is semantically identical (the two differ only by one clarifying comment).

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,0 +1,35 @@
+# Configuration for cargo-mutants (https://mutants.rs).
+#
+# Mutation testing is a flashlight, not a gate: it is run by hand over a crate
+# or a diff, and the product is the list of survivors rather than the score.
+# Nothing in CI depends on this file.
+#
+# Two categories are excluded because a survivor there is noise rather than a
+# finding. Both are excluded here rather than with `#[mutants::skip]` so that
+# production crates take no dependency on the tool and the reasoning stays in
+# one place.
+
+exclude_re = [
+    # Printers. The standing rule is "don't test printers": asserting on
+    # rendered output is sisyphean, produces low signal, and breaks whenever
+    # anyone adjusts a column. A mutated `fmt` that no test notices is telling
+    # us the rule is being followed.
+    "<impl Display for",
+    "<impl Debug for",
+    "<impl std::fmt::Display for",
+    "<impl std::fmt::Debug for",
+
+    # Test-support code inside the crate under test. The workspace convention
+    # puts bolero generators in a `contract` module next to the type they
+    # generate, so cargo-mutants finds them and mutates the test harness at
+    # itself. A generator that draws a different distribution is not a defect
+    # in the code under test; the properties are what judge it.
+    "contract::",
+]
+
+exclude_globs = [
+    # Privileged sysfs manipulation. Exercising the write path needs root and a
+    # real sysfs, and the read path is thin enough that a property would assert
+    # little beyond what the type system already does.
+    "sysfs/**",
+]
--- a/concurrency/src/sync/loom_backend.rs
+++ b/concurrency/src/sync/loom_backend.rs
@@ -28,6 +28,9 @@
 //!     returns `Some` -- the upgrade-fails-after-last-strong-drop race
 //!     real `Weak` exposes is not modelled.
 //!   * `Arc::strong_count` reflects live `Arc`s **and** live `Weak`s.
+//!   * `Weak::strong_count` likewise never reaches `0` while the `Weak`
+//!     itself is alive, so a caller using it to detect "the last strong
+//!     reference is gone" never observes that condition under loom.
 //!   * `Arc::weak_count` panics: the shim has no separate weak count
 //!     to report, and returning `0` silently would make assertions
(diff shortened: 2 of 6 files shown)


daniel-noland:
Added `ci:+concurrency`, and it is worth saying why rather than just doing it.

The `concurrency` job — which is what runs `ci::loom` and `ci::shuttle` (`.github/workflows/dev.yml:668-687`) — is gated on `needs.plan.outputs.concurrency == 'true'`, which comes from a label check. So it does not run by default, and on the first run here it reported `skipped`.

That means the `--features loom` build break reported in review would not have been caught by this PR's own CI. It was found by a reviewer building locally. Two consequences worth noting:

- This PR needs the label, because the fix depends on a change to the loom shim and the whole claim is that loom now compiles. Without it, CI would go green while never touching either half of that.
- More generally: `alloc.rs` is production code compiled in every configuration, but the configuration that broke is opt-in. Anything touching `concurrency/` or the allocator is a candidate for the same blind spot. Not something to solve in this PR, but probably worth a look at whether the gate should also trip on paths rather than labels alone.

Local results on `00963f4f4` for the two recipes that job runs — `cargo check -p dataplane-nat --features loom` compiles, and the shuttle suite is 8/8 including the regression this fix is built around — but local is not CI, which is the point of the label.

PR #1751, opened 2026-08-26
tools

daniel-noland:


justfile
```diff
@@ -751,6 +751,45 @@ coverage *args:
...
+    duvet report {{ args }}
+
+# Fail if the compliance snapshot is out of date
+[script]
+duvet-check:
+    {{ _just_debuggable_ }}
+    # `duvet report` takes milliseconds and is bit-for-bit deterministic, so unlike mutation
+    # testing this can be a gate, and it is the cheapest correctness check in the repo. It is
+    # one because the snapshot had already drifted two commits after being introduced: a
+    # regenerate-by-hand rule is one nobody runs.
```
daniel-noland: we can skip this comment

scripts/spec-interlock.ts
```diff
@@ -0,0 +1,487 @@
+#!/usr/bin/env -S deno run --allow-read --allow-run --allow-write
```
daniel-noland, line 1: this logic is functionally correct, but it is missing the level of abstraction I would prefer.  Noting it for a rework

CLAUDE.md
```diff
@@ -4,3 +4,21 @@ Follow the [development guide](development/README.md).
...
+A commit message signposts the diff; it does not restate it. Rationale belongs where it
+constrains the code -- in the code, its doc comments, or the development guide -- because a
+second copy in the message is a second thing to keep correct, and the two will drift.
+
+A message carries what the diff cannot:
+
+- the "why" of the commit,
+- non-obvious process facts,
+
+It does not re-summarize the change, narrate how it was made, or repeat an argument the
```
mvachhar: THe commit should have a line about what the change is in addition to why, but yes, a long narrative about the diff is not really needed.  The why is more important because it isn't in the comments, code, or docs.

nix/overlays/dataplane-dev.nix
```diff
@@ -12,6 +12,12 @@ let
   };
 in
 {
+  # callPackage rather than a bare import: duvet links against OpenSSL and so needs pkg-config and
```
mvachhar: What are the build reproducibility implications of this? I figure this only matters on a manual run of duvet to extract new requirements, but checking is deterministic and tag based still?

scripts/spec-interlock.ts
```diff
@@ -43,6 +43,49 @@ const FN_LINE = /\bfn\s+([A-Za-z_][A-Za-z0-9_]*)/;
...
+      "Equivalent under the allocator's invariants. `reuse_allocated_ip` skips `NoFreePort` " +
+      "and loops, so the only error it can return from a well-formed pool is `NoFreeIp`, " +
+      "which is exhaustion; the non-exhaustion arm is defensive depth against an internal " +
+      "inconsistency. The whole 210-test nat suite passes with the guard forced to `true`. " +
+      "The guard is kept because a future allocator that can fail for a non-exhaustion reason " +
+      "must not draw a second public address for a host that already holds one.",
+  },
+];
+
+/** cargo-mutants' stable name for a mutant: its `<path>: <what>`, dropping line and column. */
```
mvachhar: I don't understand what this commit is suppsoed to do and the comments read like AI slop.  Could you please explain this manually.

scripts/spec-interlock.ts
mvachhar, line 4: The second paragraph of the commit message for `feat(spec): Say why a mutant survived, using coverage` is not intelligible.  What is it trying to say?  I understand the first paragraph.

scripts/spec-interlock.ts
```diff
@@ -127,14 +127,108 @@ interface Report {
...
+    code: output.code,
+    stdout: decode(output.stdout),
+    stderr: decode(output.stderr),
+  };
+}
+
+/**
+ * How many times the cited tests executed each line of the files they are cited against.
+ *
+ * Cheap next to mutation -- one instrumented run rather than one build and test cycle per
```
mvachhar: I'm not sure we need to spill so much text over why this is better than using mutants to check if we have coverage.  We are comparing to an alternative that was only considered during development.  A sentence that says using coverage to make sure we test what we think we're testing is cheaper than using mutants, so we use that.  I think that is what the second paragraph in the commit `feat(spec): Say why a mutant survived, using coverage` is trying to say in its AI slop way.

scripts/spec-interlock.ts
mvachhar, line 2: The commit message for `fix(spec): Accept the second-half boundary the block guard makes unreachable` is not intelligible.

mvachhar (changes requested):
A few minor changes and a question.  Please at least read the commit messages and comments and see if they make sense.  Some of them sound like gibberish.

Otherwise, good to go.

scripts/spec-interlock.ts
daniel-noland: sure.  It's just talking about the phrasing of the mutants.toml file.

This method chops the line and column number from the name of the mutation in the file because they are not refactor stable.

nix/overlays/dataplane-dev.nix
daniel-noland: oh, this is just building the duvet binary itself.  Both the build of duvet and what duvet builds are stable

scripts/spec-interlock.ts
daniel-noland: I deleted virtually all comments from this PR.

I also rewrote every commit message to be very terse

daniel-noland: deleted

daniel-noland:
> A few minor changes and a question. Please at least read the commit messages and comments and see if they make sense. Some of them sound like gibberish.
> 
> Otherwise, good to go.

I am just no longer going to add in comments.  Nobody else does and they are just distracting anyway

CLAUDE.md
daniel-noland: deleted

mvachhar (approved):
Still a bit more AI slop in the commit comments than I would like, but it is acceptable for now.

PR #1752, opened 2026-08-26
net

daniel-noland:


net/src/headers/embedded_view.rs
```diff
@@ -1966,9 +1966,15 @@ mod embedded_view_properties {
...
-        // bites in CI; a deliberately short run should say nothing rather than say something false.
-        if seen > 20_000 {
+        // time it ran. A deliberately short run should say nothing rather than say something false.
+        //
+        // Ten thousand, not the twenty-five thousand a default run manages here. The thinnest shape
+        // in this module draws about eight hits per ten thousand, and coverage instrumentation costs
+        // about a fifth of the throughput -- measured at 17 hits in 20,206 cases against 21 in
+        // 25,857 without. A threshold set at the observed count would therefore switch itself off on
+        // any busier runner, silently. At ten thousand draws a shape that really is drawable misses
+        // entirely about three times in ten thousand runs, which is the trade this wants.
```
daniel-noland: this comment has historical process notes which should be trimmed.  The selected lines can be removed

net/src/headers/embedded.rs
```diff
@@ -1378,6 +1419,107 @@ mod tests {
...
+                    &v4_net(new, NextHeader::TCP),
+                    &PAYLOAD,
+                ))
+                .unwrap_or_else(|()| unreachable!());
+            assert_eq!(quoted.checksum(), Some(u16::from(expected)));
+        }
+
+        /// Eight words to fold, over a UDP pseudo-header.
+        #[test]
+        fn a_v6_address_change_matches_a_fresh_udp_checksum() {
```
mvachhar, line 1461: We need a test for the incremental checksum update where the quoted packet header is truncated below the required 128 bytes.  We won't ever generate such a packet but we can receive one.  What happens in that case when we try to update the checksum?

net/src/headers/view.rs
```diff
@@ -2575,3 +2575,76 @@ mod tests {
...
     }
 }
+
+/// Differential checks for [`HeadersView`]'s unsafe extraction boundary.
+///
+/// [`sealed::Sealed::matches`] licenses [`Look::look`] to use `unwrap_unchecked`. These tests compare
+/// that decision with the safe [`Matcher`](super::pat::Matcher) before extraction, then compare layer
+/// addresses so equal-valued duplicate layers remain distinguishable.
+#[cfg(test)]
+mod view_properties {
```
mvachhar, line 2580: Why this peculiar use of the work licenses.  I am not quite sure what it even means in this context.

net/src/headers/view.rs
```diff
@@ -2661,52 +2661,169 @@ mod view_mut_properties {
...
-            .with_generator(ShapedHeaders)
-            .for_each(|h: &Headers| {
-                let mut owned = h.clone();
-                let licensed = owned.as_view_mut::<(&Eth, &Net, &Transport)>().is_some();
-                let deliverable = owned.pat_mut().eth().net().transport().done().is_some();
-                assert_eq!(
-                    licensed, deliverable,
-                    "`matches` and the mutable walk disagree, so `look_mut` would reach \
-                     `unreachable_unchecked`: {h:?}"
+    /// The decision is what soundness turns on: if `matches` says yes where the walk that must deliver
```
mvachhar: I hate this phrasing, if you are changing other things, maybe say this like a normal person would?

mvachhar: We're just making sure that matches and look are consistent.  This is a really obtuse way of saying that.

net/src/headers/embedded_view.rs
```diff
@@ -1790,3 +1790,560 @@ mod tests {
...
+mod embedded_view_properties {
+    use super::*;
+    use crate::eth::Eth;
+    use crate::headers::{Headers, ShapedIcmpError, ShapedQuote};
+    use crate::icmp4::Icmp4;
+    use crate::icmp6::Icmp6;
+    use crate::vlan::Vlan;
+
+    /// Compare embedded-view licensing with the safe matcher walks across outer and inner arities.
+    ///
```
mvachhar: Please explain what licensing means in all these commits.  "I don't think that word means what you think it means" - Inigo Montoya, from "The Princess Bride"

mvachhar (changes requested):
I think we need one extra test regarding checksums.  Might as well fix the wierd licensing verbiage while you're at it.

net/src/headers/view.rs
daniel-noland: deleted

net/src/headers/embedded_view.rs
daniel-noland: deleted

net/src/headers/view.rs
daniel-noland: deleted

net/src/headers/embedded.rs
daniel-noland: Nothing happens tho. The truncated quote ships exactly as received.

I can write a test to assert that if you like

daniel-noland: done

net/src/headers/embedded.rs
```diff
@@ -612,39 +613,64 @@ impl EmbeddedTransport {
...
         match self {
             EmbeddedTransport::Tcp(tcp) => {
-                // Silently ignore errors if transport header is truncated
-                let _ = tcp.increment_update_checksum(
+                let updated = tcp.increment_update_checksum(
                     TcpChecksum::new(current_checksum),
                     old_value,
                     new_value,
                 );
+                let _ = tcp.set_checksum(updated);
```
qmonnet, line 621: Looks like a better fix would be to make `increment_update_checksum()` update in-place instead? The method already takes a `&mut self` (but doesn't use it). This should also fix `increment_update_checksum_32bit()` at the same time, and the `increment_update_checksum()` call from `translate_inner_icmp()` in icmp_error_msg.rs, where we seem to have the same issue as here?

qmonnet (dismissed):

mvachhar (approved):
Do we want to roll in the fix to generate Ethernet padding on the inner packet here?  It probably doesn't matter yet since we don't generate inner Ethernet frames denovo outside of testing, but it'll help the fuzzer later.

net/src/headers/embedded.rs
mvachhar: Either fix is fine, I'm not sure what semantics we really want here.

mvachhar: I also would not hold up the PR over this.  @qmonnet we are trying to cut a release this week in order to address a bunch of issues and add features.

qmonnet: OK, I can address the other two call sites as a follow-up if necessary

daniel-noland: I actually agree with you @qmonnet but I was trying to be as minimally invasive as possible.  Changing the return type was out of bounds for this PR in my mind.

PR #1753, opened 2026-08-26
(05) route

daniel-noland:


routing/src/rib/nexthop.rs
```diff
@@ -263,9 +280,14 @@ impl Nhop {
         self.resolvers.replace(resolvers);
     }
 
-    /// Auxiliary recursive method used by `Nhop::quick_resolve()`.
+    /// Recursively collect resolved keys without revisiting next-hops.
     #[cfg(test)]
-    fn quick_resolve_rec(&self, result: &mut BTreeSet<NhopKey>) {
+    fn quick_resolve_rec(&self, result: &mut BTreeSet<NhopKey>, visited: &mut Visited) {
+        if visited.contains(&self.id()) {
```
mvachhar, line 279: No need to fix but for really deep (say 1k+ routes) the performance here might start to suffer.  Isn't there a rust lib to have a small hash table that does a linear array search below a certain size?

routing/src/fib/fibtable.rs
```diff
@@ -38,10 +38,10 @@ impl FibTable {
         info!("Registering Fib with id {id} in the FibTable");
         self.entries.insert(id, entry);
     }
-    /// Delete a `Fib`, by unregistering a `FibReaderFactory` for it
+    /// Delete a `Fib` and every alias that refers to it.
     fn del_fib(&mut self, id: FibKey) {
         info!("Unregistering Fib with id {id} from the FibTable");
-        self.entries.remove(&id);
+        self.entries.retain(|_, entry| entry.id != id);
```
mvachhar, line 43: So this is a big change, are you certain that there is no place where we depend on just removing the thing indexed by id, but instead want to remove all things with a particular id?

routing/src/cli/display.rs
```diff
@@ -168,7 +175,12 @@ fn fmt_nhop_instruction(f: &mut std::fmt::Formatter<'_>, rc: &Nhop) -> std::fmt:
 
 // formats nhop using the display of the key, recoursing over resolvers
```
mvachhar: Not relevant to the PR, but recoursing isn't the right verb here, recurring is the gerund form for the adjective recursive since they both come from the verb recur.

routing/src/cli/display.rs
```diff
@@ -206,7 +221,7 @@ impl Display for NhopStore {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         Heading(format!("Next-hop Store ({})", self.len())).fmt(f)?;
         for nhop in self.iter() {
-            fmt_nhop_rec(f, nhop, 0)?;
+            fmt_nhop_rec(f, nhop, 0, &mut Visited::new())?;
```
mvachhar, line 221: This seems bad, we are mutating the NhopStore on Display, which is a bad pattern to begin with.  Moreover, it seems we never clear this so two next hops that refer to a common nexthop (no loop) would see the visited flag and show as a loop.

mvachhar: I looked at the tests and I'm not sure it would catch this case either.

mvachhar: Finally, I don't see anywhere that Visited is ever cleared.  Is the NextHop store local to the display logic, or somehow ephemeral?

routing/src/router/cpi.rs
```diff
@@ -209,14 +209,13 @@ impl RpcOperation for IpRoute {
...
         let rmac_store = &db.rmac_store;
         let vrftable = &mut db.vrftable;
-        let iftabler = &db.iftw.as_reader();
 
         if self.vrfid == Vrf::DEFAULT_VRFID {
             let Ok(vrf0) = vrftable.get_vrf_mut(self.vrfid) else {
                 error!("Unable to find default VRF!");
                 return RpcResultCode::Failure;
             };
-            vrf0.add_route_rpc(self, None, rmac_store, iftabler);
```
mvachhar, line 219: Are there reasons we want the interface name?  It seems to me that you could get two next hops that are identical apart from the interface name and they might have two different priorities.  Why is this change ok?

mvachhar (changes requested):

routing/src/router/cpi.rs
Fredi-raspall: We want the interface name because ifindex is not informative. The interface name is just informational. It does come from the kernel, but via FRR. There are other ways to address this. What is the issue that led the removal from the key?

routing/src/fib/fibtype.rs
```diff
@@ -504,3 +504,347 @@ impl FibReaderFactory {
         FibReader(self.0.handle())
     }
 }
+
+#[cfg(test)]
+mod fib_properties {
```
Fredi-raspall, line 512: I think this fuzzing adds little coverage to what we actually do at the fib level and my feeling is that we are not exercising the actual code.

routing/src/fib/fibtable.rs
```diff
@@ -38,10 +38,9 @@ impl FibTable {
         info!("Registering Fib with id {id} in the FibTable");
         self.entries.insert(id, entry);
     }
-    /// Delete a `Fib`, by unregistering a `FibReaderFactory` for it
     fn del_fib(&mut self, id: FibKey) {
         info!("Unregistering Fib with id {id} from the FibTable");
-        self.entries.remove(&id);
```
Fredi-raspall, line 44: This is not a fix but a sanity and that should be stated in the commit. Else, we don't know if we actually fixed a problem or not. I'm not against the change. However, it makes the logic implicit and retain requires checking every entry in the fibtable, O(N). The prior approach (dummy fibtable, caller decides), would just call remove() and unregister_by_vni (again remove()).

routing/src/fib/fibtable.rs
```diff
@@ -38,10 +38,9 @@ impl FibTable {
         info!("Registering Fib with id {id} in the FibTable");
         self.entries.insert(id, entry);
     }
-    /// Delete a `Fib`, by unregistering a `FibReaderFactory` for it
```
Fredi-raspall, line 41: Why deleting this comment? It's not 100% accurate, but still ...

routing/src/rib/vrftable.rs
```diff
@@ -176,6 +176,13 @@ impl VrfTable {
         vrfid: VrfId,
         iftablew: &mut IfTableWriter,
     ) -> Result<(), RouterError> {
+        if vrfid == Vrf::DEFAULT_VRFID {
```
Fredi-raspall, line 179: Again this is a sanity, not a fix. I'm fine keeping it. With the current code, only vrfs in status deleted could be removed. The default vrf can never be in such a state.

routing/src/rib/vrf.rs
```diff
@@ -113,7 +113,7 @@ impl ShimNhop {
     }
 }
 
-#[derive(Copy, Clone, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
Fredi-raspall, line 117: Why do we need Debug?

routing/src/router/rpc_adapt.rs
```diff
@@ -221,12 +221,8 @@ impl Vrf {
             }
         }
 
-        // If no next-hop was received with the route (or we could not successfully process any),
-        // install the route anyway with an action drop. This is better than not installing the
-        // route as that could break consistency (e.g. resolving via a default) and cause a loop.
         if nhops.is_empty() {
-            warn!("Route to {prefix} from RPC would have no next-hop. Will inject DROP next-hop");
-            nhops.push(RouteNhop::default());
```
Fredi-raspall, line 229: Again, not a fix but a refactor to test the prod code against the model.
I don't quite like the nhops_or_drop(). Will push an alternative.

routing/src/router/rpc_adapt.rs
```diff
@@ -128,22 +123,13 @@ impl RouteNhop {
             None => None,
         };
 
-        // lookup interface name
-        let ifname = match ifindex {
```
Fredi-raspall, line 132: I don't like getting rid of interface names from the next-hops since we lose the ability to see them in the routing tables. There may be better ways to do this, no question, but removing them from the next-hops has more drawbacks than benefits: two identical next-hops that differed only by the interface names would cause no issue. So, I'd not include this change.

routing/src/router/rpc_adapt.rs
```diff
@@ -128,22 +123,13 @@ impl RouteNhop {
...
-
-        // build key for this next hop
         let key = NhopKey::new(
             origin,
             nh.address,
             ifindex,
             encap,
             FwAction::from(nh.fwaction),
-            ifname,
+            None,
```
Fredi-raspall, line 132: If we ended up removing the ifname, this constructor should be changed and the ifname field be removed from NhopKey.

routing/src/frr/renderer/mod.rs
```diff
@@ -66,3 +66,468 @@ impl Render for InternalConfig {
         cfg
     }
 }
+
+#[cfg(test)]
+mod renderer_properties {
```
Fredi-raspall, line 71: I don't think we should use any fuzzing to test the rendering. I would drop these tests.

Fredi-raspall (approved):
I think this PR adds some sanities which are fine, and fixes an issue receiving from the frr-agent. However, there are some changes that I dislike:
* we lose the ifname (partially) from the next-hops. We won't see them in the cli. That's bad. I would drop that commit.
* I'd reconsider the removal of fibs based on retain(). To me, it adds little benefit.
* I'm concerned about the testing strategy:  I believe the tests as they are are hard to maintain, add a lot of code (sometimes more code than the impl itself) while not exercising the actual critical/sensitive  parts of the implementation.

I'm approving the PR counting that the ifname will be kept (I consider it otherwise a regression) for the sake of timing and so as not to block other PRs, with the understanding that other testing strategies may replace them.

routing/src/frr/renderer/mod.rs
daniel-noland: the problem was hunting deadlock as I recall

Fredi-raspall: In the rendering?

routing/src/fib/fibtable.rs
mvachhar: No this fixes a problem the fuzzer found.  Unfortunately there is no documentation as to what the exact fuzzer failure was.  Fredi, what is the correct behavior here?

routing/src/frr/renderer/mod.rs
mvachhar: Yes, in the rendering.

routing/src/rib/vrf.rs
mvachhar: Why not?

routing/src/frr/renderer/mod.rs
Fredi-raspall: I doubt very much the issue had to do with the rendering.

mvachhar:
If @Fredi-raspall has approved these changes then this works for me.

mvachhar (approved):

qmonnet:
> I'm approving the PR counting that the ifname will be kept (I consider it otherwise a regression)

Was this addressed? @Fredi-raspall are we good to click the merge button?

Fredi-raspall:
> > I'm approving the PR counting that the ifname will be kept (I consider it otherwise a regression)
> 
> Was this addressed? @Fredi-raspall are we good to click the merge button?

It's addressed in a separate PR. The (6) ip PR also removes those from egress objects.
We need to decide what we want to do:
   1) leave Nhops and egress objects and the like without the ifname
   2) keep them in those objects.

If we care only about seeing those in CLI, we may recover them  by looking them up each time from the iftable.
So, from that perspective, 1) and 2) are the same. If we care about those showing up in logs, then we need 2), unless we want to overcomplicate things.

qmonnet:
> We need to decide what we want to do:
> 
>  1. leave Nhops and egress objects and the like without the ifname
> 
>  2. keep them in those objects.

My understanding is that keeping them as-is in the egress objects causes an issue, from Daniel's description: “Kernel interface names can change while the ifindex and address still describe the same next hop. Including the name in NhopKey created duplicate identities and could make deletes miss installed routes.” So we'd need to change it anyway?

@mvachhar Do you have an opinion?

PR #1755, opened 2026-08-26
fuzz

daniel-noland:


nat/src/static_nat/probe.rs
```diff
@@ -0,0 +1,467 @@
...
+
+/// Build a packet with the given five-tuple.
+///
+/// # Panics
+///
+/// Panics if the two addresses are of different families. Resolution never mixes them, since an
+/// expose is of one family throughout and the peer prefix is chosen to match.
+pub(crate) fn build(
+    source: IpAddr,
+    destination: IpAddr,
```
daniel-noland, line 319: better to avoid the use of IpAddr if we can just use an IpAddress generic instead.  Cleaner

nat/src/static_nat/probe.rs
```diff
@@ -0,0 +1,467 @@
...
+    }
+}
+
+/// Build a packet with the given five-tuple.
+///
+/// # Panics
+///
+/// Panics if the two addresses are of different families. Resolution never mixes them, since an
+/// expose is of one family throughout and the peer prefix is chosen to match.
+pub(crate) fn build(
```
daniel-noland, line 317: this function should likely be rebuilt using the pat.rs / view.rs mechanics we use elsewhere

PR #1773, opened 2026-08-28
(12) test-clock

daniel-noland:
ready for review modulo trivial CI noise.


nat/src/portfw/test_expiry.rs
```diff
@@ -15,19 +16,7 @@ use std::net::IpAddr;
 const WITHIN_LIFETIME: Duration = Duration::from_secs(1);
 
 fn with_paused_clock<F: Future<Output = ()>>(body: impl FnOnce() -> F) {
-    let runtime = tokio::runtime::Builder::new_current_thread()
```
mvachhar, line 18: Can we rename both the port foward and masquerad expiry files to test_expiry.rs so it is clear that they are test files?  Sorry if I missed this on previous reviews.

clock/src/lib.rs
```diff
@@ -16,7 +20,7 @@ pub mod virtual_time;
 pub fn now() -> Instant {
     #[cfg(all(feature = "virtual", not(wall_clock)))]
     {
-        checked_now().unwrap_or_else(virtual_time::refuse)
+        checked_now().unwrap_or_else(|| virtual_time::refuse())
```
mvachhar, line 23: Why does this now need an explicit call instead of closure.  Not a blocker but just curious.

dataplane/src/packet_processor/fuzz.rs
```diff
@@ -718,6 +718,23 @@ pub(crate) struct Pick {
 #[cfg(test)]
 pub(crate) type Poll = Vec<Pick>;
 
+#[cfg(test)]
```
mvachhar, line 774: Not for this PR, but do we want to move all files that are gated by `#[cfg(test)]` test directory in the current directory?  If so, my prior comment about renaming `expiry.rs` to `test_expiry.rs` would not apply, we'd just move it inot the test subdirectory.

dataplane/src/packet_processor/fuzz.rs
```diff
@@ -718,6 +718,23 @@ pub(crate) struct Pick {
...
+    static RUNTIME: std::sync::LazyLock<tokio::runtime::Runtime> = std::sync::LazyLock::new(|| {
+        tokio::runtime::Builder::new_current_thread()
+            .enable_all()
+            .build()
+            .expect("build tokio runtime")
+    });
+
+    RUNTIME.block_on(async {
+        body();
+        for _ in 0..4 {
```
mvachhar: Is there any logic to exactly 4 yields?  Why not 1, or 10?  Or maybe a driven number by bolero?

.github/workflows/dev.yml
```diff
@@ -129,7 +129,7 @@ jobs:
         uses: *gate
         with:
           labels: "test/all-profiles"
-          on-value: '["debug", "release", "fuzz"]'
```
mvachhar, line 132: With this change, could you add something to the README or elsewhere on how to separately enable fuzzing or coverage, for a non-nix person, it isn't obvious to me reading the diff.

mvachhar (changes requested):

nat/src/portfw/test_expiry.rs
daniel-noland: done

clock/src/lib.rs
daniel-noland: now that you mention it, this is odd.  I didn't even think about it when I read it.

daniel-noland: oh, now I remember.  It is because the refuse method returns the never type.  If we don't put it in the closure it would need to return instant.  Kinda an odd hack, but it fixes the issue.

Basically, we _need_ to panic there and so the return type getting semi-conflated by the closure is helpful.

dataplane/src/packet_processor/fuzz.rs
daniel-noland: This is reworked based on analysis of your question.

The answer wasn't "4 was the wrong number."  It’s that the yields were inert.  They never did what I was trying to do in the first place.

- No sampled drains saw the live-task count change at all
- No tasks ever completed during a drain
- The prop test was bugged.  It had a linearly growing number of tasks.

FlowTable::insert spawns a timer per flow that ends only on its deadline or its cancellation. 
But dropping the table cancels nothing.

The fix is in aea2d4da91: Fabric::drop retires the case’s flows via FlowTable::remove.
That cancels each token, settles, then drains to zero instead of yielding a fixed number of times.
It also panics if it can’t.

.github/workflows/dev.yml
daniel-noland: will do.

daniel-noland: done.

mvachhar (approved):

PR #1778, opened 2026-08-28
fix(net): Bound the transport checksum by the network header's length

daniel-noland:
## The defect

`Packet::new` kept everything after the parsed headers as the payload, and nothing ever trimmed the buffer to the IPv4 total length / IPv6 payload length (`TrimFromEnd` existed but was never called). `do_serialize` then handed that whole slice to the checksum code, so **ethernet padding was summed as if it were L4 payload**.

Ethernet pads frames out to 64 octets including the Ethernet header, so any IPv4 frame under that was affected:

| protocol | pseudo header length source | effect of padding |
|---|---|---|
| TCP | `header_len + payload.len()` | **always wrong**, off by the pad length, even for zeroed padding |
| ICMPv6 | `payload.len() + header_len` | same, but IPv6 frames are ≥ 62 octets so rarely padded |
| UDP | the UDP header's own `length` field | wrong only when the padding is non-zero |
| ICMPv4 | no length in the checksum | wrong only when the padding is non-zero |

The sharp edge is IPv4 TCP: a bare ACK, RST, FIN, or option-less SYN is 54 octets on the wire. Every one of those this dataplane forwarded left with a checksum the far end drops. Flows carrying TCP timestamps produce 66-octet frames and escaped.

### DIFF
--- a/net/src/headers/mod.rs
+++ b/net/src/headers/mod.rs
@@ -890,7 +890,39 @@ impl Headers {
         }
     }
 
+    /// The number of octets which follow the transport header according to the network header.
+    ///
+    /// A frame can carry more octets than its network header accounts for: ethernet pads short
+    /// frames out to 60 octets (`IEEE 802.3` clause 4.2.3.3), and some devices append trailers.
+    /// Those octets belong to no upper layer, so a caller holding a whole frame needs this to find
+    /// where the transport payload ends before computing a checksum over it.
+    ///
+    /// # Returns
+    ///
+    /// Returns `None` when the network or transport header is absent, or when the network header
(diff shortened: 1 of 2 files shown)


mvachhar (approved):
Do we want to have a fix for the serializer when there is a short payload or is that a separate PR?

mvachhar:
Oh, and FYI, the commit message is just wrong.  802.3 includes the header in the frame and size, so the threshold is not a frame size of 60 bytes, but a frame size of 64 bytes.  I didn't check its numbers on the IP packet size.  Annoying.  Fix if you want, I won't hold up the PR for it.

PR #1796, opened 2026-09-03
(01) validator

daniel-noland:
> [!NOTE]
> The remaining CI failure is actually just bad luck.  Pre-existing vacuity guard tripping on qemu.  Will fix, but otherwise this is up for review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1551,6 +1551,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml_ng",
+ "strum",
  "thiserror",
  "tokio",
  "tracing",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1601,6 +1601,7 @@ dependencies = [
  "bnum",
  "bolero",
  "dataplane-match-action",
+ "dataplane-net",
  "ipnet",
  "num-traits",
  "prefix-trie",
@@ -1618,6 +1619,7 @@ dependencies = [
  "bolero",
  "dataplane-fixed-size",
  "dataplane-match-action-derive",
+ "dataplane-net",
 ]
 
 [[package]]
--- a/config/src/external/mod.rs
+++ b/config/src/external/mod.rs
@@ -93,6 +93,48 @@ impl ExternalConfig {
         Ok(())
     }
 
+    /// Refuse a peering *this* gateway would have to render and cannot.
+    ///
+    /// The routing configuration built from a peering is IPv4-only: its prefix lists and
+    /// route-maps are `IpVer::V4` throughout and nothing in the builder has a v6
+    /// counterpart. `build_internal_config` will error out at apply time, after the
+    /// config was accepted. Anything accepted here has to apply cleanly, so the refusal
+    /// belongs here where the validator can call it and reject early.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ConfigError::Unsupported`] if a peering this gateway renders carries
+    /// IPv6 prefixes.
+    fn check_rendered_peerings_are_ipv4(
+        gwname: &str,
+        overlay: &ValidatedOverlay,
+        underlay: &Underlay,
+        gwgroups: &GwGroupTable,
+        communities: &PriorityCommunityTable,
+    ) -> ConfigResult {
+        if underlay.vrf.bgp.is_none() {
+            // No underlay BGP means no overlay routing configuration is built at all.
+            return Ok(());
+        }
+        let renders = |peering: &ValidatedPeering| {
+            gwgroups
+                .get_group_member_rank(peering.gwgroup(), gwname)
+                .is_some_and(|rank| communities.get_community(rank).is_some())
+        };
+        if overlay
+            .vpc_table()
+            .peerings()
+            .filter(|peering| renders(peering))
+            .all(ValidatedPeering::is_v4)
+        {
+            return Ok(());
+        }
+        Err(ConfigError::Unsupported(
+            "IPv6 prefixes in a vpc peering: the routing configuration built from a peering is \
+             IPv4-only, so such a peering cannot be rendered",
+        ))
+    }
+
     /// Validate the external configuration.
     /// This method consumes `ExternalConfig` and outputs a `ValidatedGwConfig` on success.
     ///
@@ -107,6 +149,13 @@ impl ExternalConfig {
         let overlay = self.overlay.validate()?;
         let peerings = overlay.vpc_table().peerings();
         self.check_peering_gwgroups_exist(peerings)?;
+        Self::check_rendered_peerings_are_ipv4(
+            &self.gwname,
+            &overlay,
+            &underlay,
+            &self.gwgroups,
+            &self.communities,
+        )?;
 
         // if there are vpcs configured, there MUST be a vtep configured
         if !overlay.vpc_table().is_empty() && underlay.vtep.is_none() {
--- a/config/src/external/overlay/vpcpeering.rs
+++ b/config/src/external/overlay/vpcpeering.rs
@@ -1184,7 +1184,7 @@ pub mod contract {
                 Side::Private => 0u128,
                 Side::Public => 1,
             };
-            let bits = (0x2001_0db8u128 << 96) | (selector << 80) | (u128::from(index) << 64);
+            let bits = net::ipv6::DOC_PREFIX_BITS | (selector << 80) | (u128::from(index) << 64);
             prefix_v6(bits, 64)
         }
     }
@@ -1254,7 +1254,7 @@ pub mod contract {
                 Side::Private => 0u128,
                 Side::Public => 1,
             };
-            (0x2001_0db8u128 << 96) | (selector << 80)
+            net::ipv6::DOC_PREFIX_BITS | (selector << 80)
         };
 
         let mut out = Vec::with_capacity(parts.len());
@@ -1321,7 +1321,7 @@ pub mod contract {
             .first()
             .and_then(|expose| expose.ips.first().map(PrefixWithOptionalPorts::prefix))
         {
-            Some(Prefix::IPV6(_)) => "2001:db8:ffff::/64",
+            Some(Prefix::IPV6(_)) => net::ipv6_doc!(":ffff::/64"),
             _ => "3.3.3.0/24",
         };
 
@@ -1362,8 +1362,9 @@ pub mod contract {
         Some((prefix_v4(internal, len)?, prefix_v4(external, len)?))
     }
 
-    const INTERNAL_BASE: u128 = 0x2001_0db8_0000_0000_0000_0000_0000_0000;
-    const EXTERNAL_BASE: u128 = 0x2001_0db8_0001_0000_0000_0000_0000_0000;
+    /// Internal and external ranges are told apart by the same selector bit `block` uses.
+    const INTERNAL_BASE: u128 = net::ipv6::DOC_PREFIX_BITS;
+    const EXTERNAL_BASE: u128 = net::ipv6::DOC_PREFIX_BITS | (1 << 80);
 
     fn v6_pair<D: Driver>(driver: &mut D, host_bits: u8) -> Option<(Prefix, Prefix)> {
         let len = 128 - host_bits;
--- a/config/src/external/overlay/vpcpeering.rs
+++ b/config/src/external/overlay/vpcpeering.rs
@@ -1143,6 +1143,85 @@ pub mod contract {
         }
     }
 
+    #[derive(Debug, Clone, Copy, Default)]
+    pub struct StaticNatExpose;
+
+    const MAX_TOTAL_LOG: u8 = 6;
+
+    impl ValueGenerator for StaticNatExpose {
+        type Output = VpcExpose;
+
+        fn generate<D: Driver>(&self, driver: &mut D) -> Option<VpcExpose> {
+            let v4 = driver.produce::<bool>()?;
+            let total_log = driver.gen_u8(Included(&0), Included(&MAX_TOTAL_LOG))?;
+
+            let privates = place(v4, Side::Private, &split(driver, total_log)?)?;
+            let publics = place(v4, Side::Public, &split(driver, total_log)?)?;
+
+            let mut expose = VpcExpose::empty().make_static_nat().ok()?;
+            for prefix in privates {
+                expose = expose.ip(PrefixWithOptionalPorts::new(prefix, None));
+            }
+            for prefix in publics {
+                expose = expose
+                    .as_range(PrefixWithOptionalPorts::new(prefix, None))
+                    .ok()?;
+            }
+            Some(expose)
+        }
+    }
+
+    fn split<D: Driver>(driver: &mut D, total_log: u8) -> Option<Vec<u8>> {
+        let mut parts = vec![total_log];
+        for _ in 0..driver.gen_u8(Included(&0), Included(&3))? {
+            let splittable: Vec<usize> = parts
+                .iter()
+                .enumerate()
+                .filter(|(_, log)| **log > 0)
+                .map(|(index, _)| index)
+                .collect();
+            if splittable.is_empty() {
+                break;
+            }
+            let choice = usize::from(driver.gen_u8(
+                Included(&0),
+                Included(&u8::try_from(splittable.len() - 1).ok()?),
+            )?);
+            let log = parts.swap_remove(splittable[choice]);
+            parts.push(log - 1);
+            parts.push(log - 1);
+        }
+        parts.sort_unstable_by(|a, b| b.cmp(a));
+        Some(parts)
+    }
+
+    fn place(v4: bool, side: Side, parts: &[u8]) -> Option<Vec<Prefix>> {
+        let mut cursor = if v4 {
+            u128::from(match side {
+                Side::Private => 0x0A00_0000u32,
+                Side::Public => 0xAC10_0000,
+            })
+        } else {
+            let selector = match side {
+                Side::Private => 0u128,
+                Side::Public => 1,
+            };
+            (0x2001_0db8u128 << 96) | (selector << 80)
+        };
+
+        let mut out = Vec::with_capacity(parts.len());
+        for &log in parts {
+            let prefix = if v4 {
+                prefix_v4(u32::try_from(cursor).ok()?, 32 - log)?
+            } else {
+                prefix_v6(cursor, 128 - log)?
+            };
+            out.push(prefix);
+            cursor += 2u128 << log;
+        }
+        Some(out)
+    }
+
     pub const LOCAL_VNI: u32 = 100;
     pub const REMOTE_VNI: u32 = 200;
 
@@ -1255,6 +1334,27 @@ pub mod contract {
                         "could not build an overlay around {shown}"
                     );
                 });
+        }
+
+        #[test]
+        fn every_generated_static_nat_expose_validates() {
+            let mut shapes_differed = false;
+            bolero::check!()
+                .with_generator(StaticNatExpose)
+                .for_each(|expose: &VpcExpose| {
+                    let validated = expose.validate().unwrap_or_else(|e| {
+                        panic!("generated expose was rejected: {expose} -- {e:?}")
+                    });
+                    assert!(validated.has_static_nat());
+                    if validated.ips().len() != validated.as_range_or_empty().len() {
+                        shapes_differed = true;
+                    }
+                });
+            assert!(
+                shapes_differed,
+                "no generated expose had a different number of prefixes on each side, so the \
+                 mapping was never asked to fragment"
+            );
         }
 
         #[test]
--- a/config/src/external/overlay/vpcpeering.rs
+++ b/config/src/external/overlay/vpcpeering.rs
@@ -451,12 +451,6 @@ impl VpcExpose {
                     "Port forwarding requires a single prefix on each side",
                 ));
             }
-            if ips_sizes != as_range_sizes {
-                return Err(ConfigError::MismatchedPrefixSizes(
-                    ips_sizes,
-                    as_range_sizes,
-                ));
-            }
             // For port forwarding, ensure that a port range is always present. Lack of port range would imply
             // all ports, which is not allowed since port 0 is forbidden in the implementation
             for prefixes in [collapsed_expose.ips(), collapsed_expose.as_range_or_empty()] {
@@ -465,6 +459,29 @@ impl VpcExpose {
                         "Port forwarding requires a port range on each prefix",
                     ));
                 }
+            }
+
+            let internal = collapsed_expose
+                .ips()
+                .first()
+                .unwrap_or_else(|| unreachable!());
+            let external = collapsed_expose
+                .as_range_or_empty()
+                .first()
+                .unwrap_or_else(|| unreachable!());
+            if internal.prefix().length() != external.prefix().length() {
+                return Err(ConfigError::MismatchedPrefixLengths {
+                    private: internal.prefix().length(),
+                    public: external.prefix().length(),
+                });
+            }
+            let internal_ports = internal.ports().unwrap_or_else(|| unreachable!());
+            let external_ports = external.ports().unwrap_or_else(|| unreachable!());
+            if internal_ports.len() != external_ports.len() {
+                return Err(ConfigError::MismatchedPortRangeSizes {
+                    private: internal_ports.len(),
+                    public: external_ports.len(),
+                });
             }
         }
 
@@ -1306,6 +1323,59 @@ pub mod contract {
                         validated.err()
                     );
                 });
+        }
+
+        fn forwarding(internal: (&str, u16, u16), external: (&str, u16, u16)) -> VpcExpose {
+            let side = |(prefix, first, last): (&str, u16, u16)| {
+                PrefixWithOptionalPorts::new(
+                    prefix.into(),
+                    Some(PortRange::new(first, last).unwrap_or_else(|_| unreachable!())),
+                )
+            };
+            VpcExpose::empty()
+                .make_port_forwarding(None, None)
+                .unwrap_or_else(|_| unreachable!())
+                .ip(side(internal))
+                .as_range(side(external))
+                .unwrap_or_else(|_| unreachable!())
+        }
+
+        #[test]
+        fn compensating_sizes_do_not_make_a_valid_expose() {
+            let expose = forwarding(("10.0.0.0/32", 1000, 1099), ("172.16.0.0/30", 2000, 2024));
+            assert!(
+                matches!(
+                    expose.validate(),
+                    Err(ConfigError::MismatchedPrefixLengths {
+                        private: 32,
+                        public: 30
+                    })
+                ),
+                "a /32 with 100 ports opposite a /30 with 25 was accepted: {:?}",
+                expose.validate()
+            );
+        }
+
+        #[test]
+        fn port_ranges_of_different_sizes_do_not_make_a_valid_expose() {
+            let expose = forwarding(("10.0.0.0/32", 1000, 1099), ("172.16.0.0/32", 2000, 2049));
+            assert!(
+                matches!(
+                    expose.validate(),
+                    Err(ConfigError::MismatchedPortRangeSizes {
+                        private: 100,
+                        public: 50
+                    })
+                ),
+                "100 ports opposite 50 was accepted: {:?}",
+                expose.validate()
+            );
+        }
+
+        #[test]
+        fn matched_sides_still_make_a_valid_expose() {
+            let expose = forwarding(("10.0.0.0/30", 1000, 1099), ("172.16.0.0/30", 2000, 2099));
+            assert!(expose.validate().is_ok(), "{:?}", expose.validate());
         }
 
         #[test]
--- a/config/src/external/overlay/vpcpeering.rs
+++ b/config/src/external/overlay/vpcpeering.rs
@@ -1243,7 +1243,18 @@ pub mod contract {
     pub const REMOTE_VNI: u32 = 200;
 
     pub fn overlay_offering(expose: VpcExpose) -> Result<ValidatedOverlay, ConfigError> {
-        let remote_prefix = match expose.ips.first().map(PrefixWithOptionalPorts::prefix) {
+        overlay_with(expose)?.validate()
+    }
+
+    pub fn overlay_with(expose: VpcExpose) -> Result<Overlay, ConfigError> {
+        overlay_with_exposes(vec![expose])
+    }
+
+    pub fn overlay_with_exposes(exposes: Vec<VpcExpose>) -> Result<Overlay, ConfigError> {
+        let remote_prefix = match exposes
+            .first()
+            .and_then(|expose| expose.ips.first().map(PrefixWithOptionalPorts::prefix))
+        {
             Some(Prefix::IPV6(_)) => "2001:db8:ffff::/64",
             _ => "3.3.3.0/24",
         };
@@ -1252,9 +1263,15 @@ pub mod contract {
         vpc_table.add(Vpc::new("VPC-1", "AAAAA", LOCAL_VNI)?)?;
         vpc_table.add(Vpc::new("VPC-2", "BBBBB", REMOTE_VNI)?)?;
 
-        let local = VpcManifest::new("VPC-1").exposing(expose);
-        let remote =
-            VpcManifest::new("VPC-2").exposing(VpcExpose::empty().ip(remote_prefix.into()));
+        let local = exposes
+            .into_iter()
+            .fold(VpcManifest::new("VPC-1"), VpcManifest::exposing);
+        let remote = VpcManifest::new("VPC-2").exposing(
+            VpcExpose::empty().ip(remote_prefix
+                .parse::<Prefix>()
+                .unwrap_or_else(|_| unreachable!())
+                .into()),
+        );
         let mut peerings = VpcPeeringTable::new();
         peerings.add(VpcPeering::with_default_group(
             "VPC-1--VPC-2",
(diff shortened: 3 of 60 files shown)


k8s-intf/src/bolero/spec.rs
```diff
@@ -44,6 +44,85 @@ fn increment_string(s: &mut str) {
...
 
+#[derive(Debug, Clone)]
+pub struct SpecBuilder {
+    max_vpcs: u8,
+    max_peerings: u8,
+    max_exposes: u8,
+    max_subnets: u8,
+    flavours: Vec<NatFlavour>,
+    families: Vec<AddressFamily>,
+}
```
daniel-noland: can't we just use `derive(Builder)` for this?

k8s-intf/src/bolero/support.rs
```diff
@@ -423,3 +423,109 @@ mod test {
...
+        let addr = (base | within) & mask;
+        format!("{}/{len}", Ipv6Addr::from(addr))
+    }
+
+    pub fn private<D: Driver>(d: &mut D, family: AddressFamily, len: u8) -> Option<String> {
+        Some(if family.is_v4() {
+            v4(0x0A00_0000, 8, d.produce::<u32>()?, len)
+        } else {
+            v6(
+                0x2001_0db8_0000_0000_0000_0000_0000_0000,
```
daniel-noland: should likely be a const

mgmt/src/tests/mgmt.rs
```diff
@@ -1,6 +1,41 @@
...
+    //
+    //   ambiguity  reordered   796 cases -> 8.9%   |  232 cases -> 3.9%
+    //   relevance  compared   1264 cases -> 3.0%   |  230 cases -> 0.4%
+    //
+    // The left column is a developer's machine, the right a CI runner at a
+    // third of the throughput. A floor calibrated on the left fires on the
+    // right for want of cases rather than for a defect, so the gate has to sit
+    // above what the right column reaches.
+    500
+};
```
daniel-noland: should use cfg_select!

k8s-intf/src/bolero/support.rs
```diff
@@ -292,6 +306,29 @@ pub fn generate_prefixes<D: Driver>(
 
 #[cfg(test)]
 mod test {
+    /// The mask floor an interface generator imposes is a property of how many
+    /// addresses it must keep apart, nothing else. A single address needs one bit,
+    /// so it must still be free to be a /1, and ten addresses need five.
+    ///
+    /// Pinning these down deterministically rather than by sampling: the defect this
+    /// guards against silently raised the floor to 17 for v4 and 113 for v6, which a
+    /// property that only checks uniqueness cannot see.
```
daniel-noland: questionable comment

k8s-intf/src/bolero/support.rs
```diff
@@ -129,6 +129,20 @@ impl ValueGenerator for UniqueV6CidrGenerator {
     }
 }
 
+/// How many high-order bits it takes to give each of `count` addresses a distinct
+/// prefix, plus one so the all-zero prefix is not the only choice.
+///
+/// This is a property of the counter, not of the address, so it is the same for v4
+/// and v6. Sizing it from the address width instead is what pinned every generated
+/// interface address into a short suffix of the space.
```
daniel-noland: remove comment

config/src/external/overlay/vpcpeering.rs
```diff
@@ -1005,3 +1005,131 @@ impl VpcPeeringTable {
...
+
+    fn v4_pair<D: Driver>(driver: &mut D, host_bits: u8) -> Option<(Prefix, Prefix)> {
+        let len = 32 - host_bits;
+        let mask = u32::MAX.checked_shl(u32::from(host_bits)).unwrap_or(0);
+        let internal = (0x0A00_0000 | (driver.produce::<u32>()? & 0x00FF_FFFF)) & mask;
+        let external = (0xAC10_0000 | (driver.produce::<u32>()? & 0x000F_FFFF)) & mask;
+        Some((prefix_v4(internal, len)?, prefix_v4(external, len)?))
+    }
+
+    const INTERNAL_BASE: u128 = 0x2001_0db8_0000_0000_0000_0000_0000_0000;
```
mvachhar, line 1076: This will cost use some coverage for bit errors in the upper bits, or validation bugs with multicast addresses, etc.  Fine for now, just pointing it out.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -461,6 +453,29 @@ impl VpcExpose {
                     ));
                 }
             }
+
+            let internal = collapsed_expose
```
mvachhar, line 457: I'm not certain this fix is right.  The prefix lengths don't ahve to match on a cidr by cidr basis, just the total addresses on each side of the expose.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -1287,6 +1302,59 @@ pub mod contract {
...
+                .make_port_forwarding(None, None)
+                .unwrap_or_else(|_| unreachable!())
+                .ip(side(internal))
+                .as_range(side(external))
+                .unwrap_or_else(|_| unreachable!())
+        }
+
+        #[test]
+        fn compensating_sizes_do_not_make_a_valid_expose() {
+            let expose = forwarding(("10.0.0.0/32", 1000, 1099), ("172.16.0.0/30", 2000, 2024));
```
mvachhar, line 1322: So this is a more interesting case, here the number of ports matches exactly on both sides.  Do we want to allow this or do the number of *ips* on each side have to match.  I cannot recall what we decided the correct semantics are.

mvachhar: @qmonnet is this compensating port counts in fact valid?  Does ti work as expected?

k8s-intf/src/bolero/vpc.rs
```diff
@@ -69,8 +71,8 @@ impl ValueGenerator for VpcGenerator<'_> {
         )?;
 
         let subnets_cidrs = vec![
-            blocks::private_run(d, AddressFamily::V4, v4_masklen, num_v4_cidrs)?,
-            blocks::private_run(d, AddressFamily::V6, v6_masklen, num_v6_cidrs)?,
+            blocks::private_run(d, AddressFamily::V4, self.vpc, v4_masklen, num_v4_cidrs)?,
```
mvachhar, line 74: The problem with this change here is that some overlaps are legal, in particular overlaps with masquerade are perfectly fine but this is now only covered by mutation and it isn't clear if this will test the validators logic to check for *legal* overlapped prefixes.

k8s-intf/src/bolero/crd.rs
```diff
@@ -24,22 +24,37 @@ fn simple_hostname<D: Driver>(d: &mut D) -> Option<String> {
...
 }
 
-///
+fn join_own_groups<D: Driver>(d: &mut D, name: &str, spec: &mut GatewayAgentSpec) -> Option<()> {
+    for group in spec.groups.iter_mut().flatten().map(|(_, group)| group) {
+        if !d.produce::<bool>()? {
+            continue;
+        }
+        if let Some(member) = group.members.iter_mut().flatten().next() {
+            member.name = name.to_string();
```
mvachhar, line 33: This will add a gateway to a group possibly multiple times.  Is this ok?

mgmt/src/tests/mgmt.rs
```diff
@@ -853,60 +853,70 @@ mod dataplane_tables {
...
-                    .unwrap_or_else(|e| panic!("a schema-legal CRD did not convert: {e}"));
-                let Ok(validated) = external.validate() else {
-                    return;
-                };
-                built.fetch_add(1, Ordering::Relaxed);
-                build_tables(&validated, flavour);
-            });
+    /// Drive one NAT flavour through validation and every dataplane table it implies.
+    ///
+    /// A macro rather than a function because `bolero::check!()` registers a fuzz target
```
mvachhar, line 858: Lame.

Cargo.lock
```diff
@@ -1551,6 +1551,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml_ng",
+ "strum",
```
mvachhar, line 1554: This should have been part of the commit that introduced strum, but I won't block the PR because of this.

k8s-intf/Cargo.toml
```diff
@@ -55,9 +55,12 @@ thiserror = { workspace = true }
 tokio = { workspace = true, optional = true }
 
 [dev-dependencies]
+# `pub mod bolero` is gated on `any(test, feature = "bolero")`, so everything the module
+# needs has to be reachable in a plain `cargo test` too, not only behind the feature.
 bolero = { workspace = true }
 lpm = { workspace = true, features = [] }
 net = { workspace = true, features = ["bolero", "test_buffer"] }
+strum = { workspace = true, features = ["derive"] }
```
mvachhar, line 63: This change should be lumped in with where strum is first used.  If you feel like it you can squash this commit back there.  Looks like reordering and rebasing mangled where this was added and where the Cargo.lock was changed.

mvachhar (dismissed):
I think the bolero generators here assume that no expose prefixes can overlap, but that is not true for masquerade and port forwarding.  I am surprised the fuzzer hasn't failed for that case yet.  But maybe I am missing where it allows for that possibility.

Also @qmonnet I am pretty sure that the prefix length checks the AI added here are just wrong, can you look at my comments and verify?  If so, we just ahve to drop a couple commits (I think).

config/src/external/overlay/vpcpeering.rs
```diff
@@ -1060,6 +1063,91 @@ pub mod contract {
...
+                Side::Private => 0x0A00_0000 | (u32::from(index) << 16),
+                Side::Public => 0xAC10_0000 | (u32::from(index) << 8),
+            };
+            prefix_v4(bits, 24)
+        } else {
+            let selector = match side {
+                Side::Private => 0u128,
+                Side::Public => 1,
+            };
+            let bits = (0x2001_0db8u128 << 96) | (selector << 80) | (u128::from(index) << 64);
```
qmonnet, line 1120: Is this to avoid prefixes in `SPECIAL_USE_PREFIXES`?

config/src/external/overlay/vpcpeering.rs
```diff
@@ -1122,6 +1122,85 @@ pub mod contract {
...
+                Side::Private => 0x0A00_0000u32,
+                Side::Public => 0xAC10_0000,
+            })
+        } else {
+            let selector = match side {
+                Side::Private => 0u128,
+                Side::Public => 1,
+            };
+            (0x2001_0db8u128 << 96) | (selector << 80)
+        };
```
qmonnet, line 1189: This should probably be refactored with the similar stuff in `block()` in the same file. At the moment the logic is unclear (IMO) _and_ duplicated.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -432,8 +432,6 @@ impl VpcExpose {
         // - we have no exclusion prefixes (note: we could relax this constraint now that we
         //   collapse exclusion prefixes early)
         // - we have a single prefix on each side (private and public addresses)
-        // - we have the same number of addresses on each side
-        // - the list of associated port ranges also has the same size on each side
```
qmonnet, line 436: Why remove these comments? Your fix specifically makes these true

config/src/external/overlay/vpcpeering.rs
```diff
@@ -446,12 +444,6 @@ impl VpcExpose {
                     "Port forwarding requires a single prefix on each side",
                 ));
             }
-            if ips_sizes != as_range_sizes {
-                return Err(ConfigError::MismatchedPrefixSizes(
-                    ips_sizes,
-                    as_range_sizes,
-                ));
-            }
```
qmonnet, line 454: Discussed with Manish - this check may technically be enough, I think we should be able to accept identical values of IPs x ports on both sides even if the individual numbers differ.

However, it's also true we're currently rejecting such setup later, it's here: https://github.com/githedgehog/dataplane/blob/f12d6a26244e666383d9445e10ed7268a4c9ce45/nat/src/portfw/portfwtable/objects.rs#L138-L142

So it's OK to move the check to validation time, we should create an issue to remember to check whether it can work without identical IP numbers, and to revert this change if we're OK accepting a less strict configuration.

config/src/external/overlay/vpcpeering.rs
```diff
@@ -1220,7 +1220,18 @@ pub mod contract {
...
+    }
+
+    pub fn overlay_with_exposes(exposes: Vec<VpcExpose>) -> Result<Overlay, ConfigError> {
+        let remote_prefix = match exposes
+            .first()
+            .and_then(|expose| expose.ips.first().map(PrefixWithOptionalPorts::prefix))
+        {
             Some(Prefix::IPV6(_)) => "2001:db8:ffff::/64",
             _ => "3.3.3.0/24",
         };
```
qmonnet, line 1237: Note: With the hardcoded prefixes, I'd love to have this under a `#[config(test)]` gate. It's probably not possible to do just that because we use it in different crates, but maybe time to add a `testing` feature to the `config` crate?

mgmt/src/processor/confbuild/internal.rs
```diff
@@ -159,6 +165,8 @@ impl VpcRoutingConfigIpv4 {
         /* list of advertised prefixes */
         self.adv_nets.extend(nets.clone());
 
+        reject_ipv6(nets.iter().copied())?;
```
qmonnet, line 168: Could you please create an issue to track this?

k8s-intf/src/bolero/mutate.rs
```diff
@@ -114,7 +114,6 @@ fn is_static(expose: &GatewayAgentPeeringsPeeringExpose) -> bool {
         .is_some_and(|nat| nat.r#static.is_some())
 }
 
-#[allow(clippy::too_many_lines)]
```
qmonnet, line 117: Nit: Commit description doesn't mention this, but says `Also restore a displaced documentation attribute to its intended function`, although no doc attribute seems to be updated.

(`fix(k8s-intf): reject overflowing prefix slot budgets`)

mgmt/src/tests/mgmt.rs
```diff
@@ -661,9 +684,25 @@ mod peering_chain {
...
-                let internal = build_internal_config(&validated, None).unwrap_or_else(|e| {
-                    panic!("a validated {flavours:?} configuration would not build: {e}\n{exposes:#?}")
-                });
+                let internal = match build_internal_config(&validated, None) {
+                    Ok(internal) => internal,
+                    // The FRR configuration built from a peering is IPv4-only, so a
+                    // validated IPv6 peering cannot be rendered. That gap is real and the
+                    // user meets it at apply time, not at submit time. Check it here
+                    // rather than skip past it: only an IPv6 configuration may reach this,
+                    // and if one ever builds, this arm is what says the limitation lifted.
```
qmonnet, line 693: Nit: _That gap is real_. AI verbosity, too! 🙂

k8s-intf/src/bolero/reduce.rs
```diff
@@ -13,7 +14,14 @@ pub struct Dropped {
...
     pub vpc: String,
     pub index: usize,
-    pub nat: Option<&'static str>,
+    /// Which kind of translation the dropped expose asked for, if any.
+    ///
+    /// A `NatFlavour` rather than a string. `development/code/error-handling.md` calls
+    /// arbitrary string values actively hostile and matching on their contents extremely
+    /// fragile, and the consumer of this field does exactly that kind of match: a typo in
+    /// either the producer or the consumer compiled cleanly and surfaced as an
+    /// `unreachable!` during a fuzz run rather than as a build error.
```
qmonnet, line 23: Nit: Leftover from dev work, more confusing than anything when looking at the code

mgmt/src/tests/mgmt.rs
```diff
@@ -1037,6 +1051,9 @@ mod validator_completeness {
         let ruleset = build_port_forwarding_configuration(vpc_table).unwrap_or_else(|e| {
             panic!("{mutation:?}: validator accepted a config port forwarding rejects: {e}")
         });
+        // Cannot fire yet: `validate_ruleset` is a stub returning `Ok`, so `update_table`
+        // is infallible. The (04) PR in this stack gives it a real `PortFwTable::dry_run`,
```
qmonnet: The (04) PR in this stack should also not be mentioned in the code.

k8s-intf/Cargo.toml
```diff
@@ -55,9 +55,12 @@ thiserror = { workspace = true }
 tokio = { workspace = true, optional = true }
 
 [dev-dependencies]
+# `pub mod bolero` is gated on `any(test, feature = "bolero")`, so everything the module
+# needs has to be reachable in a plain `cargo test` too, not only behind the feature.
 bolero = { workspace = true }
```
qmonnet, line 60: I don't understand this comment?

k8s-intf/src/bolero/expose.rs
```diff
@@ -107,10 +107,21 @@ impl<'a> ExposeGenerator<'a> {
             return None;
         }
         let longer = d.gen_u8(Bound::Excluded(&len), Bound::Included(&max))?;
-        if private {
-            blocks::private(d, self.family, at, longer)
+        // Carve the exclusion out of the parent. Redrawing it anywhere in the parent's
+        // block, which is what this did, put it inside the parent about one time in a
```
qmonnet, line 111: “which is what this did”

mgmt/src/tests/mgmt.rs
```diff
@@ -508,3 +543,1080 @@ pub mod test {
...
+        NatTablesWriter::new().update_nat_tables(nat_tables);
+
+        let masquerade = MasqueradeConfig::new(vpc_table).set_randomize(false);
+        NatAllocatorWriter::new().update_nat_allocator(masquerade, genid, &FlowTable::new(16));
+
+        let ruleset = build_port_forwarding_configuration(vpc_table).unwrap_or_else(|e| {
+            panic!("{mutation:?}: validator accepted a config port forwarding rejects: {e}")
+        });
+        // Cannot fire yet: `validate_ruleset` is a stub returning `Ok`, so `update_table`
+        // is infallible. The (04) PR in this stack gives it a real `PortFwTable::dry_run`,
```
qmonnet, line 1160: The (04) PR in this stack should also not be mentioned in the code.

config/src/external/mod.rs
```diff
@@ -93,6 +93,67 @@ impl ExternalConfig {
...
+    /// from a `ValidatedOverlay` and handles IPv6 perfectly well, and twelve of its tests
+    /// say so.
+    ///
+    /// `Peering::validate` has already established that a peering's two manifests agree
+    /// on their IP version, so `is_v4` speaks for the whole peering.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`ConfigError::Unsupported`] if a peering this gateway renders carries
+    /// IPv6 prefixes.
```
qmonnet, line 126: Not sure we need this amount of details. “... And twelve of its tests say so.”

mgmt/src/processor/confbuild/internal.rs
```diff
@@ -69,10 +68,22 @@ fn vpc_import_prefix_list_for_peer(
...
+/// This used to be where an IPv6 peering was refused, which made it a refusal at apply
+/// time: the configuration was accepted, reported good, and then rejected as a whole
+/// when the routing config was built, taking its IPv4 vpcs with it. Anything accepted
+/// at validation has to apply cleanly, so `check_peerings_are_ipv4` now turns such a peering
+/// away at submit time and nothing reaching here can carry IPv6.
+///
+/// It is kept because the alternative to refusing by name is what this replaced: the
+/// advertise list failed `is_version_compatible` and came back as an `InternalFailure`,
+/// and the import list dropped v6 prefixes in silence. Neither is a thing to fall back
+/// to if the validation ever loosens.
```
qmonnet, line 82: Replace with `TODO: Add IPv6 support` + URL to GitHub issue?

config/src/external/mod.rs
```diff
@@ -93,6 +93,67 @@ impl ExternalConfig {
...
+                .is_some_and(|rank| communities.get_community(rank).is_some())
+        };
+        if overlay
+            .vpc_table()
+            .peerings()
+            .filter(|peering| renders(peering))
+            .all(ValidatedPeering::is_v4)
+        {
+            return Ok(());
+        }
```
qmonnet, line 150: It would be great if this snippet could instead call `reject_ipv6()` from `mgmt` (or more likely, have `reject_ipv6()` somehow adjusted and moved here, since `mgmt` depends on `config` and not the other way around). This way, when we remove the restriction in one place, we'd be sure not to miss the other one.

qmonnet (commented):
Thank you for this work!

Looks good overall. That's quite a massive amount of additional fuzzing checks. I note that many tests are here to consolidate, rather than just extend, the fuzzers, and that's great. That's also somewhat concerning, in the sense that it requires a lot of work to keep the fuzzing relevant (and ensure generated inputs cover all desired cases), I'm somewhat wary of the surface to maintain - it's easy to detect when a test breaks and fix it, but it also seems quite possible that some of the tests may no longer reflect the right invariants to check, and still pass, and I fear this will be hard to detect.

Anyway. Great to have all this coverage. I've got a few comments below: some leftover verbose comments (I only raised the worst I found), and some notes regarding IPv6 in config, mostly.

config/src/external/overlay/vpcpeering.rs
qmonnet: I think it should work, but Fredi did the code for the port forwarding so we need to check with him.

See also https://github.com/githedgehog/dataplane/pull/1796#discussion_r3980935318

config/src/external/overlay/vpcpeering.rs
mvachhar: Good call, I missed that there was no test gate or bolero/fuzz gate here.

config/src/external/overlay/vpcpeering.rs
mvachhar: @Fredi-raspall can you take a look at this test and tell me if you think the mapping here should work?  @qmonnet thinks it should work, and if so, the test is wrong.

config/src/external/overlay/vpcpeering.rs
mvachhar: @qmonnet what about the prefix length stuff here?  Do they have to match cidr by cidr or is just the total address count matching ok?

k8s-intf/src/bolero/crd.rs
mvachhar: This is resolved in a later commit in the series.

config/src/external/overlay/vpcpeering.rs
mvachhar: I'll take a look at the purpose of this and possibly consolidating it.

config/src/external/overlay/vpcpeering.rs
mvachhar: Looks like we reject this lower down in the stack so we should probably just reject it here.

k8s-intf/src/bolero/vpc.rs
mvachhar: I'm going to leave this as is for now as it is possible that the fuzzer has handled this case.  If this results in a fuzzing failure, we can fix it then.

mgmt/src/processor/confbuild/internal.rs
```diff
@@ -392,3 +420,314 @@ pub fn build_internal_config(
...
+    // supposed to make the property skip; a bare count makes it lie. Same
+    // reasoning, and the same shape, as `mgmt::tests::mgmt::ENOUGH_CASES`.
+    //
+    // Half the range rather than `usize::MAX`, which clippy reads as an absurd
+    // comparison and rejects at the call sites.
+    const ENOUGH_CASES: usize = if cfg!(instrumented) || cfg!(emulated) {
+        usize::MAX / 2
+    } else {
+        200
+    };
```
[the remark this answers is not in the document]
mvachhar: I don't think this is right.  We have enough cases in CI with 230 and we want there to be enough cases.  The case count is test by test.

k8s-intf/src/bolero/peering.rs
```diff
@@ -55,47 +90,129 @@ impl<'a> LegalValuePeeringsGenerator<'a> {
...
+                    self.flavours
+                } else {
+                    &stateless
+                };
+                let generator = LegalValuePeeringsPeeringGenerator::new(
+                    self.vpc_subnets.get(vpc_names[i]).unwrap_or(&empty_map),
+                    flavours,
+                    family,
+                    self.max_exposes,
+                    u8::try_from(vpcs[i]).unwrap_or(u8::MAX),
```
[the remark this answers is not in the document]
mvachhar: I don;'t know why the AI thinks that the subnets len is the issue, the problem would arise if max exposes is > 16 but in practice we pass 3, so I think this is fine.

k8s-intf/src/bolero/spec.rs
```diff
@@ -44,6 +45,85 @@ fn increment_string(s: &mut str) {
...
+            families: AddressFamily::all(),
+        }
+    }
+}
+
+impl SpecBuilder {
+    #[must_use]
+    pub fn max_vpcs(mut self, max: u8) -> Self {
+        self.max_vpcs = max;
+        self
```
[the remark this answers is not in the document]
mvachhar: These types of cross size checks are too difficult to validate right now and this is in the bolero code, where we don't actually use vpcs > 16 so I am going to leave this for now.  If it annoys someone later they can fix it.

k8s-intf/src/bolero/support.rs
```diff
@@ -416,3 +254,274 @@ mod test {
...
+        pub fn nth(slot: u8, sub: u8, subs: u8) -> Self {
+            Self {
+                slot,
+                sub,
+                subs: subs.max(sub.saturating_add(1)),
+            }
+        }
+
+        fn sub_bits(self) -> u8 {
+            u8::try_from(self.subs.max(1).next_power_of_two().trailing_zeros()).unwrap_or(0)
```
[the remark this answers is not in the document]
mvachhar: Fixed.

k8s-intf/src/bolero/mutate.rs
```diff
@@ -0,0 +1,546 @@
...
+        | Mutation::MixAddressFamilies => mutate_expose_shape(agent, mutation),
+        Mutation::UseReservedPrefix => {
+            let reserved = ["127.0.0.0/8", "224.0.0.0/4", "0.0.0.0/8", "ff00::/8"];
+            let choice = crate::bolero::support::choose(d, &reserved)?;
+            let mut done = false;
+            for expose in exposes_mut(agent) {
+                if let Some(entry) = expose.ips.iter_mut().flatten().next()
+                    && entry.cidr.is_some()
+                {
+                    entry.cidr = Some(choice.to_string());
```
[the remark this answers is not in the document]
mvachhar: Fixed.

k8s-intf/src/bolero/peering.rs
```diff
@@ -16,24 +18,53 @@ use crate::gateway_agent_crd::{GatewayAgentPeerings, GatewayAgentPeeringsPeering
...
 impl ValueGenerator for LegalValuePeeringsPeeringGenerator<'_> {
     type Output = GatewayAgentPeeringsPeering;
 
     fn generate<D: Driver>(&self, d: &mut D) -> Option<Self::Output> {
-        let num_expose = d.gen_usize(Bound::Included(&1), Bound::Included(&16))?;
-        let expose_gen = LegalValueExposeGenerator::new(self.subnets);
-        let expose = (0..num_expose)
-            .map(|_| expose_gen.generate(d))
-            .collect::<Option<Vec<_>>>()?;
+        let num_expose = d.gen_u8(Bound::Included(&1), Bound::Included(&self.max_exposes))?;
```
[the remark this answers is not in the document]
mvachhar: Not going to deal with this and it is test code, so it can be fixed later when it bites us.

k8s-intf/src/bolero/spec.rs
```diff
@@ -52,61 +132,117 @@ fn increment_string(s: &mut str) {
...
         for i in 0..num_vpcs {
-            let vni_offset = u32::try_from(i).expect("too many vpcs");
-            let lv_vpc = d.produce::<LegalValue<GatewayAgentVpcs>>()?;
-            let mut vpc = lv_vpc.take();
+            let mut vpc = crate::bolero::vpc::VpcGenerator::new(
+                u8::try_from(i).unwrap_or(u8::MAX),
+                knobs.max_subnets,
+                &knobs.families,
+            )
+            .generate(d)?;
```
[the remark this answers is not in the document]
mvachhar: Again, this is test code and requires deep analaysis to fix all these.  We'll deal with it as we go.

mgmt/src/tests/mgmt.rs
```diff
@@ -508,3 +543,1080 @@ pub mod test {
...
+    }
+
+    #[test]
+    fn a_configuration_with_no_nat_builds_its_tables() {
+        drive!(NatFlavour::None);
+    }
+}
+
+#[cfg(test)]
+#[cfg(test)]
```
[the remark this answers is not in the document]
mvachhar: Fixed.

config/src/external/overlay/vpcpeering.rs
mvachhar: No, this is the Ipv6 doc prefix, I'll give it ia constant.

config/src/external/overlay/vpcpeering.rs
mvachhar: This is the documentation prefix, I'll give it a constant.

config/src/external/overlay/vpcpeering.rs
qmonnet: > what about the prefix length stuff here? Do they have to match cidr by cidr or is just the total address count matching ok?

- CIDR by CIDR: Not applicable, because we allow only one CIDR in the expose for port forwarding.
- Address by address, port by port vs (addresses x ports) by (addresses x ports): I believe both approaches should work, although at the moment we only accept the former (which this commit makes more explicit, at config time).

For comparing (addresses x ports): we need to check it works in practice, and also to decide whether we want to expose this option to the user, because which (address, port) maps to which other tuple is not always obvious.

> Looks like we reject this lower down in the stack so we should probably just reject it here.

Yes

config/src/external/overlay/vpcpeering.rs
qmonnet: This is related to the check in https://github.com/githedgehog/dataplane/pull/1796#discussion_r3973445784, given that we want to match address by address and port by port, having a different combination of (addresses number / ports number) should not validate for port forwarding. As far as I can tell, this test agrees with the rejection we've clarified in the config validation.

config/src/external/overlay/vpcpeering.rs
mvachhar: Fixed.  We had this in a lot of places.

config/src/external/overlay/vpcpeering.rs
mvachhar: Fixed.

config/src/external/overlay/vpcpeering.rs
mvachhar: Fixed.

config/src/external/overlay/vpcpeering.rs
mvachhar: These are gated by `test` or `bolero` so I think it is fine but the gate is on the whole submodule.

qmonnet: Oh I missed it :facepalm:. All good, then

config/src/external/mod.rs
mvachhar: reject_ipv6 is there specifically to check for ipv6 when building routing.  The other check for the peering is more general.  I think I want to leave this as is for now, the code is very small and we'd need extensive fixes to remove it and when we do we'll add tests that catch anywhere we have this kind of check.

qmonnet: Look also at what Fredi merged in https://github.com/githedgehog/dataplane/pull/1812/changes/b832afa9262c5f65eec5ac61c7f6e07c5256380d, it doesn't seem to conflict, but might be worth taking a look. Fredi also mentioned he was adding support for IPv6 here in another PR, I think?

config/src/external/overlay/vpcpeering.rs
mvachhar: So no change needed.  Resolving.

config/src/external/overlay/vpcpeering.rs
mvachhar: See #1814

config/src/external/mod.rs
mvachhar: I am going to resolve this, unresolve if you disagree.

mgmt/src/processor/confbuild/internal.rs
mvachhar: I added a TODO but kept the comment, I think it is useful.  See #1815

config/src/external/mod.rs
mvachhar: Fixed.

mgmt/src/tests/mgmt.rs
mvachhar: Comment omitted

k8s-intf/src/bolero/expose.rs
mvachhar: Fixed.

mvachhar:
@qmonnet I've addressed most of your and code rabbits comments.  Some of the fixes (mostly the comment related ones) were fixups I applied to existing commits.  For the other substantive fixes I stacked a few more commits.  Please rereview and approve ASAP.  @daniel-noland alternatively, please look at the recently resolved comments and confirm that I addressed them adequately and approve.  Either route is fine.

mvachhar (commented):
I think I addressed all the comments.  Somone besides me should approve (or not).

qmonnet (approved):

PR #1797, opened 2026-09-03
(02) fuzz

daniel-noland:


k8s-intf/src/bolero/crd.rs
```diff
@@ -99,7 +99,6 @@ impl GatewayAgentBuilder {
 }
 
 /// Generate a random legal `GatewayAgent` value
-///
```
daniel-noland, line 102: this commit can now be dropped

nat/src/masquerade/fuzz.rs
```diff
@@ -96,21 +109,64 @@ fn destination_of(packet: &Packet<TestBuffer>) -> (IpAddr, u16) {
...
+/// anything; a half would still sit only 1.1x under the worst yield seen there, where a
+/// quarter sits 2.2x under it. The five properties that reach every probe clear a
+/// quarter by 32x. That asymmetry is the point -- this is a smoke alarm for a property
+/// that has stopped checking, not a thermometer for one that checks slightly less than
+/// it used to.
+///
+/// Confirmed against the sample [`TEST_TIME`] buys CI: 200 runs at thirty-one
+/// configurations, none failing, and the worst of them reached eighteen flows against a
+/// threshold of eight. A rebuild during a sweep like that invents failures, because the
+/// binary is replaced under the loop -- run it against a copy.
```
daniel-noland: remove all prose

nat/src/masquerade/fuzz.rs
```diff
@@ -18,6 +18,19 @@ const MAX_EXPOSES: u8 = 3;
...
+///
+/// bolero's own default is one second, which buys CI about seven configurations -- under
+/// [`ENOUGH_TO_RATE`], so the yield rate would stand down on every run and only the
+/// nonzero check would ever fire. At the 6.23 configurations a second measured on a CI
+/// runner 2026-09-09 this budget draws about thirty, which clears the floor with room for
+/// a runner slower than that one.
+///
+/// `with_test_time` is `.or()` against the environment, so `BOLERO_RANDOM_TEST_TIME_MS`
+/// still wins: miri keeps its 30s and the coverage recipe its 15s. A runner too slow to
+/// reach the floor loses the rate check and prints that it did -- it does not fail.
```
daniel-noland: remove all prose

nat/src/masquerade/fuzz.rs
```diff
@@ -96,21 +109,64 @@ fn destination_of(packet: &Packet<TestBuffer>) -> (IpAddr, u16) {
 /// 1060 to 2863 for the static-NAT twin. The difference is the cost of a case
 /// here -- a flow table and the timers that go with it -- and not, as an earlier
 /// version of this comment guessed, anything about how the runtime is driven: the
-/// counts are the same either side of the commit that changed that.
+/// counts are the same either side of the commit that changed that. Those are
+/// one-second numbers; [`TEST_TIME`] has since raised the budget, and the same
+/// machine now reaches 73 to 109.
 ///
```
daniel-noland: remove this edit

nat/src/masquerade/fuzz.rs
```diff
@@ -96,21 +109,64 @@ fn destination_of(packet: &Packet<TestBuffer>) -> (IpAddr, u16) {
...
 /// The floor is therefore *two*, not a fraction of the measured run. A CI runner
 /// at a third of this machine's throughput would sit under any floor calibrated
 /// on the numbers above, and a guard that stands down everywhere it runs is worse
 /// than no guard. Two is the smallest sample a rate can be computed from at all,
-/// which is exactly the corpus-replay case and nothing else. The ratio is the
-/// part that carries meaning -- see the commit that removed the previous absolute
-/// floor for why a bigger number measures the machine rather than the property.
+/// which is exactly the corpus-replay case and nothing else.
```
daniel-noland: remove this change

nat/src/static_nat/fuzz.rs
```diff
@@ -108,11 +108,26 @@ fn five_tuple_destination(packet: &Packet<TestBuffer>) -> (IpAddr, u16) {
 /// rather than picked.
 ///
 /// Measured natively 2026-09-06: 1060 to 2863 configurations per property here,
-/// where the masquerade twin reaches 12 to 28. The floor matches the twin's
-/// rather than this crate's own headroom, so the two do not drift apart, and it
-/// is deliberately the smallest sample a rate can be computed from at all rather
-/// than a fraction of either measurement -- a floor scaled to a fast machine
-/// stands down on a slow one, which is the failure mode of a guard.
+/// where the masquerade twin reached 12 to 28 on the one-second default it ran
```
daniel-noland: drop this commit

nat/src/static_nat/probe.rs
qmonnet, line 1: I'd like if the file name could reflect that this is only used for tests

acl-filter/src/nf_fuzz.rs
```diff
@@ -182,7 +182,8 @@ fn the_stage_agrees_with_the_configuration() {
                     continue;
                 };
 
-                let want = oracle_resolved_action(&built.overlay, &summary);
+                let judged = expected_summary(&summary);
```
qmonnet, line 185: What does `judged` mean? Wouldn't `expected` or `reference` be more explicit?

qmonnet (approved):
Looks good to me, even though I haven't read all the details of all fuzzing tests.

I note that there's no code change other than in tests.

Two "formatting" suggestions, for future work:
- I'd like if file names could be more explicit when they relate to tests only. `fuzz.rs` is explicit enough, but `probe.rs` is not so much IMO; maybe `fuzz_probe.rs`? Not blocking though, we can rename as a follow-up if necessary.
- I think I would avoid prefixing commit titles with `fix` when they only relate to tests, because I find it best to keep it for bug fixes so we can surface them more easily from the git history (although fixing a broken CI might arguably justify a `fix`, I don't know). Maybe a matter of preference, though

Thanks!

qmonnet:
Sorry I forgot that adding a label would re-trigger all the CI :facepalm: 

We could explore alternatives at some point, Cilium does it by running a GitHub app that dispatches workflows only when relevant, based on certain triggers (from `/test`-like comments, but could be adjusted to labels I suppose)

PR #1799, opened 2026-09-03
(04) spec

daniel-noland:


routing/src/cli/display.rs
```diff
@@ -46,7 +46,6 @@ use std::time::Duration;
 
 use tracing::{error, warn};
 
-//================================= Common ==========================//
```
qmonnet, line 49: Are these comments going away because the comment style conflicts with duvet syntax?

daniel-noland: yes

mgmt/src/tests/mgmt.rs
```diff
@@ -1152,9 +1152,9 @@ mod validator_completeness {
         let ruleset = build_port_forwarding_configuration(vpc_table).unwrap_or_else(|e| {
             panic!("{mutation:?}: validator accepted a config port forwarding rejects: {e}")
         });
-        // Cannot fire yet: `validate_ruleset` is a stub returning `Ok`, so `update_table`
-        // is infallible. The (04) PR in this stack gives it a real `PortFwTable::dry_run`,
-        // at which point this starts checking something. Left in place for that.
+        // Live as of this chapter: `validate_ruleset` is backed by a real
+        // `PortFwTable::dry_run`, so `update_table` can now reject a ruleset the validator
+        // accepted, and this assertion is the thing that catches the disagreement.
```
daniel-noland, line 1157: drop this

mgmt/src/processor/proc.rs
```diff
@@ -543,6 +543,22 @@ fn apply_masquerade_config(
...
+/// touching any table.
+///
+/// `apply_config` swaps the flow-filter, ACL, static-NAT and masquerade tables before it reaches
+/// port forwarding. Those swaps are not undone on a later error and the generation id is only
+/// advanced at the very end, so a ruleset rejected at that point leaves four tables live under a
+/// generation that is never published. Running the same validation up front turns that into a
+/// clean refusal before anything has been committed.
+///
+/// This narrows the window rather than closing the class: `apply_router_config` still runs after
+/// every swap and can still fail late. Closing that properly needs a real two-phase apply.
```
daniel-noland, line 556: drop prose

nat/src/portfw/portfwtable/access.rs
```diff
@@ -29,7 +29,14 @@ impl Absorb<PortFwTableChange> for PortFwTable {
...
 pub struct PortFwTableReader(ReadHandle<PortFwTable>);
 
-fn validate_ruleset(ruleset: &[PortFwEntry]) -> Result<(), PortFwTableError> {
+/// Answer whether `update_table` would accept this ruleset, without touching the table.
+///
+/// Exposed so that a caller applying several tables can find out *before* it commits any of
+/// them. `update_table` validates first and so never half-updates the port-forwarding table
+/// itself, but it is applied after the flow-filter, ACL, static-NAT and masquerade tables have
+/// already been swapped, and a failure there leaves those four live under a generation that is
+/// never published.
```
daniel-noland, line 38: drop prose

routing/src/cli/display.rs
qmonnet: That's a long file, please let's reformat the comments rather than removing them all.

net/src/headers/embedded.rs
```diff
@@ -241,16 +243,22 @@ impl EmbeddedHeaders {
...
+                    if icmp_length > buf.len() {
+                        return;
+                    }
+                    //= https://www.rfc-editor.org/rfc/rfc4884#section-3
+                    //# When the ICMP Extension Structure is appended to an ICMP message
+                    //# and that ICMP message contains an "original datagram" field, the
+                    //# "original datagram" field MUST contain at least 128 octets.
+                    if icmp_length < MIN_ORIGINAL_DATAGRAM_OCTETS {
                         return;
                     }
```
qmonnet: I don't think this check is correct? We haven't checked whether we have an ICMP Extension Structure that was appended to the ICMP message, which I read as one condition for enforcing the 128 octets minimum, so the requirement does not necessarily apply?

net/src/headers/embedded.rs
```diff
@@ -241,16 +243,22 @@ impl EmbeddedHeaders {
...
-                    // ICMPv4: Padding is on 32-bit boundaries
-                    if padding_length < 32
-                        && buf[full_packet_length..icmp_length].iter().all(|b| *b == 0)
-                    {
+                    //= https://www.rfc-editor.org/rfc/rfc4884#section-3
+                    //# When the ICMP Extension Structure is appended to an ICMPv4 message
+                    //# and that ICMPv4 message contains an "original datagram" field, the
+                    //# "original datagram" field MUST be zero padded to the nearest
+                    //# 32-bit boundary.
+                    if buf[full_packet_length..icmp_length].iter().all(|b| *b == 0) {
```
qmonnet: Similarly, there's a notion of Extension Structure here, that is not validated.

(and same for the `type=test`s, later in the file)

qmonnet (commented):
Didn't have time to go very far with the review, but I think the comments below are worth investigating, so I'm posting them now, and will resume my review later.

development/code/spec-compliance.md
```diff
@@ -0,0 +1,200 @@
+# Specification compliance with duvet
```
daniel-noland, line 1: oh, missed this.  This file should be dropped

scripts/spec-interlock.ts
```diff
@@ -620,7 +620,9 @@ function parseArgs(argv: string[]) {
     only: [] as string[],
     jobs: 4,
     output: join(REPO, "target", "spec-interlock"),
-    json: "/tmp/duvet-interlock.json",
+    // Under the repo rather than /tmp, for the same reason as `output` above: the lab
+    // runner shares /tmp between concurrent jobs.
```
daniel-noland: remove prose

justfile
```diff
@@ -768,10 +769,33 @@ duvet-check:
         exit 1
       fi
     done
+    # `duvet report` rewrites the snapshot *and* every requirement TOML in place, so the
+    # committed copies are set aside and put back either way. A check that leaves the
+    # working tree dirty is a trap anywhere; in a repo with worktrees and a shared stash
+    # stack it is a trap that costs someone else's work.
```
daniel-noland: remove prose

justfile
```diff
@@ -768,10 +769,33 @@ duvet-check:
...
+    committed="$(mktemp -d)"
+    # Two traps, in this order, and the order is the whole point. The restore trap must not
+    # be armed until the thing it restores *from* exists: armed first, a `cp` that fails --
+    # `.duvet/requirements` absent, a full disk -- fires a handler that deletes the working
+    # copies and then cannot put anything back, so the check destroys the committed state it
+    # was meant to protect. Until the backup is made, the only safe handler is one that
+    # removes the temporary directory.
```
daniel-noland: remove prose

justfile
```diff
@@ -768,10 +769,33 @@ duvet-check:
...
+    trap 'rm -rf "${committed}"' EXIT
+    # `-a` so mtimes survive the round trip. `.duvet/requirements/**` is `include_str!`d by
+    # nat's RFC contract, so restoring it with a fresh timestamp makes cargo rebuild that
+    # crate and everything downstream on every `just lint`.
```
daniel-noland: remove prose

.gitattributes
```diff
@@ -1,3 +1,7 @@
 k8s-intf/src/generated/** linguist-generated
 k8s-intf/src/generated/mod.rs -linguist-generated
 npins/* linguist-generated
+# `duvet report` writes these: the extracted requirements carry the RFC text verbatim as
+# comments, which is thousands of lines of prose that no one edits by hand.
```
daniel-noland: remove prose

net/src/headers/embedded.rs
```diff
@@ -1298,12 +1334,130 @@ mod tests {
...
+    ///
+    /// A conformant `ICMPv4` sender pads a 120-octet datagram to exactly 128 -- the nearest
+    /// 32-bit boundary, then the 128-octet floor. Nothing here checks that, deliberately:
+    /// this exercise is the receiver's side of the leniency argued at `check_full_payload`,
+    /// where a longer field still carries the whole original datagram and the surplus is
+    /// still padding. It is named for what it asserts rather than for the requirement,
+    /// because the alignment requirement holds by construction and no test can see it: the
+    /// length attribute is in 32-bit words, so `check_full_payload` cannot be handed an
+    /// unaligned length by the parser, and handing it one directly proves nothing about the
+    /// parser.
```
daniel-noland: remove prose

net/src/headers/embedded.rs
```diff
@@ -235,49 +246,52 @@ impl EmbeddedHeaders {
...
+            // `!icmp_length.is_multiple_of(32)` was worse than that, comparing an octet
+            // count against a bit count.
+            //
+            // What is checked is that the surplus over the original datagram is zero.
+            // A conformant sender pads to exactly `max(128, round_up(len, alignment))`,
+            // and a longer field is out of spec -- but a receiver has no reason to
+            // discard a determination it can make: the original datagram really is all
+            // present, and the surplus really is padding, because the extension structure
+            // starts after the declared field. Being generous about the length and strict
+            // about the content is the direction that loses nothing.
```
daniel-noland: remove prose

net/src/headers/embedded.rs
```diff
@@ -48,6 +48,17 @@ pub struct EmbeddedHeaders {
...
 
+/// RFC 4884 section 4's backwards-compatibility floor for the "original datagram" field.
+///
+/// The errata were read before this shipped, because the vendored corpus carries only the
+/// base RFC and the project had recorded the constant as written without them. RFC 4884 has
+/// exactly one erratum -- ID 3, Verified, Technical, against section 7 -- and it replaces
+/// "the data structure" with "the ICMP Extension Structure" in the description of the
+/// extension checksum. Nothing touches section 3, section 4, this floor, or the padding and
+/// length-attribute rules cited in `check_full_payload`. Checked 2026-09-06 against
+/// <https://errata.rfc-editor.org/rfc4884>; re-check if a second erratum appears.
```
daniel-noland: remove prose

nat/src/masquerade/protocol.rs
```diff
@@ -50,17 +57,32 @@ fn next_flow_status_udp(action: NatAction, status: NatFlowStatus) -> NatFlowStat
     }
 }
 
+// No arm of this machine reaches a terminal state, so an ICMP *query* -- an Echo or its
+// reply -- never ends the flow it belongs to. That is half of RFC 5382 REQ-10 and RFC 4787
+// REQ-12, and it is not the half a reader assumes: an ICMP *error* does not come through
+// here at all. It goes to `IcmpErrorHandler`, which does tear a one-way mapping down on a
+// hard error, so neither requirement is met and neither is recorded here.
```
daniel-noland: remove prose

nat/src/masquerade/expiry.rs
```diff
@@ -93,6 +114,122 @@ fn reply_to(
...
+/// as covered by nothing at all. And the walk was host-major: every one of a host's 256
+/// sessions was drawn before the next host started, so the spill onto the second public
+/// address always fell *between* hosts and the probe passed without ever posing the
+/// question. Interleaving the hosts, which is what concurrent traffic looks like, splits
+/// all 254 of them.
+///
+/// So it asserts the departure rather than the compliance, and fails when the departure is
+/// repaired -- at which point this test and the two `type=exception` records it is named by
+/// come down together. See `apalloc::alloc::allocate` for REQ-2, the requirement that would
+/// have to be implemented for a host to keep one public address.
```
daniel-noland: remove prose

nat/src/masquerade/expiry.rs
```diff
@@ -22,6 +22,27 @@ const PAST_EXPIRY: Duration = Duration::from_secs(ONEWAY * 6);
 const WITHIN_LIFETIME: Duration =
     Duration::from_secs(crate::Masquerade::MASQUERADE_CLOSING_TIMEOUT.as_secs() / 2);
 
+const NEARLY_ESTABLISHED: Duration = Duration::from_secs(100);
+
+/// Used only by the established-mapping test, which reads its stretches the other
+/// way round from everything above: a period the mapping must *survive* because
+/// outbound traffic keeps refreshing it, and then a silence that must finish it.
+/// Deliberately literal rather than derived -- that test exercises the refresh
+/// path, not a deadline, and scaling these turns "must survive" into "must expire".
```
daniel-noland: remove prose

nat/src/masquerade/expiry.rs
```diff
@@ -22,6 +22,27 @@ const PAST_EXPIRY: Duration = Duration::from_secs(ONEWAY * 6);
...
+
+/// Long enough to outlast the Established deadline, whatever that deadline becomes.
+///
+/// `DEFAULT_MASQUERADE_IDLE_TIMEOUT` is two minutes and is deliberately *not* scaled by
+/// `TIMEOUT_SCALE`, so a literal here works today. It would stop working the moment REQ-5c
+/// is taken: the snapshot this branch ships records "a default value of five minutes or
+/// more ... is RECOMMENDED" as a todo, and at five minutes a literal five-minute silence
+/// lands on the deadline rather than past it and the last assertion of
+/// `outbound_traffic_keeps_an_established_mapping_alive` fails. Doubling the timeout says
+/// what the test means and survives the change.
```
daniel-noland: remove prose

nat/src/masquerade/contract.rs
```diff
@@ -0,0 +1,157 @@
...
+                    after: self.after,
+                });
+            }
+            Ok(())
+        }
+    }
+
+    // A distinctive fragment of the quote as well as the bare identifier. Section 9 holds
+    // six requirements, so `contains(SECTION_9, "REQ-12")` alone keeps passing if the text
+    // beside REQ-12 drifts -- which is the drift a build failure is worth having.
```
daniel-noland: remove prose

nat/src/masquerade/expiry.rs
```diff
@@ -196,6 +386,74 @@ fn an_expired_flow_is_never_resurrected() {
...
+//= https://www.rfc-editor.org/rfc/rfc4787#section-4.3
+//= type=test
+//# REQ-6:  The NAT mapping Refresh Direction MUST have a "NAT Outbound
+//# refresh behavior" of "True".
+#[test]
+fn outbound_traffic_keeps_an_unanswered_mapping_alive() {
+    // Half a lifetime, so `REFRESHES + 1` of them lands past the deadline the first
+    // packet set while each single step stays inside it. The assertion below is what
+    // caught this being written as a literal: at `emulated`'s 100x scale six seconds
+    // was nowhere near the 500s deadline and neither half of the test proved anything.
```
daniel-noland: remove prose

nat/src/masquerade/expiry.rs
```diff
@@ -93,6 +114,122 @@ fn reply_to(
...
+fn pool_exhaustion_splits_a_host_across_public_addresses() {
+    use std::collections::{BTreeMap, BTreeSet};
+    with_paused_clock(|| async {
+        let (fabric, _) = fabric();
+        let (mut lookup, mut masq) = fabric.stages();
+        let peer = fabric.peer[0];
+        let mut given: BTreeMap<IpAddr, BTreeSet<IpAddr>> = BTreeMap::new();
+
+        // Port-major, so every host is still drawing when the first address runs out. The
+        // host-major order this replaced never put a boundary inside one host's sessions.
```
daniel-noland: remove prose

nat/src/masquerade/apalloc/alloc.rs
```diff
@@ -117,8 +117,22 @@ impl<I: NatIpWithBitmap> IpAllocator<I> {
...
         // describe allocator failure and must be preserved.
+        //
+        // Reuse-before-draw gives a host one public address for as long as the pool does not
+        // move on, which is why this reads as Paired in the common case -- but the identity
+        // that would make it Paired is not carried: `allocate` never sees the internal IP, so
+        // it cannot return to an address this host already holds. Interleave two hosts across
+        // an exhaustion boundary and both spill;
+        // `expiry::pool_exhaustion_splits_a_host_across_public_addresses` measures 254 of 254
+        // hosts split. Implementing REQ-2 means keying the choice on the internal IP for the
+        // lifetime of its mappings.
```
daniel-noland: remove prose

mgmt/src/processor/proc.rs
```diff
@@ -633,6 +649,9 @@ impl ConfigProcessor {
 
         let overlay = config.external().overlay();
 
+        /* reject a port-forwarding ruleset the table cannot hold *before* anything is swapped */
```
daniel-noland: remove prose

.duvet/config.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[[source]]
+pattern = "*/src/*.rs"
+
+# Every RFC tracked here restates its numbered requirements in a summary section (RFC 4787
+# section 12, RFC 5382 section 8, RFC 5508 section 9). duvet keys citations by section
+# anchor and cannot be told that two sections carry the same requirement, so each of those
+# copies reads as uncovered no matter what is annotated. The convention, settled in the
+# RFC 5508 chapter, is to cite the *normative* section -- where the requirement is argued
+# rather than tabulated -- and to read the summary sections as permanently uncited.
```
daniel-noland: remove prose

.duvet/config.toml
```diff
@@ -0,0 +1,34 @@
+'$schema' = "https://awslabs.github.io/duvet/config/v0.4.0.json"
+
+# Citations live next to the code they annotate, so every crate's `src` tree is a source.
+# Both patterns are needed: `**` does not match zero path components here, so the first
+# finds `<crate>/src/<dir>/<file>.rs` and the second finds `<crate>/src/lib.rs`. Getting
+# this wrong is silent in the dangerous direction -- an unmatched file simply contributes
+# no citations, the snapshot does not change, and `duvet-check` passes.
```
daniel-noland: remove prose

nat/src/masquerade/protocol.rs
daniel-noland: hmm, may want to keep this.  It directly bears on the the duvet citation

net/src/headers/embedded.rs
```diff
@@ -48,15 +48,12 @@ pub struct EmbeddedHeaders {
...
-/// base RFC and the project had recorded the constant as written without them. RFC 4884 has
-/// exactly one erratum -- ID 3, Verified, Technical, against section 7 -- and it replaces
-/// "the data structure" with "the ICMP Extension Structure" in the description of the
-/// extension checksum. Nothing touches section 3, section 4, this floor, or the padding and
-/// length-attribute rules cited in `check_full_payload`. Checked 2026-09-06 against
-/// <https://errata.rfc-editor.org/rfc4884>; re-check if a second erratum appears.
+/// Errata checked 2026-09-06: <https://errata.rfc-editor.org/rfc4884>. The only erratum,
+/// ID 3 (Verified, Technical), clarifies the section 7 extension checksum. It does not
+/// change this minimum or the length and padding rules in `check_full_payload`.
+/// The vendored RFC omits errata; recheck if new ones are published.
```
daniel-noland: I don't think we need this comment

net/src/headers/embedded.rs
```diff
@@ -1397,37 +1496,92 @@ mod tests {
...
+        for ipv6 in [false, true] {
+            for extension in malformed {
+                let headers = parse_icmp_quote(ipv6, 128, 0, extension, true);
+                assert!(
+                    !headers.is_full_payload(),
+                    "ipv6={ipv6}, extension={extension:?}"
+                );
+            }
+        }
+    }
```
daniel-noland: this is ugly but perhaps necessary

nat/src/masquerade/test_state_machine.rs
```diff
@@ -196,14 +197,9 @@ fn ordinary_udp_opens_and_settles() {
...
-//= type=test
-//# REQ-10:  Receipt of any sort of ICMP message MUST NOT terminate the
-//# NAT mapping or TCP connection for which the ICMP was generated.
-//= https://www.rfc-editor.org/rfc/rfc4787#section-9
-//= type=test
-//# REQ-12:  Receipt of any sort of ICMP message MUST NOT terminate the
-//# NAT mapping.
+// This tests Echo Reply transitions. RFC 5382 REQ-10 and RFC 4787 REQ-12 also cover
+// ICMP errors, which go through `icmp_handler::nf`. This test alone cannot verify
+// either requirement.
```
daniel-noland, line 200: back and forth?

development/code/spec-compliance.md
```diff
@@ -0,0 +1,200 @@
...
+  `unshare -rn`.
+- **The mirror is a drop-in for the network fetch.** Text pulled by rsync is byte-identical to what
+  duvet fetches from rfc-editor.org, and re-extraction produces identical requirements apart from
+  the `target` line.
+- **`see_also` is empty** in the per-RFC JSON, so the RFC-to-BCP mapping has to come from the `bcp/`
+  symlinks or the index files, not the metadata.
+- **`-n.json`** in the corpus is an all-null artifact of the mirror. It parses, so it will not crash
+  a loader, but anything iterating the JSON should skip a null `doc_id`.
+
+[duvet]: https://awslabs.github.io/duvet/
```
daniel-noland, line 149: this whole file should be dropped.  No justification should be given here

development/code/README.md
```diff
@@ -12,6 +12,8 @@
   wrong thing.
 
 If you need to write a test, prefer [property-based tests] over simple unit tests.
+To find out whether they are saying what the specification asked for, see the
+[specification compliance note][duvet].
```
daniel-noland: remove

development/code/README.md
```diff
@@ -21,6 +23,7 @@ If you need to [handle an error][error], follow the guidelines.
 
 [avoid-global-reasoning]: ./avoid-global-reasoning.md
 [property-based tests]: ./property-testing.md
+[duvet]: ./spec-compliance.md
```
daniel-noland, line 30: remove

.duvet/config.toml
```diff
@@ -1,8 +1,23 @@
 '$schema' = "https://awslabs.github.io/duvet/config/v0.4.0.json"
 
+# Citations live next to the code they annotate, so every crate's `src` tree is a source.
+# Both patterns are needed: `**` does not match zero path components here, so the first
+# finds `<crate>/src/<dir>/<file>.rs` and the second finds `<crate>/src/lib.rs`. Getting
+# this wrong is silent in the dangerous direction -- an unmatched file simply contributes
+# no citations, the snapshot does not change, and `duvet-check` passes.
```
daniel-noland: remove prose

.duvet/config.toml
```diff
@@ -1,8 +1,23 @@
...
 
+[[source]]
+pattern = "*/src/*.rs"
+
+# Every RFC tracked here restates its numbered requirements in a summary section (RFC 4787
+# section 12, RFC 5382 section 8, RFC 5508 section 9). duvet keys citations by section
+# anchor and cannot be told that two sections carry the same requirement, so each of those
+# copies reads as uncovered no matter what is annotated. The convention, settled in the
+# RFC 5508 chapter, is to cite the *normative* section -- where the requirement is argued
+# rather than tabulated -- and to read the summary sections as permanently uncited.
```
daniel-noland: remove prose

justfile
```diff
@@ -768,10 +769,21 @@ duvet-check:
         exit 1
       fi
     done
+    # `duvet report` overwrites the snapshot and requirement files. Back them up and
+    # restore them on exit, whether the check succeeds or fails.
```
daniel-noland: remove prose

.gitattributes
```diff
@@ -1,3 +1,6 @@
 k8s-intf/src/generated/** linguist-generated
 k8s-intf/src/generated/mod.rs -linguist-generated
 npins/* linguist-generated
+# Generated by `duvet report`; includes verbatim RFC text.
```
daniel-noland: remove prose

scripts/spec-interlock.ts
```diff
@@ -620,7 +620,9 @@ function parseArgs(argv: string[]) {
     only: [] as string[],
     jobs: 4,
     output: join(REPO, "target", "spec-interlock"),
-    json: "/tmp/duvet-interlock.json",
+    // Keep output in this checkout to avoid collisions between concurrent CI jobs
+    // sharing the runner's /tmp.
```
daniel-noland: remove prose

scripts/duvet-summary.ts
```diff
@@ -64,6 +64,31 @@ function record(stats: Stats, status: Record<string, number | number[]>) {
   if (status.todo) stats.todos += 1;
 }
 
+// Summary sections repeat requirements from the normative sections we cite. Duvet counts
+// each copy separately, leaving the summary copies uncovered. Report them separately so
+// they do not obscure coverage of the normative sections; see `.duvet/config.toml`.
```
daniel-noland: drop prose

scripts/duvet-summary.ts
```diff
@@ -64,6 +64,31 @@ function record(stats: Stats, status: Record<string, number | number[]>) {
...
+const SUMMARY_SECTIONS: Record<string, string> = {
+  "https://www.rfc-editor.org/rfc/rfc4787": "section-12",
+  "https://www.rfc-editor.org/rfc/rfc5382": "section-8",
+  "https://www.rfc-editor.org/rfc/rfc5508": "section-9",
+};
+
+// The report omits section origins, so count duplicates in the extracted requirements.
+// Missing summary sections or files return zero, leaving the coverage denominator unchanged.
```
daniel-noland: drop prose

.github/workflows/dev.yml
```diff
@@ -390,6 +390,22 @@ jobs:
...
 
+      - name: "duvet-check"
+        id: "duvet-check"
+        continue-on-error: true
+        uses: *just
+        with:
+          recipe: "duvet-check"
+
+      # Omit `id` to keep this summary out of the required lint checks enforced by
+      # `check-lint-wiring`. `duvet-check` above enforces compliance.
```
daniel-noland: drop prose

nat/src/masquerade/apalloc/mod.rs
```diff
@@ -99,6 +99,9 @@ mod setup;
 mod test_alloc;
 
 pub use port_alloc::AllocatedPort;
+// `expiry` is compiled only for tests.
```
daniel-noland: drop prose

nat/src/masquerade/expiry.rs
```diff
@@ -100,50 +103,58 @@ fn reply_to(
         .flatten()
 }
 
+/// Verify the current loss of address pairing when a public address runs out of ports.
+///
+/// Interleave allocations so each host has sessions on both sides of the capacity limit.
+/// This test records a known deviation from REQ-2 and REQ-11. When address pairing is
+/// implemented, replace it with a compliance test and update both exception annotations.
+/// See `apalloc::alloc::allocate` for the missing host-to-address association.
```
daniel-noland: drop prose

nat/src/masquerade/expiry.rs
```diff
@@ -100,50 +103,58 @@ fn reply_to(
...
         let peer = fabric.peer[0];
         let mut given: BTreeMap<IpAddr, BTreeSet<IpAddr>> = BTreeMap::new();
 
-        for host in 1..=254u16 {
-            let source: IpAddr = format!("10.0.0.{host}")
-                .parse()
-                .unwrap_or_else(|_| unreachable!());
-            for sport in 1024..1024 + 256u16 {
+        // Allocate once per host before moving to the next port, so each host spans
+        // the first public address's capacity limit.
```
daniel-noland: drop prose

nat/src/masquerade/expiry.rs
```diff
@@ -355,6 +366,72 @@ fn an_expired_flow_is_never_resurrected() {
...
 }
 
+//= https://www.rfc-editor.org/rfc/rfc4787#section-4.3
+//= type=test
+//# REQ-6:  The NAT mapping Refresh Direction MUST have a "NAT Outbound
+//# refresh behavior" of "True".
+#[test]
+fn outbound_traffic_keeps_an_unanswered_mapping_alive() {
+    // Refresh before each deadline, then check past the original deadline. Derive the
+    // interval from the timeout so the test also covers expiry under `emulated`.
```
daniel-noland: drop prose

net/src/headers/embedded.rs
```diff
@@ -48,6 +48,7 @@ pub struct EmbeddedHeaders {
     full_payload_length: Option<u16>,
 }
 
+/// Minimum original-datagram field length for RFC 4884 section 4 compatibility.
```
daniel-noland: drop prose

net/src/headers/embedded.rs
```diff
@@ -237,61 +238,42 @@ impl EmbeddedHeaders {
...
+            //# "original datagram" field MUST contain at least 128 octets.
+            if icmp_length < MIN_ORIGINAL_DATAGRAM_OCTETS {
+                return;
+            }
+            // `icmp_length` is the length attribute multiplied by four for ICMPv4 or
+            // eight for ICMPv6, so parsed lengths already satisfy the alignment rules.
+            //
+            // Senders must pad to `max(128, round_up(len, alignment))`. As a receiver,
+            // we also accept longer fields if all bytes after the datagram are zero:
+            // the datagram is complete and the extension starts after the declared field.
```
daniel-noland: drop prose

net/src/headers/embedded.rs
```diff
@@ -237,61 +238,42 @@ impl EmbeddedHeaders {
...
-                    }
-                    return;
-                }
-                None => {
-                    unreachable!() // Checked earlier in the function
-                }
+            debug_assert!(self.net.is_some(), "checked earlier in the function");
+
+            if icmp_length < full_packet_length {
+                // The embedded message is shorter than the original packet
```
daniel-noland, line 241: drop prose

net/src/headers/embedded.rs
```diff
@@ -1356,21 +1338,53 @@ mod tests {
...
+            Some(0),
+            "the quoted UDP datagram carries no payload beyond its header"
+        );
+    }
+
+    /// Accept a complete datagram with extra zero padding.
+    ///
+    /// A conformant sender pads this 120-octet datagram to 128 octets. The receiver
+    /// also accepts longer fields, as explained in `check_full_payload`. Parsed lengths
+    /// are always aligned because the ICMPv4 length attribute counts 32-bit words.
```
daniel-noland: drop prose

justfile
```diff
@@ -772,12 +772,17 @@ duvet-check:
     # `duvet report` overwrites the snapshot and requirement files. Back them up and
     # restore them on exit, whether the check succeeds or fails.
     committed="$(mktemp -d)"
+    # Install the restore trap only after the backup succeeds, so a failed copy cannot
+    # delete the originals. Until then, clean up only the temporary directory.
```
daniel-noland: drop prose

justfile
```diff
@@ -772,12 +772,17 @@ duvet-check:
...
+    trap 'rm -rf "${committed}"' EXIT
+    # Preserve timestamps: nat includes these files with `include_str!`, so fresh
+    # timestamps would trigger a rebuild after every check.
```
daniel-noland: drop prose

mgmt/src/tests/mgmt.rs
```diff
@@ -1152,9 +1152,7 @@ mod validator_completeness {
         let ruleset = build_port_forwarding_configuration(vpc_table).unwrap_or_else(|e| {
             panic!("{mutation:?}: validator accepted a config port forwarding rejects: {e}")
         });
-        // Cannot fire yet: `validate_ruleset` is a stub returning `Ok`, so `update_table`
-        // is infallible. The (04) PR in this stack gives it a real `PortFwTable::dry_run`,
-        // at which point this starts checking something. Left in place for that.
+        // Check that rules accepted by the config validator also pass the table's dry run.
```
daniel-noland: drop prose

mgmt/src/processor/proc.rs
```diff
@@ -543,6 +543,19 @@ fn apply_masquerade_config(
...
 }
 
+/// Validate port-forwarding rules without modifying any tables.
+///
+/// `apply_config` updates four other tables before calling [`apply_port_forwarding_config`].
+/// A failure there leaves those changes applied without publishing their generation ID.
+/// Validating first prevents that partial update when port-forwarding rules are rejected.
+///
+/// `apply_router_config` can still fail after the table updates. Making the entire operation
+/// atomic requires a two-phase apply.
```
daniel-noland: drop prose

nat/src/portfw/portfwtable/access.rs
```diff
@@ -29,7 +29,11 @@ impl Absorb<PortFwTableChange> for PortFwTable {
 pub struct PortFwTableWriter(WriteHandle<PortFwTable, PortFwTableChange>);
 pub struct PortFwTableReader(ReadHandle<PortFwTable>);
 
-fn validate_ruleset(ruleset: &[PortFwEntry]) -> Result<(), PortFwTableError> {
+/// Check whether `update_table` would accept this ruleset without modifying the table.
+///
+/// Call this before updating related tables to avoid partial configuration changes
+/// if the port-forwarding rules are rejected.
```
daniel-noland: drop prose

mgmt/src/processor/proc.rs
```diff
@@ -633,6 +646,9 @@ impl ConfigProcessor {
 
         let overlay = config.external().overlay();
 
+        /* reject a port-forwarding ruleset the table cannot hold *before* anything is swapped */
```
daniel-noland: drop prose

net/src/headers/embedded.rs
```diff
@@ -51,6 +51,38 @@ pub struct EmbeddedHeaders {
 /// Minimum original-datagram field length for RFC 4884 section 4 compatibility.
 const MIN_ORIGINAL_DATAGRAM_OCTETS: usize = 128;
 
+// RFC 4884 sections 7–9: validate the envelope without interpreting object classes.
```
daniel-noland: drop prose

net/src/headers/embedded.rs
```diff
@@ -51,6 +51,38 @@ pub struct EmbeddedHeaders {
...
+fn valid_icmp_extension_structure(buf: &[u8]) -> bool {
+    // A four-byte header must be followed by at least one four-byte object header.
+    if buf.len() < 8 || buf[0] != 0x20 || buf[1] != 0 {
```
daniel-noland: 0x20 should be a const

net/src/headers/embedded.rs
```diff
@@ -1316,26 +1355,88 @@ mod tests {
...
-        assert_eq!(buf.len(), 120, "the embedded packet is 120 octets");
-        buf.extend(std::iter::repeat_n(padding_byte, field_len - buf.len()));
-        assert_eq!(buf.len(), field_len);
-        buf.extend_from_slice(&[0x55u8; 32]);
-        let (headers, consumed) =
-            EmbeddedHeaders::parse_with(EmbeddedIpVersion::Ipv4, &buf).unwrap();
-        (headers, consumed.get() as usize, buf)
-    }
+    // Version 2, checksum 0xddb5, one MPLS object containing a single label.
+    const ICMP_EXTENSION: [u8; 12] = [0x20, 0, 0xdd, 0xb5, 0, 8, 1, 1, 0, 1, 1, 64];
```
daniel-noland, line 1301: this is super awkward.

net/src/headers/embedded.rs
daniel-noland: I'm tempted to replace this with a fuzzer which generates malformed headers.  We have not done that so far, and it is a scope extension, so I'm going to let it slide for now.

net/src/headers/embedded.rs
```diff
@@ -1465,8 +1617,7 @@ mod tests {
         );
         assert!(!headers.is_full_payload());
 
-        // Try again with a value too small for the ICMP payload length: a valid payload size, but
-        // the padding area does not contain zeroed bytes
+        // A field shorter than the original datagram is incomplete.
```
daniel-noland: drop prose

nat/src/masquerade/expiry.rs
```diff
@@ -100,6 +100,105 @@ fn reply_to(
...
+//= https://www.rfc-editor.org/rfc/rfc4787#section-8
+//= type=test
+//= reason=held for the mapping dimension: pairing is unchanged by exhaustion, measured below
+//# REQ-11:  A NAT MUST have deterministic behavior, i.e., it MUST NOT
+//# change the NAT translation (Section 4) or the Filtering
+//# (Section 5) Behavior at any point in time, or under any particular
+//# conditions.
+#[test]
+#[ignore = "characterization probe; run with --ignored --nocapture"]
+fn pairing_is_unchanged_by_pool_exhaustion() {
```
qmonnet: I'm not sure I understand what this tests. I expect the assert to fail today, because we'd spill on another IP in case of pressure. Is this the reason why the test is currently ignored? If it's ignored, I'm not sure I understand the motivation for adding it at all?

nat/src/masquerade/mod.rs
```diff
@@ -17,9 +18,13 @@ mod state;
 mod test;
 mod test_state_machine;
 
+// Hairpinning belongs in the source-NAT path of `nf::Masquerade`. It must detect
+// destinations in its own public address pool and translate both source and destination
+// to send the packet back inward. This needs `apalloc` and the inbound reverse lookup.
```
qmonnet: I don't think this is relevant until we define how/whether hairpinning should work with the current VPC model?

nat/src/masquerade/contract.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+pub(crate) mod rfc4787 {
+    use super::{NatFlowStatus, Violation};
+
+    #[derive(Debug, Clone, Copy)]
+    pub(crate) struct Req12 {
+        before: NatFlowStatus,
+        after: NatFlowStatus,
+    }
```
qmonnet: I'm struggling to understand the need for that. We seem to add a new requirement validation model, which in the end is used only for REQ-12 of RFC 4787, if I'm correct. This does not look consistent with the rest of the duvet tooling plumbing, and seems to be yet another aspect to maintain?

Should we keep the check local, with the relevant duvet annotations instead?

nat/src/masquerade/contract.rs
```diff
@@ -3,51 +3,99 @@
...
+    // Check both the requirement ID and a distinctive phrase. The ID alone would not
+    // detect changes to the requirement's text.
+    const _: () = assert!(
+        contains(SECTION_9, <Req12 as Requirement>::ID),
+        "rfc4787#section-9 does not state REQ-12"
+    );
+    const _: () = assert!(
+        contains(SECTION_9, "MUST NOT terminate the"),
+        "rfc4787#section-9 no longer states REQ-12 as this contract quotes it"
+    );
```
qmonnet: I'm lost here, do we expect the RFC contents to change and guarding against that?

development/code/spec-compliance.md
```diff
@@ -0,0 +1,200 @@
+# Specification compliance with duvet
+
+Status: **two specifications tracked, the RFC corpus audited, the procedure itself not yet proven.**
+The open-questions list below is expected to grow; it is written down so that it grows in one place
+rather than in four people's heads.
```
qmonnet: What is this document for? I thought you mentioned you wanted to remove it (but I could be wrong).

net/src/icmp4/mod.rs
```diff
@@ -636,17 +636,18 @@ impl Icmp4 {
             self.payload_length(&cursor.inner[start..end])
         };
 
-        let (mut headers, consumed) = EmbeddedHeaders::parse_with(
-            EmbeddedIpVersion::Ipv4,
-            &cursor.inner[cursor.inner.len() - cursor.remaining as usize..],
-        )
-        .ok()?;
+        let embedded_start = cursor.inner.len() - cursor.remaining as usize;
+        let embedded_remaining = cursor.remaining as usize;
```
qmonnet: Define `embedded_start` (I'd call it `icmp_payload_start` for consistency) before `icmp_payload_length`, and avoid re-defining the identical `let end` on L634?

```rust
        let icmp_payload_start = cursor.inner.len() - cursor.remaining as usize;
        let icmp_payload_length = {
            let icmp_header_start = end.checked_sub(self.size().get() as usize)?;
            self.payload_length(&cursor.inner[icmp_header_start..icmp_payload_start])
        };
        let icmp_payload_remaining = cursor.remaining as usize;
        ...
```

This would read clearer to me

(same for IPv6 variant, obviously)

nat/src/masquerade/apalloc/mod.rs
qmonnet: Nit: This comment seems unnecessary and likely to go stale

nat/src/portfw/portfwtable/objects.rs
```diff
@@ -252,6 +252,14 @@ impl PortFwTable {
         Self::default()
     }
 
+    pub(crate) fn dry_run(ruleset: &[PortFwEntry]) -> Result<(), PortFwTableError> {
+        let mut scratch = Self::default();
+        for rule in ruleset {
+            scratch.add_entry(Arc::new(rule.clone()))?;
+        }
+        Ok(())
+    }
```
qmonnet: `dry_run()` doesn't look like a very clean solution, it would make more sense to validate that rules are unique in the set?

net/src/headers/embedded.rs
qmonnet: Should there also be a check on max length? Section 7:

    As stated above, the total length of the ICMP message, including
    extensions, MUST NOT exceed the minimum reassembly buffer size.

(576 octets for IPv4, 1280 (minimum MTU) for IPv6)

net/src/headers/embedded.rs
qmonnet, line 255: I double-checked this since I commented yesterday.

On one side, the annotated requirement remains wrong: we don't check whether we have an Extension Structure before validating the padding, so the presence of the "original datagram" field is not enough to require the padding.

However, RFC 4884, section 4 (and not 3) says:

    For ICMPv4 messages, the length attribute represents 32-bit words.
    When the length attribute is specified, the "original datagram" field
    MUST be zero padded to the nearest 32-bit boundary.

    [...]

    For ICMPv6 messages, the length attribute represents 64-bit words.
    When the length attribute is specified, the "original datagram" field
    MUST be zero padded to the nearest 64-bit boundary.

We are under `if icmp_length > 0`, so the check on padding is correct after all; but the requirement in the annotation is not the right one.

net/src/headers/embedded.rs
```diff
@@ -247,19 +279,26 @@ impl EmbeddedHeaders {
             if icmp_length > buf.len() {
                 return;
             }
+            let extensions = &buf[icmp_length..];
+            let has_extensions = !extensions.is_empty();
+            // A nonzero length locates any extension; it does not establish its presence.
+            // Malformed trailing data leaves completeness unknown.
+            if has_extensions && !valid_icmp_extension_structure(extensions) {
+                return;
+            }
```
qmonnet: Logically, validating extensions' structure (and locating extensions in the first place) should be called from the parent function, when parsing ICMP payload, after retrieving `icmp_length`, but not from the function that asserts whether the original datagram is full or not.

Also, why would we return early here? We would still consider we have the full original datagram, even if the trailing extensions are garbage? Depends on how conservative we want to be, I suppose. Same for the check below with `MIN_ORIGINAL_DATAGRAM_OCTETS`, we may still have the full original datagram even if the sender didn't respect this requirement (although arguably, this also comes with a chance the packet is bugged)

net/src/headers/embedded.rs
qmonnet, line 1301: I had to look up why this comment mentioned MPLS. Maybe add a link to https://www.iana.org/assignments/icmp-parameters#icmp-parameters-ext-class-1, please?

net/src/headers/embedded.rs
```diff
@@ -51,6 +51,38 @@ pub struct EmbeddedHeaders {
 /// Minimum original-datagram field length for RFC 4884 section 4 compatibility.
 const MIN_ORIGINAL_DATAGRAM_OCTETS: usize = 128;
 
+// RFC 4884 sections 7–9: validate the envelope without interpreting object classes.
+fn valid_icmp_extension_structure(buf: &[u8]) -> bool {
```
qmonnet: Do we need to do this validation? Section 9 says “application software” should do it, not the dataplane? The only thing it changes is:

```
            if has_extensions && icmp_length < MIN_ORIGINAL_DATAGRAM_OCTETS {
                return;
            }
```

But if we have trailing bytes after `icmp_length` we should probably consider them as (invalid) extensions, and enforce `icmp_length < MIN_ORIGINAL_DATAGRAM_OCTETS` anyway?

I understand we won't be dealing with ICMP extensions so often, but I still don't really see the point, given that we won't reject the packet anyway.

net/src/headers/embedded.rs
```diff
@@ -273,7 +271,6 @@ impl EmbeddedHeaders {
             debug_assert!(self.net.is_some(), "checked earlier in the function");
 
             if icmp_length < full_packet_length {
-                // The embedded message is shorter than the original packet
```
qmonnet: This comment was moved, not introduced, in the PR, so I'm not sure why you remove it. This one doesn't matter much, but I hope (haven't checked) that this commit (`docs: remove prose cited in spec-compliance review`) doesn't remove other unrelated comments that predated the PR.

qmonnet (commented):
Please find some more comments below.

One side note: `duvet` is yet another tool that get introduced without getting a brief in a README, and will be hard for anyone else to maintain.

nat/src/masquerade/expiry.rs
daniel-noland: It does fail.  The test is documenting the behavior we _should_ get.  It is basically the property we need to aim for.

nat/src/masquerade/mod.rs
daniel-noland: I agree.  It is just tracking the requirement.  We can remove it, but I think it is fine as is personally.  It shows up in the report as a todo item this way

nat/src/masquerade/contract.rs
daniel-noland: hmmm, that is fair.  I think it was originally split and then refactored and this is just left over.  Will clean up if possible

nat/src/masquerade/contract.rs
daniel-noland: this was intended to interact with the interlock script.  It will be more clear later, but I think I can remove this from this PR for now.

net/src/headers/embedded.rs
daniel-noland: yup.  I found the same thing.  Will push the fix shortly.

Sadly the tests for the other half are ugly

daniel-noland:
> One side note: `duvet` is yet another tool that get introduced without getting a brief in a README, and will be hard for anyone else to maintain.

You want design docs?

net/src/headers/embedded.rs
daniel-noland: fixed.  I can't find any other examples of this.

net/src/headers/embedded.rs
daniel-noland: that's fair.  I can look into that

net/src/headers/embedded.rs
daniel-noland: hmm, that is a reasonable point.

We may need to exempt ourselves from that check.  I'll take a look

net/src/headers/embedded.rs
daniel-noland: Honestly, I hate the way I did this so much that I'm considering removing it.  I don't want to include an MPLS parse/deparse logic just to make this, but I hate the hard coded slice.

And for this obscure case... it is just ugly.

That said, this is exactly the type of thing duvet is for.  It is specifically so we can test that we comply with specs, and this is the sepc.  So I think we should test and I think we should cite why.  So yeah, I'll fix.

nat/src/portfw/portfwtable/objects.rs
daniel-noland: yeah, that makes sense.  Will fix

mvachhar:
> > One side note: `duvet` is yet another tool that get introduced without getting a brief in a README, and will be hard for anyone else to maintain.
> 
> You want design docs?

Maybe a short blurb in README or somewhere else appropriate on how to update the duvet stuff?

nat/src/masquerade/protocol.rs
mvachhar: If you think it is helpful you can keep it.

net/src/headers/embedded.rs
daniel-noland: I think this is now sorted.  Take a look please

net/src/headers/embedded.rs
daniel-noland: sorted

net/src/headers/embedded.rs
daniel-noland: fixed

net/src/headers/embedded.rs
daniel-noland: I think this is fixed but please let me know if you disagree

net/src/icmp4/mod.rs
daniel-noland: sorted

nat/src/masquerade/contract.rs
daniel-noland: removed

nat/src/masquerade/contract.rs
daniel-noland: removed

mvachhar (approved):
Ok, I think this looks good now.  I believe @qmonnet's concerns have been addressed as have mine.

PR #1808, opened 2026-09-08
dpdk + incremental checksums

daniel-noland:
> **Draft / do-not-merge.** This is proof-of-concept work to de-risk the DPDK driver's path to
> production and to make a fair kernel-vs-DPDK benchmark possible. It is one WIP commit covering
> five separable pieces; it wants splitting into a proper series before it is reviewable, and the
> git history is deliberately not clean yet.
>
> Purpose of this PR: build a release image and put it through vlab. If that is green, a benchmark
> gets scheduled.

Based on `pr/daniel-noland/dpdk-reconciled`.

## What is in it

**RSS is on.** The driver configured `rss: None`, so every frame landed on receive queue 0 and one
worker did all the work however many were configured. Now a standard Toeplitz key over whatever
subset of L3/L4 the device advertises, with `RSS_HASH` so the NIC reports the hash it steered by.
Measured on a BlueField-3: 4 workers, 40 000 frames, **9941 / 10050 / 10054 / 9955**. With RSS
disabled the same run puts all of it on worker 0.

Deliberately *not* symmetric. mlx5 rejects any hash function but the default at
`rte_eth_dev_configure` (symmetric Toeplitz is reachable only through the `rte_flow` RSS action),
and it would not help anyway: a NAT'd flow's reverse packet carries the *translated* tuple, not the
reversed one. Flow state lives in one shared `FlowTable`, so it does not need to.

**Port counters reach Prometheus.** `Dev::stats()` existed and was read only by a hardware test, so
`imissed` / `rx_nombuf` — the counters that distinguish "the wire is quiet" from "we are dropping" —
were invisible. Eight `port_*` counters now, polled by the DPDK supervisor (the only thread that
*can*: `Dev::stats` needs an `'eal`-branded device and `Eal` is `!Send`).

**Workers register as EAL lcores.** They were plain threads reporting `LCORE_ID_ANY`, for which
`rte_mempool_default_cache` returns `NULL` — so every `alloc_bulk` and every mbuf free went to the
shared ring under atomics, on every burst, both directions. The registration is now a `!Send +
!Sync` capability token (`dpdk::lcore::LCore`) that gates the APIs which genuinely require it; the
first of those is the new `dpdk::power`, the primitives for a halfway
interrupt/poll receive loop.

**Incremental NAT checksums (RFC 1624).** The algorithm already existed and was used for ICMP error
quoted headers, where a full recompute is impossible. The main NAT path did not use it: every
translation set `checksum_refresh`, and `ipforward` re-summed the whole payload. All four paths
(`masquerade` snat/dnat, `portfw` snat/dnat, `static_nat`) now fold the change in incrementally.
This is software, so **both drivers benefit** — which is what makes the comparison fair.

**Six hardware probes**, so these findings fail loudly on a DPDK bump rather than rotting:
`rss_capability`, `lcore_role`, `lcore_capacity`, `power_monitor`, `flow_async_lcore`,
`checksum_offload`.

## Verification

827 tests, clippy and rustdoc clean. Everything hardware-dependent was measured on the workstation
BF3, not inferred. Every new test was break-tested — several were vacuous on the first attempt and
are noted as such in the code.

## Known gaps, deliberately left

- **ICMP-error branches in `static_nat`** stay on the full recompute. `icmp_error_msg` maintains the
  embedded headers incrementally and claims the outer ICMP checksum is not touched again at
  deparse, but that checksum covers the whole quoted packet and proving the two agree needs its own
  test. Correct and slower is the right default for a rare path.
- **Workers are registered but not pinned.** `rte_thread_register` reads affinity and never sets it.
  Pinning needs a policy decision (which CPU, NUMA relative to the port, container cpuset,
  `isolcpus`) and is worth doing against a load test — which `port_rx_missed` now makes possible.
- **VXLAN checksum offload is unmeasured.** The offload probe covered plain IPv4/UDP; this is a
  VXLAN gateway and `OUTER_*_CKSUM` is unsupported on receive, so what mlx5 reports for the inner
  headers of an encapsulated frame is genuinely open.
- **The async `rte_flow` check could not be completed on this card** — it is not in switchdev/HWS
  mode, and both the new probe and the pre-existing `template_probe` SIGSEGV identically at the
  first HWS call. Source says the async engine has no lcore coupling at all (zero references in
  `lib/ethdev/rte_flow.c` *or* `mlx5_flow_hw.c`); the hardware confirmation is outstanding.
- The CI `EXPERIMENT (dpdk in vlab)` scaffolding inherited from the base branch still needs
  reverting before any of this merges.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -142,19 +142,26 @@ jobs:
         uses: *gate
         with:
           labels: "test/all-profiles"
-          # The build matrix has no fuzz container targets.
-          on-value: '["debug", "release"]'
-          # Pull-request lab labels require both debug and release images.
-          off-value: >-
-            ${{
-              github.event_name == 'pull_request'
-              && (
-                contains(github.event.pull_request.labels.*.name, 'ci:+vlab')
-                || contains(github.event.pull_request.labels.*.name, 'ci:+hlab')
-              )
-              && '["debug", "release"]'
-              || '["debug"]'
-            }}
+          # EXPERIMENT (dpdk in vlab): release only, both ways.
+          #
+          # The gateway is being debugged against physical hardware, and only CI can push images,
+          # so every one-line fix costs a full round trip. Half of that trip was building a debug
+          # image nothing in this loop installs. The lab wants the release artifact; the debug one
+          # comes back with the rest of the scaffolding.
+          on-value: '["release"]'
+          off-value: '["release"]'
+          # Restore with the scaffolding:
+          # on-value: '["debug", "release"]'
+          # off-value: >-
+          #   ${{
+          #     github.event_name == 'pull_request'
+          #     && (
+          #       contains(github.event.pull_request.labels.*.name, 'ci:+vlab')
+          #       || contains(github.event.pull_request.labels.*.name, 'ci:+hlab')
+          #     )
+          #     && '["debug", "release"]'
+          #     || '["debug"]'
+          #   }}
 
       - id: "wasm"
         uses: *gate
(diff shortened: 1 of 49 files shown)


daniel-noland:
closing as experiment is complete.  Will be reborn with better design elsewhere

PR #1811, opened 2026-09-10
Pr/fredi/fuzz routing

Fredi-raspall:
N.B. this targets https://github.com/githedgehog/dataplane/pull/1753


routing/src/rib/nexthop.rs
```diff
@@ -427,6 +429,35 @@ impl NhopStore {
             .map(Rc::downgrade)
             .collect()
     }
+
+    /// Set the ifname for all nhops that have an ifindex, according to the interface table
+    /// This method is for future use
+    #[allow(unused)]
+    pub fn refresh_nhop_ifnames_all(&self, iftable: &IfTable) {
```
mvachhar: Please add some tests for this please

routing/src/cli/display.rs
```diff
@@ -21,7 +21,7 @@ use crate::router::cpi::{CpiStats, CpiStatus, StatsRow};
 
 use crate::rib::VrfTable;
```
mvachhar, line 22: The verb form of recursive is recur so the correct text in the commit message for `feat(routing): revisit next-hop resolution` is `recurring` not `recoursing`.

mvachhar (changes requested):

routing/src/rib/nexthop.rs
Fredi-raspall: That method was removed. The remaining one is tested.

routing/src/cli/display.rs
Fredi-raspall: Fixed.

routing/src/rib/vrf.rs
```diff
@@ -345,17 +347,25 @@ impl Vrf {
...
             nhop_refs.push(shim);
+        } else {
+            for nhop in nhops {
+                let shared = self.nhstore.add_nhop(&nhop.key);
+                let ext_vrf = if nhop.vrfid == self.vrfid {
+                    None
+                } else {
+                    Some(nhop.vrfid)
+                };
+                shared.set_ifname(nhop.ifname.as_ref());
```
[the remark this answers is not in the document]
Fredi-raspall: If the interface is no longer in the table, keeping the name is not a valid solution as it may be wrong.

PR #1827, opened 2026-09-15
CLI clean-up + requests by VPC name + dynamic completions

### LINKED ISSUE #1781: Make sure all dp-cli commands accepts and prints vpc name
E.g. show fib/route should accept vpc name instead of only id. Same for all outputs.

Fredi-raspall:
**NOTE**: this now targets https://github.com/githedgehog/dataplane/pull/1811

1) Add support to show ip route information for a given VPC name.
  This includes showing the routing tables for a vpc, the next-hops and the fib, for both ipv4 and ipv6.
  fixes https://github.com/githedgehog/dataplane/issues/1781
2) ip routes, fib, fib-groups and next-hops can be requested for a given VNI too.
3) add support to get routes targeting a given prefix (i.e. support prefix=....)
4) clean-up and simplify CLI code.
5) add support to query the router macs for a given vni or ip address (vni=...., address=....)
6) fix completion issue where argument would be swallowed.  
7) Add dynamic auto-completion where the cli pre-fetches the identifiers for a given argument.
    This is useful so that, for instance, when typing show ip route vpc=......... the vpc needs not be typed but selected
    from a list dynamically returned by dataplane.
    This auto-completion is implemented for vpc names, interface names (unused), vnis and router mac addresses and can
be easily extended.
8) Simplify & unify CLI argument parsing
9) Add support to filter routes by prefix-length

**Sample**: get the routes for VPC called VPC-3

```
dataplane(✔)# show ip route vpc=VPC-3
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Ipv4 routes (5) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Vrf: 'CCCCC-vrf' (id: 2) description: VPC-3

   0.0.0.0/0 other [0/0] 00:00:59
       (from VRF 0) action Drop
   192.168.100.0/24 bgp [20/0] 00:00:53
        via 7.0.0.5 encap Vxlan (vni 2000), remote 7.0.0.5 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
   192.168.100.1/32 bgp [20/0] 00:00:53
        via 7.0.0.5 encap Vxlan (vni 2000), remote 7.0.0.5 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
   192.168.128.0/27 bgp [20/0] 00:00:53
        via 7.0.0.4 encap Vxlan (vni 2000), remote 7.0.0.4 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
   192.168.128.1/32 bgp [20/0] 00:00:53
        via 7.0.0.4 encap Vxlan (vni 2000), remote 7.0.0.4 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
```

**Sample**: get the routes for VPC called VPC-3, targeting a specific prefix

```
dataplane(✔)# show ip route vpc=VPC-3 prefix=192.168.100.0/24
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Ipv4 routes (5) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Vrf: 'CCCCC-vrf' (id: 2) description: VPC-3

   192.168.100.0/24 bgp [20/0] 00:01:27
        via 7.0.0.5 encap Vxlan (vni 2000), remote 7.0.0.5 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)

  (Displayed 1 routes out of 5)
```

**Sample**: get the routes for a VPC given its vni

```
dataplane(✔)# show ip route vni=2000
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Ipv4 routes (5) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Vrf: 'CCCCC-vrf' (id: 2) description: VPC-3

   0.0.0.0/0 other [0/0] 00:02:03
       (from VRF 0) action Drop
   192.168.100.0/24 bgp [20/0] 00:01:58
        via 7.0.0.5 encap Vxlan (vni 2000), remote 7.0.0.5 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
   192.168.100.1/32 bgp [20/0] 00:01:58
        via 7.0.0.5 encap Vxlan (vni 2000), remote 7.0.0.5 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
   192.168.128.0/27 bgp [20/0] 00:01:58
        via 7.0.0.4 encap Vxlan (vni 2000), remote 7.0.0.4 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
   192.168.128.1/32 bgp [20/0] 00:01:58
        via 7.0.0.4 encap Vxlan (vni 2000), remote 7.0.0.4 dmac: --
            via 10.0.0.13 interface eth0 (idx 254)
            via 10.0.1.13 interface eth2 (idx 256)
```

**Sample**: get the router macs filtering by vni

```
dataplane(✔)# show evpn rmac vni=3000
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Router macs (entries: 6 stale: 0) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 vni   address              mac                status  
 3000  7.0.0.3              aa:bb:cc:dd:ee:00  ok      
 3000  7.0.0.1              52:60:6b:10:18:a3  ok      

  (Displayed 2 entries out of 6)
```

**Sample**: get the router macs filtering by remote vtep address

```
dataplane(✔)# show evpn rmac address=7.0.0.3
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Router macs (entries: 6 stale: 0) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 vni   address              mac                status  
 3000  7.0.0.3              aa:bb:cc:dd:ee:00  ok      
 4000  7.0.0.3              aa:bb:cc:dd:ee:00  ok      

  (Displayed 2 entries out of 6)
``` 

**Sample**: Dynamic autocompletion for vpc argument

```
dataplane(✔)# | show ip route vpc=
                                  >VPC-1
                                  VPC-2
                                  VPC-3
```

**Sample**: Dynamic autocompletion for router mac address

```
dataplane(✔)# | show evpn rmac address=
                                       >7.0.0.1
                                       7.0.0.3
                                       7.0.0.4
                                       7.0.0.5
```

**Sample**: Filtering routes by prefix length. This is especially useful in evpn

```
dataplane(✔)# show ip route vpc=VPC-1 prefix-len=24
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Ipv4 routes (7) ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 Vrf: 'AAAAA-vrf' (id: 5) VPC: VPC-1 vni: 3000

   192.168.50.0/24 bgp [20/0] 00:17:17
        via 7.0.0.1 encap Vxlan (vni 3000), remote 7.0.0.1 dmac: --
            via 10.0.0.13 interface eth0 (idx 220)
            via 10.0.1.13 interface eth2 (idx 222)
        via 7.0.0.3 encap Vxlan (vni 3000), remote 7.0.0.3 dmac: --
            via 10.0.0.13 interface eth0 (idx 220)
            via 10.0.1.13 interface eth2 (idx 222)
   192.168.60.0/24 bgp [20/0] 00:17:17
        via 7.0.0.1 encap Vxlan (vni 3000), remote 7.0.0.1 dmac: --
            via 10.0.0.13 interface eth0 (idx 220)
            via 10.0.1.13 interface eth2 (idx 222)
        via 7.0.0.3 encap Vxlan (vni 3000), remote 7.0.0.3 dmac: --
            via 10.0.0.13 interface eth0 (idx 220)
            via 10.0.1.13 interface eth2 (idx 222)

  (Displayed 2 routes out of 7)
```


routing/src/rib/vrftable.rs
```diff
@@ -244,6 +244,28 @@ impl VrfTable {
...
+
+    //////////////////////////////////////////////////////////////////
+    /// Immutably access a [`Vrf`] from its description.
+    /// For vpcs, the description matches the name of the vpc.
+    //////////////////////////////////////////////////////////////////
+    #[allow(unused)]
+    pub fn get_vrf_by_descr(&self, description: &str) -> Result<&Vrf, RouterError> {
+        self.by_id
+            .values()
+            .find(|vrf| vrf.description.as_deref() == Some(description))
```
qmonnet: Could we imagine having several VRFs with identical descriptions in the future? (At the moment, from what I understand, the description is the name of the VPC, so it's unique; but it could be something else in the future?) Should we handle it as such, or add a `debug_assert!()` to catch if we have several matches in the future?

routing/src/cli/handler.rs
```diff
@@ -113,23 +114,61 @@ fn show_ipv6_routes_multi(
     CliResponse::from_request_ok(request, out)
 }
 
-fn route_filter_v4(request: &CliRequest) -> RouteV4Filter {
-    let filter: RouteV4Filter = if let Some(protocol) = &request.args.protocol {
+fn route_filter_v4(request: &CliRequest) -> Result<RouteV4Filter, CliError> {
```
qmonnet: Reading the code (commit `feat(cli): dynamic filter building for show ip route/fib`), you can only refine the selection (“and”) by passing multiple filters, never extend it (“or”), right? You couldn't have something like `show ip protocol=bgp protocol=ospf` for example to print either of these?
(This is just a question, no change required)

cli/bin/cmdtree_dp.rs
```diff
@@ -182,9 +182,11 @@ fn cmd_show_evpn() -> Node {
         .desc("Show EVPN VRFs")
         .action(CliAction::ShowRouterEvpnVrfs);
 
-    root += Node::new("rmac-store")
-        .desc("Show the contents of the router mac store")
-        .action(CliAction::ShowRouterEvpnRmacStore);
+    root += Node::new("rmac")
```
qmonnet, line 205: We don't have any user relying on the CLI at the moment, right? So it's safe to change the name of an existing command?

qmonnet (commented):
It would be great to get some tests, for example to making sure that filters behave as expected, or that the prefetch works as expected, too.

Please find some other comments/questions inline below.

routing/src/rib/vrftable.rs
qmonnet: Oh that's actually one of the items that CodeRabbit reported, too

Fredi-raspall: I don't think so. The description is something that does NOT come from the user. We set it internally ourselves, always to the vpc name (exept in default vrf). I did not rename the "description" field, because that was the original purpose of the field and so as not to pollute it with something so specific as a vpc.
So, in practice, no two vrfs should get the same description. That being said, the vrftable does not enforce it by itself. I can add that.

routing/src/cli/handler.rs
Fredi-raspall: Yes, the conditions are ANDed, so we only refine (exclude things) as we add conditions.
What you're saying could be done, but we'd need to code the algebra. Also, when packing the parameters to dataplane, we only add one field (e.g. protocol) as a singleton, so unless we'd change that, no chance.
Finally, the cli client won't offer "protocol=" again if it has already been offered by the completion (that does not mean that a user could not write it, but it would be ignored, or it would err).

cli/bin/cmdtree_dp.rs
Fredi-raspall: It should be safe, but yes, we should treat that as an API.

routing/src/cli/handler.rs
qmonnet: Sound good enough for now, thanks for the clarification :+1:

cli/bin/cmdtree_dp.rs
```diff
@@ -88,15 +84,31 @@ fn cmd_show_vpc() -> Node {
...
+fn vni_arg() -> NodeArg {
+    NodeArg::new("vni").selector(PrefetchSelector::Vnis)
+}
+fn rmacaddr_arg() -> NodeArg {
+    NodeArg::new("address").selector(PrefetchSelector::RmacAddr)
+}
+#[allow(unused)]
+fn ifname_arg() -> NodeArg {
+    NodeArg::new("ifname").selector(PrefetchSelector::Interfaces)
+}
```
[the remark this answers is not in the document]
Fredi-raspall: The method is purposely unused. It is not clear in which cmds the "ifname" will be used.

routing/src/cli/handler.rs
```diff
@@ -58,145 +67,190 @@ fn show_vrf_ipv4_routes(vrf: &Vrf, filter: &RouteV4Filter) -> String {
...
+fn lookup_vrf<'a>(
+    vrftable: &'a VrfTable,
+    request: &CliRequest,
+) -> Result<Option<&'a Vrf>, CliError> {
+    if let Some(vpc) = &request.args.vpc {
+        let vrf = vrftable
+            .get_vrf_by_descr(vpc.as_str())
+            .map_err(|_| CliError::NotFound(format!("VRF with with descr {vpc}")))?;
+
+        Ok(Some(vrf))
```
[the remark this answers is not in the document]
Fredi-raspall: No, this is not the right fix.

mvachhar (changes requested):
@Fredi-raspall can you add some test coverage for these changes?  I know dataplane-cli doesn't have the best coverage and it is a debug tool, but it would be good to at least get something.  You can see here that almost none of the new code is covered by any testing.  Doesn't have to be perfect, but if we slowly add coverage over time, we'll eventually get good coverage here.

https://app.codecov.io/gh/githedgehog/dataplane/pull/1827?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=githedgehog

THe link shows coverage for the diffs in the PR and you can see almost none of the lines are exercised by tests

cli/bin/cmdtree_dp.rs
mvachhar: Yes, I think change a command name is fine provided that the `show tech-support` command for our overall system isn't using the command.  @Fredi-raspall did you check that our `show tech` scripts don't use this command?

Fredi-raspall: Yes, I believe we only call "show tech" and not show rmac-store.

Fredi-raspall:
> @Fredi-raspall can you add some test coverage for these changes? I know dataplane-cli doesn't have the best coverage and it is a debug tool, but it would be good to at least get something. You can see here that almost none of the new code is covered by any testing. Doesn't have to be perfect, but if we slowly add coverage over time, we'll eventually get good coverage here.
> 
> https://app.codecov.io/gh/githedgehog/dataplane/pull/1827?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=githedgehog
> 
> THe link shows coverage for the diffs in the PR and you can see almost none of the lines are exercised by tests

Ok. Will see where that's worth it.

routing/src/rib/vrftable.rs
Fredi-raspall: I've changed the code:
1) renamed description -> vpcname
2) I don't think we want to enforce in the vrftable that two vrfs have distinct vpcnames, as that would forbid a vpc with multiple vrfs. So, instead, we live with the fact that that may happen in the future and, if we look up vrfs from vpcname (we should only do that for the cli), we allow for multiple responses.

qmonnet: OK, I'm happy with the `.find()` -> `.filter()` change to return multiple responses if necessary, thanks!

cli/bin/prefetch.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::os::unix::net::UnixDatagram;
+
+/// Send a prefetch request. `PrefetchSelector` indicates the type of data to prefetch.
+fn do_prefetch(sock: &UnixDatagram, selector: PrefetchSelector) -> Option<PrefetchedData> {
+    let args = RequestArgs::with_selector(selector);
+    let request = CliRequest::new(CliAction::Prefetch, args);
+    if request.send(sock).is_err() {
+        return None;
+    }
+    CliResponse::recv_sync(sock)
```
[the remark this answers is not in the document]
Fredi-raspall: This has been addressed, by adding a recv() method with a time out.

Fredi-raspall:
> It would be great to get some tests, for example to making sure that filters behave as expected, or that the prefetch works as expected, too.
> 
> Please find some other comments/questions inline below.

I've refactored the filtering code and added tests for it.

cli/src/cliproto.rs
```diff
@@ -656,4 +656,153 @@ mod tests {
...
+            for _ in 0..num {
+                let mut raw = vec![0xAAu8; 16];
+                raw.push(1);
+                if sock.send(&raw).is_err() {
+                    break;
+                }
+                thread::sleep(every);
+            }
+        });
+    }
```
daniel-noland, line 682: you should use a scoped thread to ensure we join at the end.  This can leak

cli/src/cliproto.rs
```diff
@@ -656,4 +656,153 @@ mod tests {
...
     }
+
+    /// Build a directory to hold the sockets of a test, named after it so that
+    /// tests running in parallel can't collide.
+    fn sock_dir(test: &str) -> std::path::PathBuf {
+        let dir = std::env::temp_dir().join(format!("cliproto-{test}-{}", std::process::id()));
+        let _ = std::fs::remove_dir_all(&dir);
+        std::fs::create_dir_all(&dir).expect("sock dir should be creatable");
+        dir
+    }
```
daniel-noland, line 667: this can leak resources if you run the tests and they don't clean up.

Can we use an anonymous unix socket for this instead?  That is a much cleaner test

cli/src/cliproto.rs
```diff
@@ -656,4 +656,153 @@ mod tests {
...
+    /// rather than block the cli forever.
+    #[test]
+    #[cfg_attr(
+        emulated,
+        ignore = "Unix sockets unsupported under miri / flaky under qemu-user"
+    )]
+    fn recv_timeout_expires_when_unanswered() {
+        let dir = sock_dir("unanswered");
+        let sock = UnixDatagram::bind(dir.join("cli.sock")).expect("cli sock should bind");
+        let budget = Duration::from_millis(200);
```
daniel-noland, line 694: all use of Duration need to come from the clock facade

daniel-noland: the same goes for all time constructs

cli/src/cliproto.rs
```diff
@@ -656,4 +656,153 @@ mod tests {
...
+
+    /// Send `num` chunks, `every` apart, all flagged as "more chunks follow",
+    /// so that the receiver never sees the end of the message.
+    fn dribble_chunks(sock: UnixDatagram, num: usize, every: Duration) {
+        thread::spawn(move || {
+            for _ in 0..num {
+                let mut raw = vec![0xAAu8; 16];
+                raw.push(1);
+                if sock.send(&raw).is_err() {
+                    break;
```
daniel-noland, line 677: no log?

routing/src/cli/display.rs
daniel-noland: I ran a concurrency analysis on this and I think we still have some troubles:

```mermaid
sequenceDiagram
    autonumber
    participant FRR as FRR (zebra/CPI)
    participant CPI as cpi.rs IpRoute::add
    participant RA as rpc_adapt from_rpc_nhop
    participant IFT as IfTable
    participant VRF as Vrf::add_route_complete
    participant NH as Nhop (shared, key has no name)
    participant FIB as FibGroup / EgressObject
    participant CLI as show ip fib / next-hop

    Note over FRR,CLI: Case A: a route arrives before its interface does (the startup race)
    FRR->>CPI: route R1 via 10.0.0.1 ifindex 7
    CPI->>RA: from_rpc_nhop(nh, iftable)
    RA->>IFT: get_interface(7)
    IFT-->>RA: None (interface not in the table yet)
    RA-->>VRF: RouteNhop { key, ifname: None }
    VRF->>NH: create Nhop, set_ifname(None)
    VRF->>NH: instructions empty, so build_nhop_instructions()
    NH->>FIB: EgressObject { ifindex 7, ifname: None } (copy of the name)

    Note over IFT: later: interface 7 "eth0" is added
    CLI->>NH: display
    NH-->>CLI: "via 10.0.0.1 (idx 7)" (no name)

    Note over FRR,CLI: Case B: a second route shares the next-hop and the name is known now
    FRR->>CPI: route R2 via 10.0.0.1 ifindex 7
    CPI->>RA: from_rpc_nhop(nh, iftable)
    RA->>IFT: get_interface(7)
    IFT-->>RA: "eth0"
    RA-->>VRF: RouteNhop { key, ifname: "eth0" }
    VRF->>NH: reuse the shared Nhop, set_ifname("eth0")
    VRF--xNH: instructions not empty, so NOT rebuilt (vrf.rs:341)
    CLI->>NH: show ip next-hop
    NH-->>CLI: "interface eth0" (new name)
    CLI->>FIB: show ip fib
    FIB-->>CLI: ifname None (stale copy) until the next full refresh_nhops()

    Note over FRR,CLI: Case C: a later route arrives in a window where the lookup misses again
    FRR->>CPI: route R3 via 10.0.0.1 ifindex 7
    RA->>IFT: get_interface(7)
    IFT-->>RA: None (e.g. the interface is being re-created, or after a rename or restart)
    VRF->>NH: set_ifname(None) overwrites the known "eth0" name
    CLI->>NH: display
    NH-->>CLI: name gone again

    Note over FRR,CLI: Case D: a rename (eth0 to enp1s0) with no new route
    Note over IFT: the table now says ifindex 7 = "enp1s0"
    Note over NH,FIB: no path updates the Nhop or its EgressObject
    CLI->>NH: display
    NH-->>CLI: still "eth0" (stale), indefinitely
```

PR #1832, opened 2026-09-19
maintenance: frr 10.7.1, rdma-core 65.0, drop opengrep --config auto

daniel-noland:
Routine maintenance, folded into one PR rather than three so the review queue
sees a single change.

- **`ci: drop opengrep's --config auto`** — `auto` fetches
  `https://semgrep.dev/c/p/default` at scan time, and that URL serves two
  different documents depending on which client asks: the YAML rule file
  opengrep wants, or a JSON API envelope `{"missed": 1856, "rules": [...]}`
  that opengrep rejects with exit 7. Which one you get tracks neither
  User-Agent nor Accept, so the step was a coin flip and it has lost twice.
  Of the 1074 rules in that pack, 3 target rust and 3 bash, and it had never
  produced a finding here; `.semgrep/rules` is untouched and still fails the
  build on a hit. Full reasoning is in the comment above the recipe.
- **`fix(nix): read the FRR version from whichever field the pin carries`** —
  `nix/pkgs/frr` read `frrSrc.branch`, which only exists on an npins `Git`
  pin, so the release pin below broke evaluation of `frr.host`. Behaviour is
  unchanged for a branch pin; this commit stands alone green.
- **`bump: frr to 10.7.1`** — upstream FRR moves from a branch pin on
  `stable/10.6` to a release pin bounded below 10.8, so a pin bump lands on a
  tagged 10.7.x release and cannot wander off the train `frr-dp` and
  `dplane-plugin` are on.
- **`bump: rdma-core 64.0 -> 65.0`**

Commits are ordered so each one is green on its own.

This is still a draft to see if we make it through vlab.


daniel-noland:
split into 3 PRs as #1833 #1834 and #1835

PR #1833, opened 2026-09-20
unblock ci: drop opengrep's `--config auto`

daniel-noland:
`auto` is a live, unauthenticated fetch of
https://semgrep.dev/c/p/default.

We are using opengrep, not semgrep, and the rules fetch is rate limited.  This rate limit causes random failures in CI, so we (sadly) need to disable this.

The main reason to use opengrep is for our custom rules anyway.


justfile
```diff
@@ -534,7 +534,7 @@ check-deps-reuse:
 [script]
 opengrep:
     {{ _just_debuggable_ }}
-    opengrep scan --experimental --verbose --error --config auto --config .semgrep/rules
+    opengrep scan --experimental --verbose --error --config .semgrep/rules
```
qmonnet, line 537: Couldn't we make it a fallback?

```justfile
    opengrep scan --experimental --verbose --error --config auto --config .semgrep/rules || \
        opengrep scan --experimental --verbose --error --config .semgrep/rules
```

qmonnet (approved):
Discussed during the meeting - OK to drop the flag entirely. Thanks!

PR #1839, opened 2026-09-21
Minor NAT masquerade clean-ups

qmonnet:
I'm preparing a PR for EIM for NAT and it's starting to grow. Let's take these commits out as a separate PR.


qmonnet:
Superseded with #1847

